PR #24, opened 2015-01-29
Fix the sample loop correction code.

ejulien:
Hello, please consider the following patch for inclusion in the official repository.

I am not sure of the exact intent of the original code but it was clearing causing quality problems as you can see below. The code was duplicating the last sample of the instrument loop end before copying samples from the loop start.

libxmp_loopbug

The picture was done while debugging [`explora.xm` by Clawz](http://modarchive.org/index.php?request=view_by_moduleid&query=140994) at roughly 23 seconds in the lead synth sounds very dirty even when spline interpolation is used.

No regression was noted on various .mod and .xm I tested.


cmatsuoka:
Thank you. I don't know the original purpose of the faulty code, but it makes much more sense after the fix. I'll update the regression tests properly to reflect the fixed waveforms.

PR #37, opened 2015-08-11
Fix memory leak and buffer overrun.

cspiegel:


cmatsuoka:
Merged, thanks!

cmatsuoka:
Merged, thanks!

PR #74, opened 2016-09-03
Change Android preprocessor macro from ANDROID to __ANDROID__

Ghabry:
I usually don't use ndk-build to crosscompile libs for Android but just the normal configure && make-way because this just works for any library.

The problem here is that ndk-build defines ```ANDROID``` but the built-in macro that is defined for any Android built is ```__ANDROID__```.

See also: https://stackoverflow.com/questions/6374523/how-to-detect-compilation-by-android-ndk-in-a-c-c-file

In the end you get strange linker errors like ```error: symbol xmp_get_player has undefined version XMP_4.2``` because the macro was wrong.

Is this here the same repository as for xmp-lite? Because I'm using xmp-lite which has the same problem.


carstene1ns:
@Ghabry: Do not forget to update `config.sub` too, the version here is from 2009 and has the needed host triplet not included (see [our toolchain script](https://github.com/EasyRPG/buildscripts/blob/master/android/2_build_toolchain.sh#L37-L39)). The latest file is available [here](http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD).

Ghabry:
Thanks. Done.

cmatsuoka:
Merged, thanks.

PR #76, opened 2016-09-22
Fix for IT's Volume Panning Effect

NoSuck:
I noticed with IT files that volume-column panning parameters of both 64
(the max) and 63 resulted in identical panning.  This patch corrects
that.


cmatsuoka:
Merged, thanks.

PR #80, opened 2016-10-14
IT Message (Comment) Length Miscalculation

NoSuck:
The IT message length includes the null terminator, yet up until now, libxmp has allocated an extra character for a null terminator.  This has resulted in every IT song message being appended with an extraneous period.


cmatsuoka:
Merged, thanks.

PR #82, opened 2016-11-06
Fix libxmp-lite building (wrong format loaders)

carstene1ns:
Currently trying to use the lite version fails with undefined
references to the old format loader names.

This bug was introduced in d018decfb98d: "Add prefix to format loaders"


cmatsuoka:
Merged, thanks.

PR #83, opened 2016-12-08
some fixes, watcom and os2 support

sezero:


cmatsuoka:
All merged, thanks!

sezero:
Note about commit d438c2b :  there _might_ be a better fix, the one I did was quick and lazy.

PR #84, opened 2016-12-11
src/depackers/vorbis.c: Fix warnings:

sezero:
The lookup_values warning was valid. The other three were bogus, but silenced them anyway.

src/depackers/vorbis.c(981): Warning! W107: Missing return value for function 'get8'
src/depackers/vorbis.c(1010): Warning! W107: Missing return value for function 'getn'
src/depackers/vorbis.c(3609): Warning! W136: Comparison equivalent to 'unsigned == 0'
src/depackers/vorbis.c(4299): Warning! W107: Missing return value for function 'stb_vorbis_get_file_offset'

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -978,6 +978,7 @@ static uint8 get8(vorb *z)
    return c;
    }
    #endif
+   return 0;/* silence pesky compiler warnings */
 }
 
 static uint32 get32(vorb *f)
@@ -1007,6 +1008,7 @@ static int getn(vorb *z, uint8 *data, int n)
       return 0;
    }
    #endif
+   return 0;/* silence pesky compiler warnings */
 }
 
 static void skip(vorb *z, int n)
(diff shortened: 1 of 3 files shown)


sezero:
Pushed another changeset to fix two warnings about less-than-zero checks on unsigned vars.

sagamusix:
stb_vorbis warnings should probably be reported upstream (https://github.com/nothings/stb)

carstene1ns:
Some warnings might even be [fixed in an updated version](https://github.com/nothings/stb/blob/14e6a98469d8f378e5f6a1d77d166244fe452d6c/stb_vorbis.c#L40-L51) already. Maybe it is worth to pull in some upstream changes as well.

sezero:
Personally, I don't care much about stb_vorbis and the latest version
of it has those assert()s too which aren't suitable for a library.
(One may build with NDEBUG, but I don't know how safe is that.)

On top of that, libxmp seems to have added its own sanity checks some
of which seem wrong, e.g. compute_codewords() shouldn't return -1 for
error but 0. Things need cleaning, assert()s need replacing. Upgrading
to a new version might make things harder, IMO.

sagamusix:
@sezero I'm not quite sure what your concerns are with assertions. assert() is useful in libraries just like in executables, and of course NDEBUG should be defined when doing a "release" build.

sezero:
Yeah, abort() from a library is so _very_ useful.

Just return proper error codes and handle them instead.

Issue #92, opened 2017-02-13
Samples corrupted in fracture.stm and other STM modules


viiri:
https://github.com/cmatsuoka/libxmp/blob/master/test-dev/data/m/fracture.stm

viiri:
[stm.zip](https://github.com/cmatsuoka/libxmp/files/770168/stm.zip)
More samples attached.

sezero:
I think this is fixed by commit ca11190066 (#93). Closing.

PR #93, opened 2017-02-13
Update stm_load.c

viiri:
Samples corrupted in fracture.stm and other STM modules  #92
https://github.com/cmatsuoka/libxmp/issues/92

### DIFF
--- a/src/loaders/stm_load.c
+++ b/src/loaders/stm_load.c
@@ -270,7 +270,8 @@ static int stm_load(struct module_data *m, HIO_HANDLE * f, const int start)
 	D_(D_INFO "Stored samples: %d", mod->smp);
 
 	for (i = 0; i < mod->ins; i++) {
-		if (mod->xxs[i].len > 1) {
+		if (sfh.ins[i].volume && sfh.ins[i].length) {
+			hio_seek(f, start + sfh.ins[i].rsvd1 << 4, SEEK_SET);
 			if (libxmp_load_sample(m, f, 0, &mod->xxs[i], NULL) < 0)
 				return -1;
 		} else {


src/loaders/stm_load.c
```diff
@@ -270,7 +270,8 @@ static int stm_load(struct module_data *m, HIO_HANDLE * f, const int start)
 	D_(D_INFO "Stored samples: %d", mod->smp);
 
 	for (i = 0; i < mod->ins; i++) {
-		if (mod->xxs[i].len > 1) {
+		if (sfh.ins[i].volume && sfh.ins[i].length) {
```
sagamusix: I keep seeing this check for default sample volume everywhere (probably because everyone copies libmikmod/libmodplug) but it's wrong. Any sample should be loaded regardless of its default volume. Yes, even those garbage samples in fracture.stm.

viiri: Volume check exists in original Scream Tracker 2 loader, i have disassemled it. Problem is not in garbage samples (in fact there is no "garbage samples", ST2 does not load sample if it's volume and length equals zero), but in that the sfh.ins[i].rsvd1 << 4 is actually sample offset. Try to load wonder.stm from my attachement and you will be surprised.

sagamusix: Okay, after some further tests, ST2.3 plays random noise from memory when trying to trigger a sample with a default volume of 0. I first interpreted this noise as being the same as in the file (e.g. with fracture.stm), but when hex-editing some "normal" sample to have a default volume of 0, ST2 plays some other random noise.
As a conclusion, a default volume of 0 is undefined behaviour, so you can actually load those samples or not - if they are used in a pattern, they will sound broken either way.

viiri: As you can see in [code](http://pastebin.com/c5kcFPHA), ST2 treats volume 0 as a flag of empty sample. Much more important, that the ST2 does not load samples as a continuous stream, but uses sfh.ins[i].rsvd1 to calculate offset (look at seg003:18D1 in code) from begining of the file. libxmp (and other libraries) does not do that and corrupt samples while loading, you can hear it on wonder.stm and other modules with non serial sample arrangement.

sagamusix: Yes, I'm not arguing against seeking to the correct sample position based on "rsvd1" -- in fact, I have done the same in libopenmpt/OpenMPT for a long time already. I'm just saying that whether you load samples with a default volume of 0 or not doesn't really matter because when you try to play such a sample in ST2, it results in undefined behaviour (noise is played).

viiri: Ok, I just offer to reproduce original loader. Playing noise instead of silence looks like a bug in a player part of ST2. Actually tracker and player are independent parts in ST2. I think that silent empty samples will be more correct than noisy, and as I know, other trackers do just that.

sagamusix:
This should really be merged, together with some of the nice new findings in st2play.

Slightly off-topic... @viiri do you actually have any STMs older than format version 2.21? I see you added support for them in st2play but I cannot find any such files in the wild. Also no files in the wild seem to make use of the "special" note values that reduce a pattern entry from 4 to 1 byte, but hex-editing a file suggests that this works as intended in ST2.

cmatsuoka:
Will add findings from st2play in xmp's st_load.c. WUZAMOD! magic added
recently (does anyone know what this comes from?), and tempo settings using
the "sample_rate / (49 - ((tempo_mul[ticks_per_row] * (tempo & 0x0f)) >>
4))" formula to be fixed next.

viiri:
Hello again.
I have not seen any STMs older than 221. The STM loader in st2play is from disassemled Scream Tracker 2 (I have versions 2.2, 2.21, 2.24, 2.25 and 2.3, If you need them - I will share). I will try to ask exUSSR demoscene community for old tracker modules for tests. By the way, I have made "minimal" valid STM of version 2.00 (https://files.fm/down.php?cf&i=8gpt8z2u&n=stm200%5B1%5D.stm), that should be loaded by full compatible loader.
Compressed patterns always used in STS song, but not in STM modules, seems like a bug in save function. STS has no internal samples in file, samples were loaded from external library. I will commit support of external samples in st2play soon.

WUZAMOD! is a signature of one of many mod2stm converters, I have this one somewhere in my DOS-software collection. There is also DNS.STM, you can download it by this link: https://files.fm/down.php?cf&i=ayxshudh&n=DNS%5B1%5D.STM. This one was created in Swave Pro and has different signature. As I mentioned earlier, Scream Tracker does not rely on the signature (see the "minimal" STM, there is no signature at all). So, I think, it will be correct behavior to ignore this string and use it just for information purpose.

viiri:
Oh! Forget to mention, I have fully disassemled all available ST2 versions and found that they are all slightly different, except of st225 (constants, frequencies, tables, supported effects - yes, I have reconstructed famous "buggy" arpeggio from st224) and now working on modular player which allows to switch st2 engine dynamically.

sagamusix:
> So, I think, it will be correct behavior to ignore this string and use it just for information purpose.

Technically it's correct, yes, but for libraries that are often used together with other libraries (like libopenmpt in ffmpeg), it can be dangerous. That is because there aren't many ways to detect if an STM file is actually valid without these magic bytes. And we definitely want to create as few false positives as possible when trying to load modules.

However, I just got an idea that's worth trying out: Accept only tracker signatures that:
- Only contain ASCII characters 0x20 - 0x7E
- Have no non-null character after the last null
I will have to check if this would create more false positives with random binary files, though. Especially sparse files with lots of null bytes (e.g. a small ID3 tag in front of an MP3 file, padded to 2KB) would most likely be mis-detected as STM in that scenario.

> I have versions 2.2, 2.21, 2.24, 2.25 and 2.3, If you need them - I will share

That would be appreciated. ModLand has 2.21, 2.24 and 2.3 only.

> yes, I have reconstructed famous "buggy" arpeggio from st224

I'm not really knowledgeable about the differences in various ST2 versions, would you care to elaborate?

sagamusix:
By the way @cmatsuoka, the tempo calculation underflows for tempo values 0x06-0x0F, producing a negative amount of samples per frame. The samples per tick value then wraps around in ST2 (it's an unsigned 16-bit value), so this needs to be taken care of. As the exact amount of wraparound depends on the mix rate, I chose to first do the calculation for ST2's highest internal mix rate (23863 Hz), and then scale that to the actually used mix rate. This way, the result is at consistent.

viiri:
> That would be appreciated. ModLand has 2.21, 2.24 and 2.3 only.

http://old-dos.ru/dl.php?id=16829 - registered ST2.2 or ST2.21, I still cannot determine this exactly! :)
ftp://ftp.scene.org/pub/mirrors/flerp/Util.PC/Tracker/st22.zip - registered ST2.25, also there is ST20.EXE executable in this archive, this one is not Scream Tracker 2.0, but broken SpacePigs SoundTracker V2.0.

> I'm not really knowledgeable about the differences in various ST2 versions, would you care to elaborate?

Here is a small piece of code from currently unreleased st2play v2.24:

		case FX_ARPEGGIO:
			/*
			 * Skaven's note from http://www.futurecrew.com/skaven/oldies_music.html
			 *
			 * FYI for the tech-heads: In the old Scream Tracker 2 the Arpeggio command
			 * (Jxx), if used in a single row with a 0x value, caused the note to skip
			 * the specified amount of halftones upwards halfway through the row. I used
			 * this in some songs to give the lead some character. However, when played
			 * in ModPlug Tracker, this effect doesn't work the way it did back then.
			 */

			note_add = 0;

			if((ctx->current_tick / 3) == 0)
				note_add = ch->event_infobyte & 0x0f;

			octa =  ch->last_note & 0xf0;
			note = (ch->last_note & 0x0f) + note_add;

			if(note >= 11) {
				note -= 12;
				octa += 16;
			}

			ch->period_current = period_table[octa | note] * 8448 / (ch->smp_c2spd ? ch->smp_c2spd : 8888);
			ch->period_target = ch->period_current;

			update_frequency(ctx, chn);
			break;
`

cmatsuoka:
Thanks Sergei, this information will be also useful to complete the ST2 lineage in the tracker-history graph (http://helllabs.org/tracker-history/trackers.svg).

sagamusix:
@viiri Hmm, interesting! I always thought Skaven's comment just refers to the fact that Arpeggio stays on the last played note even after the end of the row, which he definitely makes use of in MORPH.STM, but I can hear the difference between 2.24 and 2.3 now.

AliceLR:
I found this bug independently of this PR while investigating for https://github.com/sezero/mikmod/issues/6. Here's another .STM from [Mod Archive](https://modarchive.org/index.php?request=view_by_moduleid&query=180379) that is broken by the absence of this check.

[georythm.stm.zip](https://github.com/cmatsuoka/libxmp/files/5370430/georythm.stm.zip)

sezero:
Is this patch well-tested?

sagamusix:
OpenMPT uses exactly the same logic. In that sense, the logic itself is well-tested. ;)

AliceLR:
Seconded; I wasn't as thorough testing this patch as I was for the MikMod patch but the logic is exactly the same.

sezero:
OK, I pulled it in my fork

sagamusix:
Have you been in contact with Claudio recently? It seems that he is busy with different things, which is completely okay, but maybe it's time to pull in a co-maintainer for this repository rather than having yet another fork.

sezero:
> Have you been in contact with Claudio recently? It seems that he is busy with different things, which is completely okay, but maybe it's time to pull in a co-maintainer for this repository rather than having yet another fork.

I pinged here multiple times for my patches, no answer.

sezero:
> Wow, I just got a huge flood of email activity from here. Thanks for merging all of these. : )

You're welcome.

@cmatsuoka gave me co-maintainership, so I'm merging things which are tested well-enough.

sagamusix:
Lesson learned: Don't just trust that people read their GitHub notification, a personal email can often achieve more :)

Issue #97, opened 2017-07-13
.it playback bugs


AliceLR:
I have encountered one (possibly two) playback bugs with .it files.  I am testing these files using libxmp 4.4.1 and xmp 4.1.0 built with MSYS2 on Windows 10.

**Bug 1**: This .it file uses the retrigger command (Qxy) on a snare drum. When the volume parameter is C, it increases the volume of the playing sample as expected (volume +8). However, when the volume parameter is 4 (volume -8), the volume does not change at all.  This file plays back correctly in Modplug Tracker, OpenMPT, and Schism Tracker.

The .it file: http://a.spectere.net/music/labyrinth/ZX_07.it
How it should sound: https://www.youtube.com/watch?v=Fd8UTuoWLi0

**Bug 2**: This .it file has at the end of its first pattern a loop (SB0 / SB2) with a C00 on the same row as the SB2.  The C00 is not necessary to the file, and was probably left in by mistake.  Modplug and OpenMPT (possibly incorrectly) repeat the loop all three times before continuing to the start of the next pattern.  Schism Tracker does not loop, but continues to the start of the next pattern.  xmp reaches the C00 and then seemingly _jumps to the end of the next pattern_.  I'm not sure which of these three behaviors is supposed to be correct, but the libxmp behavior still seems questionable.  They all correctly loop three times when the C00 is removed.

The .it file: http://a.spectere.net/music/labyrinth/ZX_11.it
How it sounds in Modplug: https://www.youtube.com/watch?v=6UzfU-Fbb4U

sagamusix:
> Modplug and OpenMPT (possibly incorrectly) repeat the loop all three times before continuing to the start of the next pattern. 

This behaviour is correct (consistent with Impulse Tracker).

AliceLR:
This change seems to fix bug #1 (effects.c, line 683). I moved the retrig.type line into the fxp check instead of the note check.  It's only updating the volume parameter when a new note is played, instead of when a new effect is executed.

`	case FX_MULTI_RETRIG:	/* Multi retrig */
		EFFECT_MEMORY_S3M(fxp);
		if (fxp) {
			xc->retrig.val = fxp;
			xc->retrig.type = MSN(xc->retrig.val);
		}
		if (note) {
			xc->retrig.count = LSN(xc->retrig.val) + 1;
		}
		SET(RETRIG);
		break;`

AliceLR:
The correct fix to bug 2 seems to be:

[...]

PR #98, opened 2017-07-14
Fixed .it bug where Qxy would ignore the volume parameter

AliceLR:
This fixes a bug with the .it retrigger command (#97) where the volume parameter would be updated only on rows with a note.  With this fix, it will correctly update on any row the retrigger command is present on.


cmatsuoka:
Merged, thanks.

PR #99, opened 2017-07-14
Fixed .it bug where Cxx on same row as SBx would not be ignored

AliceLR:
This fixes a bug with .it playback (#97) where, when the end of a loop (e.g. SB2) occurs on the same line as a break (e.g. C00), libxmp would attempt to break to the next pattern and go to the start of the loop at the same time.  Instead, it will now ignore the break until the loop is finished.


cmatsuoka:
Merged, thanks.

NoSuck:
Nice catch!

Issue #100, opened 2017-07-25
Contribution Clarification


NoSuck:
Hello.  In the never-ending quest to make libxmp perfect, I thought the following functions would be useful (particularly for people writing trackers):

* `xmp_set_row` moves current playback position to a specified row; this pairs well with `xmp_set_position`.
* `xmp_add_channels` modifies a MOD's number of channels; it is versatile, so as not to clog up libxmp's API
* `xmp_move_channels` offers versatility similar to `xmp_add_channels`

I have already made and tested the functions but have two questions:

* Are API additions like the above better suited for `master` or `libxmp-devel`?
* Please confirm the process of adding to libxmp's API.  I see the following that need to be added:  actual code, `libxmp.h` EXPORT declaration, `libxmp.map` declaration (under 4.4?), documentation in `libxmp.rst`, test module in `test-dev` directory.  Did I miss anything?

Perhaps a `CONTRIBUTING` file could be added that addresses these questions (and any others that may arise in the future).

Thanks for everything and happy tracking!

cmatsuoka:
These API calls are interesting and could be a worthful addition to the 5.0 API. The current 4.x API is limited by the internal data organization but in devel we have more freedom to change the way things work.

AliceLR:
I'd like to second `xmp_set_row`; this would be useful for MegaZeux, which was capable of row seeking with libmodplug but isn't with libxmp. I don't think any MegaZeux games actually use this feature.

Another feature libmodplug has that would be useful for MegaZeux is to be able to perform `xmp_set_position` without resetting the channels. Currently, when a game changes the order of a module, the channel panning and volume in libxmp are reset to center and maximum.  This makes things like changing order during gameplay to dynamically fade channels less flexible. Again, I don't think any MegaZeux games actually rely on this.

NoSuck:
By limitations of 4.x's "internal data organization", would you happen to be referring to the `xxt` (`xmp_track`) member of `xmp_module`?  Sadly, I was unable to create a watertight `xmp_add_channels` or `xmp_move_channels` because of how different MOD formats conceive of their `xmp_track` arrays.  (XM in particular is an oddball.)

`xmp_set_row` is a different story and should work fine even for 4.x.  I apologize for sitting on the patch so long and will submit shortly.

```shell
# xxt is a trk-long array of xmp_tracks.  Reorganization of channels or patterns requires reorganizing this array.
# grep -RIn '\<trk \?=[^=]' libxmp/src

[...]

PR #105, opened 2017-10-14
API Addition: xmp_set_row

NoSuck:
This patch adds code, documentation, and unit testing for a new function, xmp_set_row, as discussed in issue #100.  I must confess, however, that I do not fully understand the purpose of the final if block in set_position, which is the reason I had to make a special case for a negative pos member of player_data.

Please review and share your thoughts.  Thank you for maintaining libxmp.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -632,6 +632,22 @@ int xmp_set_position(xmp_context c, int pos)
     The new position index, ``-XMP_ERROR_INVALID`` of the new position is
     invalid or ``-XMP_ERROR_STATE`` if the player is not in playing state.
 
+.. _xmp_set_row():
+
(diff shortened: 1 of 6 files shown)


AliceLR:
It doesn't look like there will be a new release soon and this feature would be useful for MegaZeux.  Can we patch this into our copy of libxmp?

NoSuck:
Please feel free.  Claudio has been active over the holiday, though, so you may want to see how he handles this.

Note that this patch is affected by issue #110.

NoSuck:
Any word on this?

cmatsuoka:
Hi,

I'll bump the API version to include the new call.

cmatsuoka:
Merged, thanks.

Issue #110, opened 2017-11-03
Travis CI Check Fails for Clang


NoSuck:
I created a branch based off master (19ad5b326e228bdf7a80a35244c31a7b5c8e0897), added a one-line comment, and attempted to commit.  Travis CI complained that "all checks failed."  Can anyone shed light on this?

https://github.com/cmatsuoka/libxmp/pull/109

This affects some of my other PRs, so I would like to resolve it.

AliceLR:
I believe this was (temporarily) fixed here: 01951bf3f650372d81493f550e6844c423dba3e5

This should probably be resolved anyway, though, because it makes running `make devcheck` with ASan very annoying.

Copying two (of many) test failure outputs here as examples:
```
test 20: test_sample_load_8bit: 
=================================================================
==24840==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 646 byte(s) in 5 object(s) allocated from:
    #0 0x7f76e9c933cf in __interceptor_malloc (/lib64/libasan.so.6+0xab3cf)
    #1 0x456cc6 in libxmp_load_sample ../src/loaders/sample.c:272

Direct leak of 32 byte(s) in 1 object(s) allocated from:
    #0 0x7f76e9c93587 in __interceptor_calloc (/lib64/libasan.so.6+0xab587)
    #1 0x450dc9 in hio_open ../src/hio.c:322

SUMMARY: AddressSanitizer: 678 byte(s) leaked in 6 allocation(s).
**fail**
```
```
test 26: test_depack_pp: 
=================================================================
==24852==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 9296 byte(s) in 1 object(s) allocated from:
    #0 0x7f76e9c93587 in __interceptor_calloc (/lib64/libasan.so.6+0xab587)
    #1 0x7f76e9a37d72 in xmp_create_context src/control.c:40

Indirect leak of 40570 byte(s) in 6 object(s) allocated from:
    #0 0x7f76e9c933cf in __interceptor_malloc (/lib64/libasan.so.6+0xab3cf)
    #1 0x7f76e9a428e9 in libxmp_load_sample src/loaders/sample.c:272

Indirect leak of 24056 byte(s) in 1 object(s) allocated from:
    #0 0x7f76e9c93587 in __interceptor_calloc (/lib64/libasan.so.6+0xab587)
    #1 0x7f76e9a413c9 in libxmp_init_instrument src/loaders/common.c:56

[...]

PR #111, opened 2017-11-04
Clamp Number of IT Envelope Nodes at Load Time

NoSuck:
Previously, libxmp would return an error upon loading an IT file containing an instrument envelope with an impossibly large number of nodes.  Now, it clamps to the IT spec maximum (25).  The new behavior has been tested to match Impulse Tracker, Schism Tracker, and DUMB.


NoSuck:
This PR may be affected by issue #110.

cmatsuoka:
Merged, thanks.

PR #116, opened 2018-01-30
New and shiny STM loader

viiri:
Support all format versions of modules and songs except 1.00 because of different encoding schema.


sagamusix:
Please note that, as mentioned in #93, it's a bad idea to completely remove any validity checks on the tracker string in the context of a decoder library that may be used in conjunction with other libraries. I would, at the very least, reject all files containing characters 1-31 and 127-255 in the tracker name to reduce the number of false positives.

src/loaders/stm_load.c
```diff
@@ -71,15 +95,17 @@ static int stm_test(HIO_HANDLE * f, char *t, const int start)
 	hio_seek(f, start + 20, SEEK_SET);
 	if (hio_read(buf, 1, 8, f) < 8)
 		return -1;
-	if (memcmp(buf, "!Scream!", 8) && memcmp(buf, "BMOD2STM", 8) && memcmp(buf, "WUZAMOD!", 8))
-		return -1;
 
-	hio_read8(f);
+	if (hio_read8(f) != 0x1a)
```
sagamusix: I found that ST2 ignores this, ST3 doesn't. putup11.stm from ModLand has 0x02 instead of 0x1A here.

src/loaders/stm_load.c
```diff
@@ -131,12 +159,48 @@ static int stm_load(struct module_data *m, HIO_HANDLE * f, const int start)
 	sfh.type = hio_read8(f);	/* 1=song, 2=module */
 	sfh.vermaj = hio_read8(f);	/* Major version number */
 	sfh.vermin = hio_read8(f);	/* Minor version number */
-	sfh.tempo = hio_read8(f);	/* Playback tempo */
-	sfh.patterns = hio_read8(f);	/* Number of patterns */
-	sfh.gvol = hio_read8(f);	/* Global volume */
-	hio_read(&sfh.rsvd2, 13, 1, f);	/* Reserved */
+	version = (100 * sfh.vermaj) + sfh.vermin;
+
+	if (version == 100 || version > 221) {
```
sagamusix: Should probably be more tight and reject unknown versions not found in the wild?

src/loaders/stm_load.c
```diff
@@ -269,10 +337,28 @@ static int stm_load(struct module_data *m, HIO_HANDLE * f, const int start)
...
+
 	for (i = 0; i < mod->ins; i++) {
-		if (mod->xxs[i].len > 1) {
-			if (libxmp_load_sample(m, f, 0, &mod->xxs[i], NULL) < 0)
-				return -1;
+		if (sfh.ins[i].volume && sfh.ins[i].length) {
+			if (sfh.type == STM_TYPE_SONG) {
+			    HIO_HANDLE *s;
+			    char sn[256];
+			    snprintf(sn, XMP_NAME_SIZE, "%s%s", pathname, mod->xxi[i].name);
```
sagamusix: This should probably guard against path traversal issues by not allowing characters such as / and \ in the filename. Maybe only accept alphanumerics in filenames to completely minimize the risk posed by malicious files.

src/loaders/stm_load.c
viiri: Yes, ST2 ignores this as you can see in st2play, but for now this is the simplest method to probe format. Where did you get putup11.stm with 0x02 instead of 0x1a? Mine copy from SWave Pro distribution has correct header. It looks like patched manually.

sagamusix: From ModLand: https://www.exotica.org.uk/mediawiki/index.php?title=Special%3AModland&md=search&mod=putup11&aut=&for=&col=&ins=

src/loaders/stm_load.c
viiri: I have copied this piece of code from mod_load.c

sagamusix: Okay, then Claudio should review this probably in all similar places, not just here.

src/loaders/stm_load.c
viiri: Here you can download STM song of version 1.10: https://files.fm/u/xe3g4ncq
This one was ripped from FCSLIDE1 demo. This is the earliest STM that I have seen.
So I offer to check for fixed list of version numbers: 100 (I will implement decoding of patterns data), 110, 200, 210, 220 and 221.

sagamusix: Sounds reasonable to me.

src/loaders/stm_load.c
viiri: Thank you. Looks very strange, putup10.stm from Modland is Putup11.stm from Swave Pro distribution and vice versa. The files absolutely identical except of 0x1a byte after tracker "signature". Seems like someone has patched files with unknown reason.

sagamusix: I think the file on ModLand was simply a result of a broken FTP upload in the dark ages of file transfer errors. If you could point me at the original file with the correct 0x1A byte, that would be appreciated, because then I can tell ModLand admins to exchange the file.

viiri: Soundwave Pro: http://old-dos.ru/index.php?page=files&mode=files&do=show&id=5414
Putup songs: https://files.fm/u/4mtj3w2f

sezero:
> I can try to verify the changes here when I get a chance, I just haven't yet (the other STM patch was much simpler to check...). Also, I can look at the aforementioned security issues when generating sample/etc. filenames.

Well, I haven't followed the lengthy discussion here. (And now there is a merge conflict...)

viiri:
Oops! Something went wrong when I tried to update my fork. Too much commits. I have to reopen PR.

PR #121, opened 2018-03-28
Include a copy of the LGPL 2.1 in the root dir for GitHub

superusercode:
The README states that the library should include a copy of the LGPL, as seen below:
```
You should have received a copy of the GNU Lesser General Public
License along with this library; if not, write to the Free Software
Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301 USA
```

This pull request does that, and as a result GitHub now recognizes the repo as LGPL-licensed.

screenshot-2018-3-28 superusercode libxmp

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,504 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 2.1, February 1999
+
+ Copyright (C) 1991, 1999 Free Software Foundation, Inc.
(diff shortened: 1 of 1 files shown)


sezero:
The tree already has docs/COPYING

superusercode:
I'm aware of that, I've just done this for the sake of GitHub recognizing that it has been licensed.

PR #125, opened 2018-10-14
fix build in c90 mode

sezero:


cmatsuoka:
Oops. That was a typo, thanks for spotting and fixing.

PR #127, opened 2018-10-25
Bug Fix for Mute Status on Player Creation

NoSuck:
This was the cause of some confusion.  Muted channels would be muted in playback but not according to the status returned by `xmp_channel_mute`.


NoSuck:
... another PR affected by #110, hehe.

PR #128, opened 2018-10-26
Prevent Clobbering of Muted Channels' Volumes

NoSuck:
Libxmp was throwing away volume values for IT files' muted channels, even though mute status is completely independent of volume.


NoSuck:
#110

PR #129, opened 2018-10-26
T00 Now Repeats Previous Slide (IT)

NoSuck:
Previously, libxmp would ignore `T00` IT effects.  I tested with it214v5 and observed that `T00` does indeed repeat its channel's most recent slide.


NoSuck:
... and for anyone just tuning in, this PR is affected by #110.

PR #131, opened 2018-12-11
updates to make it possible building a proper static win32 lib

sezero:
c.f.: PR #126


cmatsuoka:
Thanks Ozkan, excellent.

PR #134, opened 2019-01-03
Add xmp_test_module_from_memory and xmp_test_module_from_file

Ghabry:
This adds two new helper functions which are equivalent to xmp_test_module but they support memory and file handles (similar to xmp_load_module_*).

Same limitations as for load_module applies: Memory doesn't support compressed data (which appears to be caused by the unpackers because they use C File IO instead of HIO).

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -299,6 +299,74 @@ int xmp_test_module(char \*path, struct xmp_test_info \*test_info)
     and uncompression failed, or ``-XMP_ERROR_SYSTEM`` in case of system error
     (the system error code is set in ``errno``).
 
+.. xmp_test_module_from_memory():
+
+int xmp_test_module_from_memory(void \*mem, long size, struct xmp_test_info \*test_info)
+````````````````````````````````````````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.5]* Test if a memory buffer is a valid module. Testing
+  memory does not affect the current player context or any currently loaded
+  module.
+
+  **Parameters:**
+    :mem: a pointer to the module file image in memory. Multi-file modules
+      or compressed modules can't be tested in memory.
+
+    :size: the size of the module, or 0 if the size is unknown or not
+      specified. If size is set to 0 certain module formats won't be
+      recognized, the MD5 digest will not be set, and module-specific
+      quirks won't be applied.
+
+    :test_info: NULL, or a pointer to a structure used to retrieve the
+      module title and format if the memory buffer is a valid module.
+      ``struct xmp_test_info`` is defined as::
+
+        struct xmp_test_info {
+            char name[XMP_NAME_SIZE];      /* Module title */
+            char type[XMP_NAME_SIZE];      /* Module format */
+        };
+
+  **Returns:**
+    0 if the memory buffer is a valid module, or a negative error code
+    in case of error. Error codes can be ``-XMP_ERROR_FORMAT`` in case of an
+    unrecognized file format or ``-XMP_ERROR_SYSTEM`` in case of system error
+    (the system error code is set in ``errno``).
+
+.. xmp_test_module_from_file():
+
+int xmp_test_module_from_file(FILE \*f, long size, struct xmp_test_info \*test_info)
+````````````````````````````````````````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.5]* Test if a modle from a stream is a valid module.
+  Testing streams does not affect the current player context or any
+  currently loaded module.
+
+  **Parameters:**
+    :f: the file stream. Compressed modules that need an external depacker
+      can't be tested from a file stream. On return, the stream position is
+      undefined.
+
+    :size: the size of the module, or 0 if the size is unknown or not
+      specified. If size is set to 0 certain module formats won't be
+      recognized, the MD5 digest will not be set, and module-specific
+      quirks won't be applied.
+
+    :test_info: NULL, or a pointer to a structure used to retrieve the
+      module title and format if the memory buffer is a valid module.
+      ``struct xmp_test_info`` is defined as::
+
+        struct xmp_test_info {
+            char name[XMP_NAME_SIZE];      /* Module title */
+            char type[XMP_NAME_SIZE];      /* Module format */
+        };
+
+  **Returns:**
+    0 if the stream is a valid module, or a negative error code
+    in case of error. Error codes can be ``-XMP_ERROR_FORMAT`` in case of an
+    unrecognized file format, ``-XMP_ERROR_DEPACK`` if the stream is compressed
+    and uncompression failed, or ``-XMP_ERROR_SYSTEM`` in case of system error
+    (the system error code is set in ``errno``).
+
 .. _xmp_load_module():
 
 int xmp_load_module(xmp_context c, char \*path)
--- a/src/load.c
+++ b/src/load.c
@@ -105,7 +105,7 @@ static int execute_command(const char *cmd, const char *filename, FILE *t)
 	 * indefinitely. _popen works properly in a Console application.
 	 * To create a Windows application that redirects input and output,
 	 * read the section "Creating a Child Process with Redirected Input
-	 * and Output" in the Win32 SDK. -- Mirko 
+	 * and Output" in the Win32 SDK. -- Mirko
 	 */
 	p = popen(line, "rb");
 #else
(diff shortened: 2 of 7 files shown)


src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+int xmp_test_module_from_file(void *file, long size, struct xmp_test_info *info)
+{
+	HIO_HANDLE *h;
+	char buf[XMP_NAME_SIZE];
+	int i;
+	int ret = -XMP_ERROR_FORMAT;
+#ifndef LIBXMP_CORE_PLAYER
+	char *temp = NULL;
+#endif
+	FILE *f = fdopen(fileno((FILE *)file), "rb");
```
AliceLR, line 441: This line has the same bug as `xmp_load_module_from_file` that would be fixed by one of the patches in pull request #173.

Ghabry: ohh, there is progress here again? This is awesome 👍 . Will take a look at the code later.

Ghabry: For testing if a handle is supported it is imo not nice that afterwards my FILE-handle is trashed. I would expect that I can seek backwards and pass it directly to ``xmp_load_module_from_file``.
Though I won't really need the ``FILE`` case in my code. I only operate on memory buffers because the caller uses ``std::fstream`` which is not ``FILE``-compatible anyway.

AliceLR: Currently, the progress here seems to be just sezero and I reviewing PRs...

Re: closing the `FILE` handle passed to `xmp_load_module_from_file` in that PR I linked, I agree that's not a great way to handle it. sezero posted a patch further down that PR that (IMO correctly) adds an hio function so that it can leave the user-provided `FILE` open without causing any portability issues with that `fdopen` hack.

sezero:
`xmp_load_module_from_file()` is the all same familiar story: At first,
it looks like a good idea to add a `load_something_from_file()` api to
a library, then you do it, then make a release, and the day after that
you realize how bad and doomed an idea it was and try (and fail) fixing
it in impossible ways.

#173 is about that...

sezero:
P.S.: I created https://github.com/cmatsuoka/libxmp/pull/178 as a better (IMO) approach to #173.

AliceLR:
> `xmp_load_module_from_file()` is the all same familiar story: At first,
> it looks like a good idea to add a `load_something_from_file()` api to
> a library, then you do it, then make a release, and the day after that
> you realize how bad and doomed an idea it was and try (and fail) fixing
> it in impossible ways.

I wonder if a better way would be to adopt the Vorbis and MikMod approach of asking for a set of user-defined callbacks to be used with the handle. That more clearly communicates that closing the handle is the user's responsibility (or a close callback can be provided).

I think the main challenge applying something like that to libxmp is you would have to provide 1) a nominal filename and 2) a callback to create more handles, for things like Startrekker parameter files and samples for song files. It's probably not worth the headache...

sezero:
Definitely not worth the headache, at least when (at the moment) libxmp seems abandoned by its author.

sezero:
@Ghabry: Please change your patches after commit bda3ba4fed30d

docs/libxmp.rst
```diff
@@ -299,6 +299,74 @@ int xmp_test_module(char \*path, struct xmp_test_info \*test_info)
...
+````````````````````````````````````````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.5]* Test if a modle from a stream is a valid module.
+  Testing streams does not affect the current player context or any
+  currently loaded module.
+
+  **Parameters:**
+    :f: the file stream. Compressed modules that need an external depacker
+      can't be tested from a file stream. On return, the stream position is
+      undefined.
```
sezero, line 347: Append here: `Caller is responsible for closing the file stream.`

docs/libxmp.rst
```diff
@@ -299,6 +299,74 @@ int xmp_test_module(char \*path, struct xmp_test_info \*test_info)
...
+    in case of error. Error codes can be ``-XMP_ERROR_FORMAT`` in case of an
+    unrecognized file format or ``-XMP_ERROR_SYSTEM`` in case of system error
+    (the system error code is set in ``errno``).
+
+.. xmp_test_module_from_file():
+
+int xmp_test_module_from_file(FILE \*f, long size, struct xmp_test_info \*test_info)
+````````````````````````````````````````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.5]* Test if a modle from a stream is a valid module.
```
sezero, line 340: Typo here:  `modle` -> `module`

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+	/* Use size < 0 for unknown/undetermined size */
+	if (size == 0)
+		size--;
+
+	if ((h = hio_open_mem(mem, size)) == NULL)
+		return -XMP_ERROR_SYSTEM;
+
+#ifndef LIBXMP_CORE_PLAYER
+	/* get size */
+	if (hio_size(h) < 256) {	/* set minimum valid module size */
```
sezero, line 388: I don't know why this is needed? AFAIK, `xmp_load_module_from_memory` doesn't have it.

src/load.c
sezero, line 441: Lose this var...

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+	HIO_HANDLE *h;
+	char buf[XMP_NAME_SIZE];
+	int i;
+	int ret = -XMP_ERROR_FORMAT;
+#ifndef LIBXMP_CORE_PLAYER
+	char *temp = NULL;
+#endif
+	FILE *f = fdopen(fileno((FILE *)file), "rb");
+
+	if ((h = hio_open_file(f)) == NULL)
```
sezero, line 443: Change here: `f`  ->  `(FILE *)file`

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+
+#ifndef LIBXMP_CORE_PLAYER
+			if (strcmp(format_loader[i]->name, "prowizard") == 0) {
+				hio_seek(h, 0, SEEK_SET);
+				pw_test_format(h, buf, 0, info);
+				is_prowizard = 1;
+			}
+#endif
+
+			fclose(h->handle.file);
```
sezero, line 477: Lose `fclose()` here

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+	return ret;
+}
+
+int xmp_test_module_from_file(void *file, long size, struct xmp_test_info *info)
+{
+	HIO_HANDLE *h;
+	char buf[XMP_NAME_SIZE];
+	int i;
+	int ret = -XMP_ERROR_FORMAT;
+#ifndef LIBXMP_CORE_PLAYER
```
sezero, line 438: change this guard to `LIBXMP_NO_DEPACKERS`

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+	char *temp = NULL;
+#endif
+	FILE *f = fdopen(fileno((FILE *)file), "rb");
+
+	if ((h = hio_open_file(f)) == NULL)
+		return -XMP_ERROR_SYSTEM;
+
+#ifndef LIBXMP_CORE_PLAYER
```
sezero, line 446: change this guard to `LIBXMP_NO_DEPACKERS`

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+			if (strcmp(format_loader[i]->name, "prowizard") == 0) {
+				hio_seek(h, 0, SEEK_SET);
+				pw_test_format(h, buf, 0, info);
+				is_prowizard = 1;
+			}
+#endif
+
+			fclose(h->handle.file);
+
+#ifndef LIBXMP_CORE_PLAYER
```
sezero, line 479: change this guard to `LIBXMP_NO_DEPACKERS`

src/load.c
```diff
@@ -362,6 +369,136 @@ int xmp_test_module(char *path, struct xmp_test_info *info)
...
+			if (info != NULL && !is_prowizard) {
+				strncpy(info->name, buf, XMP_NAME_SIZE - 1);
+				strncpy(info->type, format_loader[i]->name,
+							XMP_NAME_SIZE - 1);
+			}
+			return 0;
+		}
+	}
+
+#ifndef LIBXMP_CORE_PLAYER
```
sezero, line 492: change this guard to `LIBXMP_NO_DEPACKERS`

sezero (changes requested):

sezero:
This is now tracked / reviewed at PR #237

Issue #147, opened 2019-09-04
File "STARDSTM.GDM" does not play correctly


AliceLR:
When I play STARDSTM.GDM (what appears to be [this file](https://modarchive.org/index.php?request=view_by_moduleid&query=59344) after conversion for use with BWSB), many notes do not play. ~~This appears to be because the volume resets when the instrument number is repeated for MOD and S3M files, but libxmp doesn't do this when playing a GDM.~~ This is because empty notes are incorrectly loaded as low pitched notes that either hard to hear or do not play at all.

libxmp correctly plays the MOD and also correctly plays the S3M produced by saving this GDM in OpenMPT.  Both BWSB and OpenMPT play the first pattern of this GDM the same as the MOD I linked above (as far as I can tell).

[STARDSTM.GDM.zip](https://github.com/cmatsuoka/libxmp/files/3572972/STARDSTM.GDM.zip)

```
Loading stardstm.gdm (1 of 1)
Module name  : stardust memories
Module type  : GDM 1.00 (2GDM 1.23)
Module length: 31 patterns
Patterns     : 20
Instruments  : 31
Samples      : 31
Channels     : 4 [ 7 7 7 7 ]
Duration     : 3min49s
Speed[06] BPM[7D] Pos[00/1E] Pat[02/13] Row[37/3F] Chn[04/04]      0:00:06.5
```

AliceLR:
Closing this since the related pull request was merged.

AliceLR:
Sorry, actually reopening this one because it still needs a test case.

sezero:
Commit 044a98960ff should finalize this.

PR #148, opened 2019-09-04
Fix GDM loader to correctly handle empty notes

AliceLR:
This patches the GDM loader so it will correctly handle empty notes instead of trying to convert them like a regular note. This fixes STARDSTM.GDM (issue #147), which relies on empty notes.


cmatsuoka:
Merged, thanks!

PR #149, opened 2019-09-04
Fix test-dev builds for Windows

AliceLR:
This hack fixes test-dev builds for Windows (tested with MSYS2 3.0.7). I don't know if this is the particular fix you actually want for this or if it's something you want at all, but this worked for me.


sezero:
See #158 which is a possibly better fix for this.

PR #161, opened 2019-12-05
depackers/vorbis.c: fix sanity check error returns

sezero:
1. compute_codewords():
- compute_codewords() returns success (boolean), not error: replace one
  instance of negative return with FALSE.
- the (k == n) sanity check was duplicated: remove the libxmp-added one
  and just return (c->sorted_entries == 0) replacing the assert() along
  the way.

2. start_decoder():
- sanity checks added by libxmp hadn't set the error value, now they do.
  also fixes a wrong return value of -1 instead of FALSE.

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -712,10 +712,9 @@ static int compute_codewords(Codebook *c, uint8 *len, int n, uint32 *values)
    for (k=0; k < n; ++k) if (len[k] < NO_CODE) break;
 
    // sanity check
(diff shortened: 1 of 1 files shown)


sezero:
PING?

sezero:
PING?

sezero:
PING ?

sezero:
Friendly PING?

sezero:
PING ?

sezero:
PING ?

PR #163, opened 2020-04-04
Fix S3M ADPCM4 samples

AliceLR:
This branch fixes playback of S3Ms containing ADPCM4 samples using the same method libmodplug uses to detect them. 

Example S3M file that relies on ADPCM4 samples: [mm2flash.s3m.zip](https://github.com/cmatsuoka/libxmp/files/4432582/mm2flash.s3m.zip) (seems to be a resaved copy of [this](https://modarchive.org/index.php?request=view_by_moduleid&query=50304)).

### DIFF
--- a/src/loaders/s3m_load.c
+++ b/src/loaders/s3m_load.c
@@ -518,6 +518,7 @@ static int s3m_load(struct module_data *m, HIO_HANDLE * f, const int start)
 		struct xmp_instrument *xxi = &mod->xxi[i];
 		struct xmp_sample *xxs = &mod->xxs[i];
 		struct xmp_subinstrument *sub;
+		int load_sample_flags;
 
 		xxi->sub = calloc(sizeof(struct xmp_subinstrument), 1);
 		if (xxi->sub == NULL) {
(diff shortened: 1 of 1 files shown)


src/loaders/s3m_load.c
```diff
@@ -619,6 +620,11 @@ static int s3m_load(struct module_data *m, HIO_HANDLE * f, const int start)
 			xxs->flg |= XMP_SAMPLE_16BIT;
 		}
 
+		load_sample_flags = (sfh.ffi == 1) ? 0 : SAMPLE_FLAG_UNS;
+		if (sih.pack == 4) {
+			load_sample_flags = SAMPLE_FLAG_ADPCM;
```
NoSuck, line 625: Any reason not to bitwise OR this?  The logic in `libxmp_load_sample` does not indicate the flags here are mutually exclusive--and if they were, this block should be refactored to prevent the redundant check/assignment.

NoSuck:
Nice addition!

src/loaders/s3m_load.c
AliceLR: You're right, nice catch. The reason I implemented it like this is [that's how the copy of libmodplug 0.8.9.0 I used as a reference does it](https://github.com/AliceLR/megazeux/blob/master/contrib/libmodplug/src/load_s3m.cpp#L397), but there's no guarantee that's correct. I'll look into it more...

AliceLR: Update: from what I can gather [MOD Plugin](http://efr.worldsapart.net/modreadme.html) is the source of ADPCM4 samples in S3Ms (and [OpenMPT's handling of these](https://github.com/OpenMPT/openmpt/blob/master/soundlib/S3MTools.cpp#L117) attributes them to MOD Plugin), so libmodplug's behavior is probably correct. Probably also worth noting that in the same file I linked in the previous comment, the code for saving ADPCM4 samples [similarly overrides the U8 flag](https://github.com/AliceLR/megazeux/blob/master/contrib/libmodplug/src/load_s3m.cpp#L621).

sagamusix: If you still need a confirmation: Yes, ModPlug ADPCM is exclusively 8-bit mono. It must not be combined with any other flags because it is neither defined how to handle stereo ADPCM samples, nor is there any other algorithm defined apart from 4-to-8-bit expansion.

PR #172, opened 2020-09-26
Add LIBXMP_NO_DEPACKERS to load.c

AliceLR:
This separates the `LIBXMP_CORE_PLAYER` define in load.c to allow disabling the depackers separately from the core player. [My fork of libxmp for MegaZeux](https://github.com/AliceLR/megazeux/tree/cadf40a085339982f2098799ded261e90c96d1fe/contrib/libxmp) currently does a messier and more intrusive version of this to reduce the binary size for embedded platforms (and since MegaZeux has never supported depackers for its music files anyway). I'd like to get away from [these hacks](https://github.com/AliceLR/megazeux/tree/cadf40a085339982f2098799ded261e90c96d1fe/contrib/patches/libxmp) as much as possible, though, and I thought this patch might potentially have some use to someone else.

This currently only modifies load.c and doesn't do anything to the build system, so review and suggestions re: this are definitely welcome.


sezero:
> [....] I'd like to get away from [these hacks]

Off-topic:  Is this particular patch tested and/or submitted to upstream?
https://github.com/AliceLR/megazeux/blob/master/contrib/patches/libmodplug/02-libmodplug-0.8.9.0-fix-X86_Convert32To24.diff

AliceLR:
This patch does not appear to have been submitted upstream in its current form: https://github.com/Konstanty/libmodplug/blob/master/src/fastmix.cpp#L1870

I don't know enough about libmodplug's inner workings to know how well that has been tested, but the original code is obviously quite wrong. I traced this patch back to [a commit](https://github.com/adelva1984/megazeux/commit/3473694e800160c2e263ca4405d93cf56e2665ae) by @adelva1984 from 2007. Judging from the original diff this code was broken in a different way and was ["fixed" incorrectly](https://github.com/Konstanty/libmodplug/commit/e5210c08c857bb8a5910baebdff02274c697860c) after being reported. Regardless, if this fix has been part of MZX's libmodplug fork since 2007 I suspect it has been tested (MZX didn't switch to libxmp until 2017...).

AliceLR:
Also off-topic @sezero, but while I was diagnosing crashes in the devkitPro 3DS build of tremor, I encountered a crash that appears to be the same crash as was fixed by this patch: https://github.com/sezero/tremor/commit/a76e41f6ece93d10deac5f9ef3a84dce5b9c2a84

I noticed this patch was [committed to libvorbis](https://gitlab.xiph.org/xiph/vorbis/-/commit/128f0f812b39185d884c67c5f1c780b19aca34ac) but I haven't found a corresponding commit in the tremor repository and I'm not sure why. Is xiph just not porting fixes back to tremor anymore?

sezero:
> edit: I submitted a [pull request for this patch](https://github.com/Konstanty/libmodplug/pull/50) upstream.

Thanks.

sezero:
> I noticed this patch was [committed to libvorbis](https://gitlab.xiph.org/xiph/vorbis/-/commit/128f0f812b39185d884c67c5f1c780b19aca34ac) but I haven't found a corresponding commit in the tremor repository and I'm not sure why. Is xiph just not porting fixes back to tremor anymore?

I am not sure how to create pull requests for tremor: github doesn't mirror
tremor, and I don't know how to do that in the official gitlab xiph repo.
But, I did sent them patches through the mailing list, once:
http://lists.xiph.org/pipermail/vorbis-dev/2018-March/020502.html
http://lists.xiph.org/pipermail/vorbis-dev/2018-March/020503.html
.. Although no one cared, so I didn't try further for that particular patch.

sezero:
Here is a slightly modified version of your patch.
It touches the Makefile so that you can run after
configure: `make USE_DEPACKERS=0`  to disable the
depackers functionality. However, Makefile.vc and
Makefile.os2 needs regenerating, like:
`make USE_DEPACKERS=0 vc-prepare`
or: `make USE_DEPACKERS=0 os2-prepare`

It should be possible to modify configury for this,
but I didn't bother.

Use as you see fit.

[patch.txt](https://github.com/cmatsuoka/libxmp/files/5315833/patch.txt)

sezero:
Looks like my patch was incomplete: revised one attached. Changes:
- Ifdef surgery in src/loaders/sym_load.c (your patch missed this.
  otherwise libxmp_read_lzw_dynamic will be a missing symbol.)
- Touches Makefile.os2[.in] so it knows USE_DEPACKERS=0 make flag.
- Makefile.vc is still broken (don't know NMakefile syntax much.)

[patch2.txt](https://github.com/cmatsuoka/libxmp/files/5315924/patch2.txt)

AliceLR:
Thanks for looking at this! I went ahead and applied your patch. I can look into the changes necessary to get Makefile.vc to support this.

AliceLR:
I added what I (think) is the correct implementation of this for Makefile.vc. nmake errors with the inner directive unless `USE_DEPACKERS` is explicitly defined and I didn't see an equivalent to `ifeq` in the docs, hence the outer directive.

sezero:
Here is an incremental patch to make things more usable:
Changes OBJS definitions in makefiles so that Makefile.vc
and Makefile.os2 needn't be generated every time. Tested
build using Visual Studio 6, Open Watcom 1.9, and gcc.
Note: Review Makefile.in changes to make sure I did not
screw up.
[patch3.txt](https://github.com/cmatsuoka/libxmp/files/5317471/patch3.txt)

sezero:
>  applied your patch.

I am not comfortable with the changes I did to sym_load.c:  that format
seems to inherently need an lzw depacker to function.  A better solution
would be needed.

AliceLR:
I haven't managed to find many Digital Symphony mods to test with but what I've found so far (http://jerome.mathevet.free.fr/symphony.htm) does not use LZW and works with this patch. It doesn't seem like there are tests for this format, either.

sezero:
I made my own no-depackers branch in my fork with patches from
here and with sym_load.c retaining its LZW depacking (in a kind of
ugly way but whatever.)
https://github.com/sezero/libxmp/tree/depackers
Feel free to use / pull from it.

AliceLR:
I merged patch3.txt, and I think overall it does a better job of handling the Makefile changes than before. I'm conflicted on the handling of `sym_load.c` but I'd lean towards the handling in your linked branch being the better implementation of that. I'm curious if @cmatsuoka has any thoughts on how this should be handled, though.

sezero:
> I'm conflicted on the handling of `sym_load.c` but I'd lean towards the handling in your linked branch being the better implementation of that. I'm curious if @cmatsuoka has any thoughts on how this should be handled, though.

There is truly no point in crippling sym_load.c that way. I was the first to patch it
in order to resolve missing symbols, _and I did it lazily_ without thinking much.
I'd say that you really should do it my new way.

AliceLR:
Fair enough—I went ahead and updated this PR to match your updated patch.

sezero:
@AliceLR: If you can squash this into a single commit, I can merge it tomorrow.

AliceLR:
Ok, I squashed this branch and rebased it to the current master.

Issue #167, opened 2020-05-12
Memory leak and/or file descriptor leak in xmp_load_module_from_file


AliceLR:
The function `xmp_load_module_from_file` takes a file pointer and loads a module from that file pointer, but it does this in a way that I've found to cause portability issues. This function also necessarily causes memory leaks.

`xmp_load_module_from_file` creates a file pointer from the file descriptor of the provided file pointer [at line 581 of load.c](https://github.com/cmatsuoka/libxmp/blob/master/src/load.c#L581). This allocates a separate file pointer from the user-provided pointer given to the function, which is then closed by `hio_close` [at line 364 of hio.c](https://github.com/cmatsuoka/libxmp/blob/master/src/hio.c#L364). When this file pointer is closed, the file descriptor shared between it and the user file pointer is also closed, meaning the user file pointer can not be safely closed and thus not safely freed.

The intention of duplicating the file pointer in the first place seems to be to avoid closing the user provided file pointer, but that is explicitly not how fdopen works. Note this line from [the fdopen manpage](https://linux.die.net/man/3/fdopen):
> The file descriptor is not dup'ed, and will be closed when the stream created by fdopen() is closed.

In the case of MSVC/MinGW/MacOS, the file descriptor is released but the OS still seems to think it's leaking file descriptors, which will eventually break file IO. I think removing the usage of fdopen in `xmp_load_module_from_file` and making `HIO_HANDLE *`s created by `hio_open_file` not close the file pointer on `hio_close` would be closer to the intended behavior.

Spectere:
To elaborate a bit further on this issue, on macOS we determined that Apple's libc uses reference counting to track the number of used descriptors for both `fopen` and `fdopen` calls, causing the counter to increment by two with every file operation but only decrement by one when `fclose` is called. This would cause the system to think that it ran out of file descriptors after 252 operations (252 + stdin + stdout + stderr = 255, plus two more references would bring it to macOS's limit of 256).

[...]

PR #173, opened 2020-09-26
Fix leaks caused by fdopen in xmp_load_module_from_file

AliceLR:
This fixes file descriptor/memory leaks caused by incorrect usage of `fdopen()` as described in issue #167. The short version is that [this behavior is nonportable](https://man7.org/linux/man-pages/man3/fdopen.3.html) and tricks the operating system into thinking two file descriptors are open. The original `FILE *` and the `FILE *` opened by `fdopen()` can not be safely closed separately, meaning this necessarily leaks both a file descriptor (as far as the OS is concerned) and memory (the second `FILE` struct).

The way this fix is implemented means `xmp_load_module_from_file` will always close the provided `FILE *`, but it was essentially already doing this anyway.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -578,7 +578,7 @@ int xmp_load_module_from_file(xmp_context opaque, void *file, long size)
 	struct context_data *ctx = (struct context_data *)opaque;
 	struct module_data *m = &ctx->m;
 	HIO_HANDLE *h;
-	FILE *f = fdopen(fileno((FILE *)file), "rb");
+	FILE *f = (FILE *)file;
 	int ret;
 
 	if ((h = hio_open_file(f)) == NULL)
--- a/src/load.c
+++ b/src/load.c
@@ -581,8 +581,12 @@ int xmp_load_module_from_file(xmp_context opaque, void *file, long size)
 	FILE *f = (FILE *)file;
 	int ret;
 
-	if ((h = hio_open_file(f)) == NULL)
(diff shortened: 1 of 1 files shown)


sezero:
The second patch seems dangerous: What happens when the user (caller)
tries to fclose() his file when xmp_load_module_from_file() returns error?

AliceLR:
I didn't really like that either, but considering at line 603 xmp_load_module_from_file() can also return an error, I think it either has to unconditionally call fclose() (and be documented as doing such) or unconditionally not call fclose() (which requires changes to the way hio_close() works when an HIO_HANDLE is created with an existing FILE *).

I do think the fdopen was indeed intended to duplicate the FILE so the user can close their own copy, but unfortunately that's just not how fdopen() works in practice.

sezero:
Here is a test patch: what do you think?  If good, I can make a new PR for it.
[patch0.txt](https://github.com/cmatsuoka/libxmp/files/5323164/patch0.txt)

AliceLR:
I'd like @cmatsuoka's opinion on which direction this should go, but FWIW I definitely think your version is the correct way to do this.

sezero:
> I'd like @cmatsuoka's opinion [...]

He's been ignoring libxmp for quite some time, but we can't lose hope..

sezero:
I created https://github.com/cmatsuoka/libxmp/pull/178 with my approach to this issue.

sezero:
I merged #178 instead of this one.

PR #178, opened 2020-10-26
don't use fdopen() in hio_open_file, leave closing the file to caller:

sezero:
cf.: https://github.com/cmatsuoka/libxmp/pull/173

for internal uses of hio_open_file, invent new hio_open_file2() which
does allow the library to close the file.


sezero:
This closes: #167.

Note that this one and #173 are mutually exclusive.

AliceLR:
This is definitely a better approach to fixing this than #173 IMO. The function name `hio_open_file2` seems a little cheesy/non-descriptive but I don't have any better ideas.

sezero:
> The function name `hio_open_file2` seems a little cheesy/non-descriptive

Well, yeah.. :)

PR #179, opened 2020-10-28
Fix GCC stringop-truncation warnings.

AliceLR:
Fixes a bunch of annoying GCC 10(?) warnings caused by usage of strncpy. Some of the strncpy uses were legitimate but a few (like the Galaxy, PSM, and RTM loaders) were potentially cutting off terminating nulls.

Also fixes an operator precedence error from https://github.com/cmatsuoka/libxmp/pull/93 that GCC caught but somehow none of us noticed (sorry!).

### DIFF
--- a/src/loaders/gal4_load.c
+++ b/src/loaders/gal4_load.c
@@ -68,9 +68,10 @@ static int get_main(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	struct xmp_module *mod = &m->mod;
 	char buf[64];
 	int flags;
-	
+
 	hio_read(buf, 1, 64, f);
 	strncpy(mod->name, buf, 63);	/* ensure string terminator */
+	mod->name[63] = '\0';
 	libxmp_set_type(m, "Galaxy Music System 4.0");
 
 	flags = hio_read8(f);
--- a/src/loaders/gal5_load.c
+++ b/src/loaders/gal5_load.c
@@ -73,9 +73,10 @@ static int get_init(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	struct local_data *data = (struct local_data *)parm;
 	char buf[64];
 	int flags;
-	
+
 	hio_read(buf, 1, 64, f);
 	strncpy(mod->name, buf, 63);	/* ensure string terminator */
+	mod->name[63] = '\0';
 	libxmp_set_type(m, "Galaxy Music System 5.0");
 	flags = hio_read8(f);	/* bit 0: Amiga period */
 	if (~flags & 0x01)
--- a/src/loaders/psm_load.c
+++ b/src/loaders/psm_load.c
@@ -57,13 +57,14 @@ static int psm_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	uint32 p_ord, p_chn, p_pat, p_ins;
 	uint32 p_smp[64];
 	int type, ver /*, mode*/;
- 
+
 	LOAD_INIT();
 
 	hio_read32b(f);
 
 	hio_read(buf, 1, 60, f);
-	strncpy(mod->name, (char *)buf, 60);
+	memcpy(mod->name, (char *)buf, 59);
+	mod->name[59] = '\0';
 
 	type = hio_read8(f);		/* song type */
 	ver = hio_read8(f);		/* song version */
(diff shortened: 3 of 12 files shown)


src/loaders/gal4_load.c
```diff
@@ -68,9 +68,9 @@ static int get_main(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	struct xmp_module *mod = &m->mod;
 	char buf[64];
 	int flags;
-	
+
 	hio_read(buf, 1, 64, f);
-	strncpy(mod->name, buf, 63);	/* ensure string terminator */
+	snprintf(mod->name, XMP_NAME_SIZE, "%.*s", 63, buf);
```
sezero, line 73: 63 should be 64 here.

src/loaders/gal5_load.c
```diff
@@ -73,9 +73,9 @@ static int get_init(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	struct local_data *data = (struct local_data *)parm;
 	char buf[64];
 	int flags;
-	
+
 	hio_read(buf, 1, 64, f);
-	strncpy(mod->name, buf, 63);	/* ensure string terminator */
+	snprintf(mod->name, XMP_NAME_SIZE, "%.*s", 63, buf);
```
sezero, line 78: 63 should be 64 here.

src/loaders/gal4_load.c
sezero: OK confused.

AliceLR: I added the 63 to these mostly to avoid another annoying warning GCC emits re: snprintf truncation. It doesn't show up when I remove that so I'll just remove both of those.

sezero: OK: https://stackoverflow.com/questions/7899119
Never used precision myself..

AliceLR: I removed .* and 63 from both of the Galaxy loader snprintfs.

sezero:
Almost all of these come from the `strncpy(dst, src, size-1);` calls.
Do we still get the warnings if we add `dst[size-1] = 0;` after them?

AliceLR:
That does appear to fix the warnings as well (I hadn't tried it...). I don't think relying on strncpy is a good idea in general but if you'd prefer that to snprintf I can change it to that.

sezero:
If it resolves the case, then yes please.

AliceLR:
Ok, I updated this to just add null termination to most affected spots. A small handful still required replacing strncpy to get rid of the warning (I used memcpy+manual termination instead this time).

src/loaders/psm_load.c
```diff
@@ -57,13 +57,14 @@ static int psm_load(struct module_data *m, HIO_HANDLE *f, const int start)
...
- 
+
 	LOAD_INIT();
 
 	hio_read32b(f);
 
 	hio_read(buf, 1, 60, f);
-	strncpy(mod->name, (char *)buf, 60);
+	memcpy(mod->name, (char *)buf, 60);
+	mod->name[60] = '\0';
```
sezero, line 67: Do we want [60] or [59] here? psm.txt says that the last 60th byte should be a `^Z`

AliceLR: Nice catch, I was just going off of what the original code was doing. Those should both probably be 59.

AliceLR: I updated this to use 59 there instead.

src/loaders/rtm_load.c
```diff
@@ -226,8 +226,9 @@ static int rtm_load(struct module_data *m, HIO_HANDLE *f, const int start)
 			return -1;
 		}
 	}
-	
-	strncpy(mod->name, oh.name, 20);
+
```
sezero, line 229: I wonder what the reason was for 20 limit. The loader had it that way since it was first added.

AliceLR: I'm not sure what that was about either—maybe was supposed to be 0x20 or was mistaken for the tracker field length.

I just made a test RTM (albeit with Real Tracker 2.01) and it looks like 1) the name field definitely can fill the entire 32 bytes and also 2) those name fields aren't guaranteed to be null terminated.

For whatever reason GitHub won't let me attach it :/

sezero:
OK, merged.

PR #180, opened 2020-10-28
Check ProTracker song file sample names before opening them.

AliceLR:
This adds a check to the code that generates ProTracker song sample filenames to make sure there isn't anything suspicious like slashes or ".." (as requested in the comments for https://github.com/cmatsuoka/libxmp/pull/116). I'm reasonably sure this patch is correct but I don't actually have any ProTracker song files to test this with right now, so please don't merge it. edit: I have tested it with a ProTracker song file now, so if you think the behavior of this new function is reasonable, I am OK with it being merged.

edit: also found this being used in the MTP loader, which currently isn't used, but I added this check to it anyway.

### DIFF
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -258,6 +258,56 @@ int libxmp_test_name(uint8 *s, int n)
 	return 0;
 }
 
+int libxmp_copy_name_for_fopen(char *dest, const char *name, int n)
+{
+	int converted_colon = 0;
+	int i;
+
+	/* libxmp_copy_adjust, but make sure the filename won't do anything
+	 * malicious when given to fopen. This should only be used on song files.
+	 */
+	if (!strcmp(name, ".") || strstr(name, "..") ||
+	    name[0] == '\\' || name[0] == '/' || name[0] == ':')
+		return -1;
+
+
+	for (i = 0; i < n - 1; i++) {
+		uint8 t = name[i];
+		if (!t)
+			break;
+
+		/* Reject non-ASCII symbols as they have poorly defined behavior.
+		 */
+		if (t < 32 || t >= 0x7f)
+			return -1;
+
+		/* Reject anything resembling a Windows-style root path. Allow
+		 * converting a single : to / so things like ST-01:samplename
+		 * work. (Leave the : as-is on Amiga.)
+		 */
+		if (i > 0 && t == ':' && !converted_colon) {
+			uint8 t2 = name[i + 1];
+			if (!t2 || t2 == '/' || t2 == '\\')
+				return -1;
+
+			converted_colon = 1;
+#ifndef LIBXMP_AMIGA
+			dest[i] = '/';
+			continue;
+#endif
+		}
+
+		if (dest[i] == '\\') {
+			dest[i] = '/';
+			continue;
+		}
+
+		dest[i] = t;
+	}
+	dest[i] = '\0';
+	return 0;
+}
+
 /*
  * Honor Noisetracker effects:
  *
--- a/src/loaders/loader.h
+++ b/src/loaders/loader.h
@@ -37,6 +37,7 @@ char	*libxmp_instrument_name		(struct xmp_module *, int, uint8 *, int);
 struct xmp_sample* libxmp_realloc_samples(struct xmp_sample *, int *, int);
 
 char	*libxmp_copy_adjust		(char *, uint8 *, int);
+int	libxmp_copy_name_for_fopen	(char *, const char *, int);
 int	libxmp_test_name		(uint8 *, int);
 void	libxmp_read_title		(HIO_HANDLE *, char *, int);
 void	libxmp_set_xxh_defaults		(struct xmp_module *);
--- a/src/loaders/mod_load.c
+++ b/src/loaders/mod_load.c
@@ -392,8 +392,7 @@ static int mod_load(struct module_data *m, HIO_HANDLE *f, const int start)
     struct xmp_event *event;
     struct mod_header mh;
     uint8 mod_event[4];
-    char pathname[PATH_MAX] = "";
-    const char *x, *tracker = "";
+    const char *tracker = "";
     int detected = 0;
     char magic[8], idbuffer[32];
     int ptkloop = 0;			/* Protracker loop */
@@ -753,9 +752,6 @@ skip_test:
 
     /* Load samples */
 
-    if (m->filename && (x = strrchr(m->filename, '/')) != NULL)
-	strncpy(pathname, m->filename, x - m->filename);
-
     D_(D_INFO "Stored samples: %d", mod->smp);
 
     for (i = 0; i < mod->smp; i++) {
@@ -768,9 +764,18 @@ skip_test:
 
 	if (ptsong) {
 	    HIO_HANDLE *s;
-	    char sn[256];
-	    snprintf(sn, XMP_NAME_SIZE, "%s%s", pathname, mod->xxi[i].name);
-	
+	    char sn[PATH_MAX];
+	    char tmpname[32];
+	    const char *instname = mod->xxi[i].name;
+
+	    if (!instname[0] || !m->dirname)
+		continue;
+
+	    if (libxmp_copy_name_for_fopen(tmpname, instname, 32))
+		continue;
+
+	    snprintf(sn, PATH_MAX, "%s%s", m->dirname, tmpname);
+
 	    if ((s = hio_open(sn, "rb")) != NULL) {
 	        if (libxmp_load_sample(m, s, flags, &mod->xxs[i], NULL) < 0) {
 		    hio_close(s);
--- a/src/loaders/ssmt_load.c
+++ b/src/loaders/ssmt_load.c
@@ -181,16 +181,19 @@ static int mtp_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	D_(D_INFO "Instruments    : %d ", mod->ins);
 
 	for (i = 0; i < mod->ins; i++) {
+		struct xmp_instrument *xxi = &mod->xxi[i];
 		HIO_HANDLE *s;
 		char filename[1024];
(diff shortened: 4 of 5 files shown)


AliceLR:
Rebased to account for changes in commit 1b8783d4.

sezero:
What's in the way of applying this one?

AliceLR:
Mostly I just don't have an example ProTracker song file to test with yet. Shouldn't be hard to make one and I should do that already so I don't have to keep rebasing this...

AliceLR:
Test ProTracker song file, samples, and some hex edited variants:
[pugsy_title.zip](https://github.com/cmatsuoka/libxmp/files/5504600/pugsy_title.zip)

Notes now that I've actually tested this:

* Replacing the `m->filename` hack with using `m->dirname` had the extra unintended consequence that now I can load this file from (example) `/some/path/here/pugsy_title` and the samples will be correctly loaded. In a build of master the samples are not loaded in this situation and I have to change to the directory the song is in first to play it (probably due to the previous hack trimming the slash off).
* `pugsy_title` is a song created from an arbitrary .MOD I picked that was preinstalled in my Amiga VM.
* `pugsy_title_invalid` has some chars <32 and >=127 inserted into some of the sample filenames. This loads but those samples are silent because `libxmp_instrument_name` filters those chars out and replaces them with `.`.
* `pugsy_title_missing` similarly renames a sample and it doesn't load.
* `pugsy_title_parent` uses `..` for a filename and gets rejected entirely.
* `pugsy_title_passwd` uses `/etc/shadow` for a filename and gets rejected entirely.

AliceLR:
To clarify, I think this branch works now, but to be sure:

* Is rejecting the song file when `m->dirname` is NULL the correct thing to do (i.e. if the song file is loaded from `xmp_load_module_from_memory` or `xmp_load_module_from_file`), or should it just allow it to load without samples?
* Is rejecting the song file when slashes/etc are found in the sample name the correct thing to do, or should it just allow it to load without that sample?

sezero:
> Is rejecting the song file when `m->dirname` is NULL the correct thing to do (if the song file is loaded from `xmp_load_module_from_memory` or `xmp_load_module_from_file`), or should it just allow it to load without samples?

I'd say ignore silently and load w/o samples

>  Is rejecting the song file when slashes/etc are found in the sample name the correct thing to do, or should it just allow it to load without that sample?

'..' is bad. Slashes? Hm, say it is "samples/mysamps1" which may be perfectly legal, but absolute paths can be an issue. (This is a can of worms.)

What do others say?

AliceLR:
I hex edited the same song file to refer to its samples as `pugsy/[samplename]` and unfortunately, you're right. ProTracker (3.15 since 2.3d was giving me issues) loads it just fine as long as the samples are in that directory. I don't know if it's possible to actually create a song with sample names like this in ProTracker but they do load. I suspect sample names in the form `ST-##:samplename` would work too if I tried them. So, this patch needs to be revised...

[pugsy_title_samplesdir.zip](https://github.com/libxmp/libxmp/files/5504715/pugsy_title_samplesdir.zip)

sezero (commented):
One thing can be done is replacing `!strcmp(name, "..")` with `strstr(name,"..")`
in test_name_for_fopen: That one is almost universally considered harmful.

AliceLR:
Updated to change the following things:

* `libxmp_test_name_for_fopen` is now `libxmp_copy_name_for_fopen` because it requires slightly transforming the name in rare situations (see below).
* `libxmp_copy_name_for_fopen` now rejects `..` anywhere in the instrument name, but only rejects slashes at the start of the name or following the first `:`.
* `libxmp_copy_name_for_fopen` now converts the first `:` to a `/` if it is not followed by a `/` or `\\` (see below). This function also normalizes `\\` to `/` (just so this doesn't make a string with mixed slashes).
* When `m->dirname` is NULL the MOD loader will allow loading the song without samples. The MTP loader rejects the song altogether since the loader appears to rely on being able to load samples to work at all (not that this loader is currently active though...).
* When `libxmp_copy_name_for_fopen` fails, loading is allowed to continue. This is because checking for `..` anywhere in the sample name is more likely to turn up matches if the sample name was generated by `libxmp_instrument_name`, which changes anything <32 and >=127 to `.`.

The reason for the `:` → `/` edge case behavior is because as I suspected, ProTracker does allow loading instruments from `ST-01` and other disks. This allows these sample names to be handled if a copy of `ST-01` is present in the directory the song file is located in. I can't really think of a better sensible default handling for this in the scope of this patch. Example file with an instrument referencing `ST-01` (sample used included):

[bad_song.zip](https://github.com/libxmp/libxmp/files/5505391/bad_song.zip)

There's a second song file in this archive `bad_song2` that doesn't include the `ST-01:` prefix for the instrument. This fortunately does not default to chekcing for the sample in `ST-01` when loaded in ProTracker, so this can just be rejected when the sample is not in the current directory.

src/loaders/common.c
```diff
@@ -258,6 +258,51 @@ int libxmp_test_name(uint8 *s, int n)
...
+
+		/* Reject non-ASCII symbols as they have poorly defined behavior.
+		 */
+		if (t < 32 || t >= 0x7f)
+			return -1;
+
+		/* Reject anything resembling a Windows-style root path. Allow
+		 * converting a single : to / so things like ST-01:samplename work.
+		 */
+		if (t == ':' && !converted_colon) {
```
sezero, line 286: Do you want to check i==1 here? (i.e. you want to catch `c:\something`, yes?)

src/loaders/common.c
```diff
@@ -258,6 +258,51 @@ int libxmp_test_name(uint8 *s, int n)
...
+			if (name[i + 1] == '/' || name[i + 1] == '\\')
+				return -1;
+
+			converted_colon = 1;
+			dest[i] = '/';
```
sezero, line 291: Do we want to do this if we're running on amiga?

AliceLR: You're right, on Amiga this should probably be left as-is.

src/loaders/common.c
AliceLR: Not necessarily—some console SDKs (and maybe OSes) allow things like `sdcard:/`, `hd0:/`, etc. I think checking for the slash afterward (or maybe also `\0`) is a better way to detect that.

sezero: I see, OK.

src/loaders/common.c
AliceLR: Just to be sure, what exactly is the correct way to check for Amiga here?

sezero: > Just to be sure, what exactly is the correct way to check for Amiga here?

That wolud be a pain.. Some well-known compiler macros are `__amigaos4__`,
`__MORPHOS__`, `__amiga__`, `__AMIGA__`, `__AROS__`

In uhexen2, I have the following in arch_def.h (not all of those checks
may be needed or relevant...) :
```
#elif defined(__MORPHOS__) || defined(__AROS__) || defined(AMIGAOS)	|| \
      defined(__amigaos__) || defined(__amigaos4__) ||defined(__amigados__) || \
      defined(AMIGA) || defined(_AMIGA) || defined(__AMIGA__)

#   if !defined(PLATFORM_AMIGA)
#	define	PLATFORM_AMIGA		1
#   endif
```

src/loaders/common.c
AliceLR: I changed this to also check for `\0`, so I think it should be fine now.

src/loaders/common.c
AliceLR: OK, I added these checks pretty much verbatim to src/common.h (since presumably this macro might be useful for other things as well). On Amiga this should just copy the `:` as usual now (I don't have an Amiga toolchain to test with though right now).

sezero: I will make the AMIGA checks into a new PR now (out of scope of this one.)
I will be using LIBXMP_AMIGA as the macro.

sezero: See #203

sezero: I merged #203 so you can rebase and move forward with this

AliceLR: OK, I rebased this branch and changed it to use `LIBXMP_AMIGA`.

sezero:
@cmatsuoka: can you have a one last look at this one?

AliceLR:
Rebased to master and applied some misc. patches for the STM loader that need to be addressed from #186:

* Removed the `__native_client__` check.
* `stm_test` now tests the tracker name to make sure it contains ASCII chars.
* Sample filenames are now checked before attempting to fopen them.
* Sample filenames are now derived with `m->dirname` instead of the `m->filename` hack copied from the MOD loader.

sezero (approved):
This looks OK to me.

But a second pair of eyes will be useful: @cmatsuoka? or @sagamusix?

src/loaders/common.c
```diff
@@ -258,6 +258,55 @@ int libxmp_test_name(uint8 *s, int n)
...
 
+int libxmp_copy_name_for_fopen(char *dest, const char *name, int n)
+{
+	int converted_colon = 0;
+	int i;
+
+	/* libxmp_copy_adjust, but make sure the filename won't do anything
+	 * malicious if it's given to fopen. This should only be used on song files.
+	 */
+	if (!strcmp(name, ".") || strstr(name, "..") || name[0] == '\\' || name[0] == '/')
```
sagamusix, line 269: this should also check if the first character is a colon, because otherwise (due to colon version) you can still sneak in an absolute path.

AliceLR: I updated this to filter out anything with `:` as the first character (can't imagine that was ever valid even on Amiga) and also to limit the colon converstion to i>=1.

sezero: This looks good.  OK to merge now?

sezero:
I think this is good enough - merging.  If issues show up, we deal with them later.

PR #182, opened 2020-10-31
Fix test-dev FAR, STM, and mseek tests.

### LINKED ISSUE #181: test-dev issues.
There were a couple of bugs preventing test-dev from building that I fixed and doing so revealed some broken tests. I would just fix them and submit a patch since they're mostly related to fixed bugs but I can no longer get `gen_module_data.py` to work.

Branch with the test-dev fixes (fnmatch.o → xfnmatch.o, plus moving some variables inside of the test scope to avoid redefinitions, plus slightly improved test failure output): https://github.com/cmatsuoka/libxmp/compare/master...AliceLR:fix-test-dev

Failed tests:
```
test 12: test_read_mem_hio: FAILED at test_read_mem_hio.c:99: hio_seek
test 76: test_loader_stm: FAILED at util.c:200: number of subinstruments
test 82: test_loader_far: FAILED at util.c:295: track data
```

Patches that likely caused these:
https://github.com/cmatsuoka/libxmp/pull/157
https://github.com/cmatsuoka/libxmp/pull/93
https://github.com/cmatsuoka/libxmp/pull/176

The .STM and .FAR tests should just be a matter of regenerating the module data hashes. The hio_seek fail might be resolved by the patch for #160 or that check might just need to be changed.

AliceLR:
This is a squashed version of this branch: https://github.com/cmatsuoka/libxmp/compare/master...AliceLR:fix-test-dev.

* Fixes the mseek test that stopped working after 0db60b18.
* Fixes a FAR test that stopped working after 7d6d8243. The diff is so big on that file because it affected the loaded volume values in pretty much every pattern and the data file stores (pattern count)\*(channel count)=30\*16=480 track hashes(!).
* Fixes an STM test that stopped working after ca111900.
* Adds a C implementation of gen_module_data.py because between that file and libxmp-python, something seems to be broken (even when I hack both so they use python2 instead of /usr/bin/python). This puts the code to generate the data files right next to the code that verifies them so this may be a better solution anyway.
* There was a multiple definition bug caused by two tests using the same variable. That might have only affected MSYS2 or something else related to my setup but I fixed it anyway.
* The failed test output lines weren't printing a newline, the file they failed in, or anything that could be consistently searched to find where the broken tests were, so I tweaked that.

Fixes #181.


AliceLR:
Looks like the lack of "**fail**" output is actually only a problem in MSYS2, will modify my changes to that.

sezero:
Merged, thanks.

PR #184, opened 2020-10-31
Revert -Wstringop-truncation fix in libxmp_iff_register.

AliceLR:
Reverts a faulty warning fix from commit c7dccce6. This was the rare legitimate usage of strncpy and I didn't realize this function gets provided id strings shorter than 4 bytes (happens in the MDL loader...).

### DIFF
--- a/src/loaders/iff.c
+++ b/src/loaders/iff.c
@@ -170,7 +170,7 @@ int libxmp_iff_register(iff_handle opaque, const char *id,
 	if (f == NULL)
 		return -1;
 
-	memcpy(f->id, id, 4);
+	strncpy(f->id, id, 4);
 	f->loader = loader;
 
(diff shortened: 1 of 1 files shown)


sezero:
Just lose the pragmas, older gcc emits a warning for that instead:
```
src/loaders/iff.c:163: warning: expected [error|warning|ignored] after ‘#pragma GCC diagnostic’
src/loaders/iff.c:164: warning: unknown option after ‘#pragma GCC diagnostic’ kind
src/loaders/iff.c:184: warning: expected [error|warning|ignored] after ‘#pragma GCC diagnostic’
```

AliceLR:
Ok, I dropped the pragmas. I assumed older GCCs would just ignore those, sorry...

Issue #166, opened 2020-04-13
Digibooster module not recognized


softwarefailure:
Check out the attached Digibooster module. XMP fails to open it.
[Krashan - The Call of Quantum Field.zip](https://github.com/cmatsuoka/libxmp/files/4468714/Krashan.-.The.Call.of.Quantum.Field.zip)

AliceLR:
This happens because the IFF sample chunk in this module is larger than `IFF_MAX_CHUNK_SIZE` (4 MiB). If I raise this value to 16 MiB it loads fine. This bound seems arbitrary so I wonder if it can safely just be bumped up.
```
load_module [src/load.c:397] load format: DigiBooster Pro
dbm_load [src/loaders/dbm_load.c:389] Module title: "The Call of Quantum Field"
dbm_load [src/loaders/dbm_load.c:389] Module type: DigiBooster Pro 3.00 DBM0
iff_chunk [src/loaders/iff.c:75] chunk id size: 4
iff_chunk [src/loaders/iff.c:80] chunk id: [INFO]
iff_chunk [src/loaders/iff.c:99] size: 10
iff_process [src/loaders/iff.c:51] Load IFF chunk INFO (10) @68
iff_chunk [src/loaders/iff.c:75] chunk id size: 4
iff_chunk [src/loaders/iff.c:80] chunk id: [INST]
iff_chunk [src/loaders/iff.c:99] size: 500
iff_process [src/loaders/iff.c:51] Load IFF chunk INST (500) @86
get_inst [src/loaders/dbm_load.c:153] Instruments: 10
get_inst [src/loaders/dbm_load.c:180] [ 0] Krashan 2013-11-12             #00 V40 P80 44100
get_inst [src/loaders/dbm_load.c:180] [ 1] Tracked with DigiBooster 3.0   #01 V28 P6e 44100
get_inst [src/loaders/dbm_load.c:180] [ 2] on MorphOS                     #02 V0a P49 44100
get_inst [src/loaders/dbm_load.c:180] [ 3] Playtime 3:46                  #03 V40 P80 44100
get_inst [src/loaders/dbm_load.c:180] [ 4]                                #04 V2e P9d 44100
get_inst [src/loaders/dbm_load.c:180] [ 5]                                #05 V21 P54 44100
get_inst [src/loaders/dbm_load.c:180] [ 6] Greetings for #ppa on freenode #06 V24 Pac 44100
[...]

PR #185, opened 2020-11-03
Fix MEDs with pattern lengths >256, MED instrument corruption.

AliceLR:
This fixes some MED bugs I found while looking into why the module "New Dimension" does not load.

* This MED contains patterns with lengths greater than 256. MMD1+ patterns use a word to store the pattern length but I'm not sure what a realistic bound for this is (edit: it's 3200 judging by the docs). The fix for this in this patch should be revised BEFORE this is merged because the 256 row sanity check assumption is true for most other formats and 32768 is an arbitrary number I chose for testing. Do you have any recommendations for a better way to do this @sezero?
* The MMD loaders were writing the OctaMED instrument name (40 chars long) directly into the instrument name (32 chars long), corrupting the instrument volume and sample count fields.
* Several spots were not adding `start` to calculated offsets when seeking. I don't know if this was actually breaking anything but I fixed it anyway.

[new dimension.med.zip](https://github.com/cmatsuoka/libxmp/files/5484184/new.dimension.med.zip)

### DIFF
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -138,7 +138,27 @@ int libxmp_alloc_tracks_in_pattern(struct xmp_module *mod, int num)
 int libxmp_alloc_pattern_tracks(struct xmp_module *mod, int num, int rows)
 {
 	/* Sanity check */
-	if (rows < 0 || rows > 256)
+	if (rows <= 0 || rows > 256)
+		return -1;
+
+	if (libxmp_alloc_pattern(mod, num) < 0)
+		return -1;
+
+	mod->xxp[num]->rows = rows;
+
+	if (libxmp_alloc_tracks_in_pattern(mod, num) < 0)
+		return -1;
+
+	return 0;
+}
+
+/* Some formats explicitly allow more than 256 rows (e.g. OctaMED). This function
+ * allows those formats to work without disrupting the sanity check for other formats.
+ */
+int libxmp_alloc_pattern_tracks_long(struct xmp_module *mod, int num, int rows)
+{
+	/* Sanity check */
+	if (rows <= 0 || rows > 32768)
 		return -1;
 
 	if (libxmp_alloc_pattern(mod, num) < 0)
--- a/src/loaders/loader.h
+++ b/src/loaders/loader.h
@@ -32,6 +32,7 @@ int	libxmp_alloc_pattern		(struct xmp_module *, int);
 int	libxmp_alloc_track		(struct xmp_module *, int, int);
 int	libxmp_alloc_tracks_in_pattern	(struct xmp_module *, int);
 int	libxmp_alloc_pattern_tracks	(struct xmp_module *, int, int);
+int	libxmp_alloc_pattern_tracks_long(struct xmp_module *, int, int);
 char	*libxmp_instrument_name		(struct xmp_module *, int, uint8 *, int);
 struct xmp_sample* libxmp_realloc_samples(struct xmp_sample *, int *, int);
 
--- a/src/loaders/mmd1_load.c
+++ b/src/loaders/mmd1_load.c
@@ -92,6 +92,7 @@ static int mmd1_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	int annotxt_offset;
 	int pos;
 	int bpm_on, bpmlen, med_8ch, hexvol;
+	char name[40];
 
 	LOAD_INIT();
 
@@ -348,7 +349,11 @@ static int mmd1_load(struct module_data *m, HIO_HANDLE *f, const int start)
 			block.lines = hio_read8(f);
 		}
 
-		if (libxmp_alloc_pattern_tracks(mod, i, block.lines + 1) < 0)
+		/* Sanity check--Amiga OctaMED files have an upper bound of 3200 lines per block. */
+		if (block.lines + 1 > 3200)
+			return -1;
+
+		if (libxmp_alloc_pattern_tracks_long(mod, i, block.lines + 1) < 0)
 			return -1;
 
 		if (ver > 0) {		/* MMD1 */
@@ -451,11 +456,13 @@ static int mmd1_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		if (expdata_offset && i < expdata.i_ext_entries) {
 			struct xmp_instrument *xxi = &mod->xxi[i];
 			int offset = iinfo_offset + i * expdata.i_ext_entrsz;
-			
-			if (offset < 0 || hio_seek(f, offset, SEEK_SET) < 0) {
+
+			if (offset < 0 || hio_seek(f, start + offset, SEEK_SET) < 0) {
 				return -1;
 			}
-			hio_read(xxi->name, 40, 1, f);
+			hio_read(name, 40, 1, f);
+			strncpy(xxi->name, name, 32);
+			xxi->name[31] = '\0';
 		}
 
 		D_(D_INFO "[%2x] %-40.40s %d", i, mod->xxi[i].name, instr.type);
@@ -464,7 +471,7 @@ static int mmd1_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		if (expdata_offset && i < expdata.s_ext_entries) {
 			int offset = expsmp_offset + i * expdata.s_ext_entrsz;
 
-			if (offset < 0 || hio_seek(f, offset, SEEK_SET) < 0) {
+			if (offset < 0 || hio_seek(f, start + offset, SEEK_SET) < 0) {
 				return -1;
 			}
 			exp_smp.hold = hio_read8(f);
--- a/src/loaders/mmd3_load.c
+++ b/src/loaders/mmd3_load.c
@@ -92,6 +92,7 @@ static int mmd3_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	int playseq_offset;
 	int pos;
 	int bpm_on, bpmlen, med_8ch, hexvol;
+	char name[40];
 
 	LOAD_INIT();
 
@@ -349,7 +350,13 @@ static int mmd3_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		block.lines = hio_read16b(f);
 		hio_read32b(f);
 
-		if (libxmp_alloc_pattern_tracks(mod, i, block.lines + 1) < 0)
+		/* Sanity check--Amiga OctaMED files have an upper bound of 3200 lines per block,
+		 * but MED Soundstudio for Windows allows up to 9999 lines.
+		  */
+		if (block.lines + 1 > 9999)
+			return -1;
+
+		if (libxmp_alloc_pattern_tracks_long(mod, i, block.lines + 1) < 0)
 			return -1;
 
 		for (j = 0; j < mod->xxp[i]->rows; j++) {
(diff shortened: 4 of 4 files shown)


sezero:
> This fixes some MED bugs I found while looking into why the module "New Dimension" does not load.
> 
>     * This MED contains patterns with lengths greater than 256. MMD1+ patterns use a word to store the pattern length but I'm not sure what a realistic bound for this is. The fix for this in this patch should be revised BEFORE this is merged because the 256 row sanity check assumption is true for most other formats and 32768 is an arbitrary number I chose for testing. Do you have any recommendations for a better way to do this @sezero?

Yes, this is a problem..  I can think of only ugly workarounds like
libxmp_alloc_pattern_tracks2, for now.  I'll think.

>     * Several spots were not adding `start` to calculated offsets when seeking. I don't know if this was actually breaking anything but I fixed it anyway.

As far as I can see, the public module load procedures all call
load_module() with 0 start parameter, so I _think_ it shouldn't
break anything.

sezero:
> > This fixes some MED bugs I found while looking into why the module "New Dimension" does not load.
> > ```
> > * This MED contains patterns with lengths greater than 256. MMD1+ patterns use a word to store the pattern length but I'm not sure what a realistic bound for this is. The fix for this in this patch should be revised BEFORE this is merged because the 256 row sanity check assumption is true for most other formats and 32768 is an arbitrary number I chose for testing. Do you have any recommendations for a better way to do this @sezero?
> > ```
> 
> Yes, this is a problem.. I can think of only ugly workarounds like
> libxmp_alloc_pattern_tracks2, for now. I'll think.

In mmdX_load.c, we can cheat by calling libxmp_alloc_pattern_tracks()
with row==256, then manually set module rows, e.g. something like:
```
-               if (libxmp_alloc_pattern_tracks(mod, i, block.lines + 1) < 0)
+               /* MMD? can have rows > 256 */
+               if (libxmp_alloc_pattern_tracks(mod, i, 256) < 0)
                        return -1;
+               mod->xxp[i]->rows = block.lines + 1;
```

AliceLR:
> In mmdX_load.c, we can cheat by calling libxmp_alloc_pattern_tracks()
with row==256, then manually set module rows, e.g. something like:

I don't think that would work, because then the tracks would also all have to be manually reallocated too. Aside from a new function my only thought is providing a new param to that function with the rows upper bound for each loader.

sezero:
> I don't think that would work, because then the tracks would also all have to be manually reallocated too.

Yeah missed that

> Aside from a new function my only thought is providing a new param to that function with the rows upper bound for each loader.

For that mmd (and amf as you ref'ed below) would be the only two users needing that new param
Don't like that either

> edit: there's also something the DSMI AMF loader does, which is to individually allocate the tracks. This bypasses the sanity check.

That's an option, aside from a new func.

sezero:
Off-topic: The `rows < 0`check in libxmp_alloc_pattern_tracks()
seems wrong, it should be `rows <= 0`: see libxmp_alloc_track()

AliceLR:
> Off-topic: The `rows < 0`check in libxmp_alloc_pattern_tracks()
> seems wrong, it should be `rows <= 0`: see libxmp_alloc_track()

Yeah, I'm not sure what's up with that. A pattern with length 0 does not make any sense.

> For that mmd (and amf as you ref'ed below) would be the only two users needing that new param
Don't like that either

That actually got me wondering what even relies on that upper bound check at all, since the vast majority of these formats use 64 row patterns.

```
$ grep -nrI libxmp_alloc_pattern_tracks
common.c:138:int libxmp_alloc_pattern_tracks(struct xmp_module *mod, int num, int rows)
loader.h:34:int libxmp_alloc_pattern_tracks     (struct xmp_module *, int, int);
```

Loaders with variable length patterns with a >=u16 range:
```
dbm_load.c:206:         if (libxmp_alloc_pattern_tracks(mod, i, hio_read16b(f)) < 0)
dmf_load.c:215:         if (libxmp_alloc_pattern_tracks(mod, i, hio_read16l(f)) < 0)
gtk_load.c:248:         if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
masi_load.c:267:        if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
mmd1_load.c:352:                if (libxmp_alloc_pattern_tracks(mod, i, block.lines + 1) < 0)
mmd3_load.c:353:                if (libxmp_alloc_pattern_tracks(mod, i, block.lines + 1) < 0)
okt_load.c:245: if (libxmp_alloc_pattern_tracks(mod, data->pattern, rows) < 0)
```

Loaders with variable length patterns with a >=u16 range BUT are bounded to 256 rows max with an internal sanity check:
```
imf_load.c:338: if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
rtm_load.c:281:         if (libxmp_alloc_pattern_tracks(mod, i, rp.nrows) < 0)
xm_load.c:93:   if (libxmp_alloc_pattern_tracks(mod, num, r) < 0) {
```

Loaders with variable length patterns restricted to a u8 range:
```
arch_load.c:315:        if (libxmp_alloc_pattern_tracks(mod, i, data->rows[i]) < 0)
dtt_load.c:144:         if (libxmp_alloc_pattern_tracks(mod, i, plen[i]) < 0)
emod_load.c:124:                if (libxmp_alloc_pattern_tracks(mod, i, hio_read8(f) + 1) < 0)
gal4_load.c:161:        if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
gal4_load.c:444:                        if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0) {
gal5_load.c:150:        if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
gal5_load.c:368:                        if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0) {
med4_load.c:410:                if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
```

Loaders with variable length patterns restricted to a u8 range BUT are bounded to 64 rows max with an internal sanity check:
```
psm_load.c:179:         if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
```

Loaders referencing this that use 64-row patterns (NOTE: GDM should be able to go up to 256 theoretically but in practice that doesn't seem to have ever been used):
```
669_load.c:201: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
abk_load.c:591:        if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
alm_load.c:129: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
amd_load.c:125:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
asylum_load.c:127:              if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
coco_load.c:264:                if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
digi_load.c:186:        if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
dt_load.c:228:          if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
dt_load.c:324:                  if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0) {
flt_load.c:428:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
fnk_load.c:217: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
gdm_load.c:344:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
hmn_load.c:252:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
hsc_load.c:163: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
ims_load.c:246: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
med2_load.c:149:                if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
med3_load.c:348:                if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
mfp_load.c:186:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
mod_load.c:599: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
no_load.c:213:          if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
polly_load.c:151:               if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0) {
pt3_load.c:255:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
ptm_load.c:254:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
pw_load.c:198:          if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
rad_load.c:179:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0) {
s3m_load.c:451:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
sfx_load.c:190:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
ssmt_load.c:120:                if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
stc_load.c:175:         if (libxmp_alloc_pattern_tracks(mod, dest, 64) < 0)
stim_load.c:127:                if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
stm_load.c:230:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
stx_load.c:291:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
st_load.c:384:          if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
tcb_load.c:118:         if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
ult_load.c:258: if (libxmp_alloc_pattern_tracks(mod, i, 64) < 0)
```

So not very many formats are affected by changing the upper bound in `libxmp_alloc_pattern_tracks` and if any of them are, there is precedent for having more format-specific sanity checks in the affected loaders. So we should probably figure out what the DigiBooster Pro (dbm_load.c), X-Tracker (dmf_load.c), Graoumf Tracker (gtk_load.c), Epic MegaGames MASI, OctaMED, and Oktalyzer formats actually do re: upper bounds on rows next.

I can start by linking [this bit of the MMD1 documentation](https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/mmd1_r1.txt#L307) where it says the realistic upper bound on rows in older OctaMED versions is 3200 and [this bit of the MMD3 documentation](https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/mmd3_r6.txt#L449) that confirms this limit likely persisted to Soundstudio 1.03.

AliceLR:
Checked the documentation for the rest of those formats:

* DigiBooster Pro documentation doesn't indicate an upper row limit: https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/DigiBoosterPro.txt
* X-Tracker documentation doesn't indicate an upper row limit: https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/dmf-fileformat.txt
* Old Graomf Tracker modules appear to have an intended limit of 256 lines per pattern, not clear about the IFF variant: https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/graoumf.txt#L387
* MASI documentation is not complete and does not indicate an upper bound: https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/psm-format.txt
* Oktalyzer documentation doesn't indicate an upper bound:
https://github.com/cmatsuoka/libxmp/blob/master/docs/formats/oktafmt.txt

sezero:
Sent @cmatsuoka a private email to have a look at this PR about rows upper bound sanity check.

sezero:
> Sent @cmatsuoka a private email to have a look at this PR about rows upper bound sanity check.

He still hasn't responded.

To be least intrusive, I can only see either a new function or what AMF loader does as the solution for rows upper bound.

AliceLR:
OK. I think a new function (`libxmp_alloc_pattern_tracks_long` maybe) would result in the least copypaste of those two options. I still want to gather some data from the ModLand DBM collection to get a better idea of what kind of upper limit that function would have (since I need to gather some stats for #166 anyway), but I suspect 32768 would be fine if it's only being used by 6 formats.

sezero:
I gave conflicts to your patch in git master (sorry.) Can be resolved by changing
`&xxi->name` to `xxi->name` in the patch.

AliceLR:
Good to know, I will update this patch momentarily. I don't know why those &s were even there to begin with.

AliceLR:
From checking the ModLand collection I found an entire 0 DigiBooster Pro modules containing a pattern with more than 256 rows.

AliceLR:
I have revised this patch to implement a new function with a relaxed sanity check that is currently only used for allocating MED patterns, and I added MED-specific pattern length checks. For Amiga MEDs the 3200 row limit seems accurate but Windows MED Soundstudio allows blocks of length 9999, hence the different limit in the MMD2/3 loader.

sezero:
Thanks, this is in.

PR #186, opened 2020-11-04
New STM loader, based on disassembled ST2.

viiri:
Copy of #116 (because of unsuccessful rebase).
Support all format versions of modules and songs except 1.00 because of different encoding schema.
Based on behavior of disassemled original ST2 loader: https://pastebin.com/TDQbyBLW


sezero:
> Copy of #116

- Including the description (intent) of this PR here would be helpful, instead of a link to previous PR.
- Have all of the reviews been taken into account in this new version?
- Please squash the two commits into one.

AliceLR:
I gave this a quick scan earlier and it looked identical to #116, just rebased to the current master. So the complaint that #180 addresses in the MOD loader still stands at least (I should really find PT song files to test...) but that method of deriving sample names is at least consistent with the current MOD loader.

Other issues from the original PR: the request regarding version checks was addressed and the EOF byte issue was a problem with the original files on ModLand. Don't know about the tracker string checks. I can give this a closer look when I get a chance (sorry I put this off).

AliceLR:
@viiri are there STM 1.10, 2.00, 2.10, and 2.20 test files? Everything I've found so far is 2.21 and the upload URL on the old pull request has expired...

src/loaders/stm_load.c
```diff
@@ -71,15 +95,17 @@ static int stm_test(HIO_HANDLE * f, char *t, const int start)
 	hio_seek(f, start + 20, SEEK_SET);
 	if (hio_read(buf, 1, 8, f) < 8)
 		return -1;
-	if (memcmp(buf, "!Scream!", 8) && memcmp(buf, "BMOD2STM", 8) && memcmp(buf, "WUZAMOD!", 8))
-		return -1;
```
AliceLR, line 75: Ok, I see what Saga Musix meant by removing validation on the tracker name check—it's essentially this format's magic string. I also think removing this entirely is a bad idea. Aside from `PCSTV` and `SWavePro` and any case sensitivity issues is there anything else missing from this check? I think a list of correct strings here is a better idea than allowing anything. I'm not convinced just a check to make sure these are ASCII chars (like Saga Musix suggested) is a good way to handle this either, since, for example, Soundtracker 15-instrument .MODs will have ASCII chars here too.

AliceLR (changes requested):

sezero:
I can see that none of the concerns noted in the previous PR are
addressed here: As it is, this is not acceptable. Here is what we
need:

- Commit title: A self-explaining one-liner, that describes the patch.
  In your case, maybe `Add support for more versions of  STM files",
  or something.

- Commit message: Give details about the patch, and give reasons for
  changes where necessary. In your case something like: "Support for
  version x, y and z of STM files added." Pointing to documentation
  about format version differences is recommended. Also pointing to
  some documentation about `spd` calculation differences across the
  versions is recommended.

- You removed the magic string checks "!Scream!", "BMOD2STM", and
  "WUZAMOD!": If you have a very good reason, explain it, otherwise
  put them back. (See concerns raised by @sagamusix in previous PR
  and by @AliceLR in this version.)

- Give working links to STM files that this PR adds support for.

Others can add to this list, I think: @sagamusix , @AliceLR: please
do so.

AliceLR:
I think it's worth noting that 1) one of the concerns in the original PR was not an issue with the PR but an issue with files Saga Musix was testing with; 2) the version checks were fixed as requested; and 3) there were test files in #93 and #116 but the links expired (so a reupload would be appreciated if anyone still has those). I would also be fine leaving the sample loading as-is in this PR and I can just rebase #180 and fix it there.  But yes, I think that list of concerns is accurate.

src/loaders/stm_load.c
sagamusix: We don't need to bring the format check back as-is. With thorough enough header validation (see `ValidateHeader` in OpenMPT's Load_stm.cpp) you can be reasonably sure that something is really an STM file. There is more validation happening outside that function too in OpenMPT, e.g. if the order list contains any item above 63 that is not 99 or 255, the file is rejected as well.

viiri: STM has no format magic string, just a field that contains some random text... or doesn't.

viiri:
> 
> 
> I can see that none of the concerns noted in the previous PR are
> addressed here: As it is, this is not acceptable. Here is what we
> need:
> 
>     * Commit title: A self-explaining one-liner, that describes the patch.
>       In your case, maybe `Add support for more versions of  STM files",
>       or something.
> 
>     * Commit message: Give details about the patch, and give reasons for
>       changes where necessary. In your case something like: "Support for
>       version x, y and z of STM files added." Pointing to documentation
>       about format version differences is recommended. Also pointing to
>       some documentation about `spd` calculation differences across the
>       versions is recommended.
> 
>     * You removed the magic string checks "!Scream!", "BMOD2STM", and
>       "WUZAMOD!": If you have a very good reason, explain it, otherwise
>       put them back. (See concerns raised by @sagamusix in previous PR
>       and by @AliceLR in this version.)
> 
>     * Give working links to STM files that this PR adds support for.
> 
> 
> Others can add to this list, I think: @sagamusix , @AliceLR: please
> do so.

Sorry, but there is only one incomplete document about STM format internals - tech.doc from original tracker distribution:
https://pastebin.com/GHR5tJQA
Missing pieces of puzzle could be found by disassemling original binaries, so I did it:
https://pastebin.com/TDQbyBLW
Here is simplified (STM v1 support is missing) loader in my st2play project:
https://github.com/viiri/st2play/blob/master/stmload.c
and here is similar loader from st23play by 8bitbubsy:
https://pastebin.com/ifXSCZ71
as I know he has access to original source code and, I think, he can give advice.

As I said earlier, there is no magic string, just 20 bytes of data completely ignored by original loader.

https://files.fm/u/k8myhnqa - FCSLIDE1.STS, example of STM v1.10 format song. It's ripped from FCSLIDE1 demo.

src/loaders/stm_load.c
sagamusix: While it may be technically possible to have completely random strings, only a handful of strings are found in the wild, that's what matters. It's not like most module players are unable to play a majority of STM modules because of this. While it may not be important for a single-purpose program like st2play to do thorough validation of this string, rejecting any non-ASCII string is a really valuable for multi-format libraries like libxmp that are also used in combination with other media playing libraries. But to move forward, I think we can merge this and have a follow-up PR to address these issues later.

viiri: Yes, STM format has very few traits, so I agree with idea of additinal string validation. It will reduce false positives without hardcoding.

sezero:
I'm leaving this to @cmatsuoka and others

src/loaders/stm_load.c
```diff
@@ -20,6 +20,11 @@
  * THE SOFTWARE.
  */
 
+#ifdef __native_client__
+#include <sys/syslimits.h>
```
cmatsuoka, line 24: Do we support native client? Wasn't it deprecated in Chrome some time ago in favor of wasm?

cmatsuoka (commented):
Thanks. It's always interesting to support these lesser known variants of the module formats, maybe we can find more files in this format in the future (I tested fcslide1 but it's not very exciting without the sample files). Regarding file magic checking, we can improve robustness of detection later if we get false positives.

But we need to address the conflict with master before we can land this, and also adding some comments to the source code explaining where the v1 info came from would be very useful (so we -- or someone else -- won't be confused when refactoring this code in the future).

cmatsuoka:
Just to be clear, I think that a comment like "v1 format header based on disassembled ST2" is enough.

cmatsuoka:
Also basic CI tests are fixed in master, so merging master into this branch (and solving the conflicts) should make this PR ready to land. We can add tests for v1 loading in a follow-up.

AliceLR (approved):
I'm in support of just merging this as-is and fixing up the minor leftover tasks (tracker name validation, sample name checks per #180, a test) in a separate patch.

cmatsuoka:
I agree. I'll resolve the conflict and make the branch ready for landing, could you review my changes afterwards?

cmatsuoka:
Hmm interesting, we have a test error with clang but not with gcc, and valgrind doesn't report any violation.

AliceLR:
The changes look good but the new test fails on my end at track 120...

src/loaders/stm_load.c
AliceLR: I think that is correct, it is deprecated.

AliceLR:
The test works on my end now. Not sure why it's still failing the Travis check.

cmatsuoka:
Are you testing on a Mac? The travis tests are failing only on osx and valgrind doesn't report anything strange in my Linux machine.

AliceLR:
The test passed just fine on my iMac (Core Duo, Snow Leopard) using the Xcode GCC 4.2.1 but fails with "Bus error" when compiled with Clang 10 (MacPorts). Looking into it now.

AliceLR:
The test module instrument count reports 32 instruments but the instruments array in the loader is only size 31. @viiri should this array be extended or is it reporting one more than the actual number of instruments? What is the maximum number of instruments that should be expected in v1 files?

sezero:
> The test module instrument count reports 32 instruments but the instruments array in the loader is only size 31. @viiri should this array be extended or is it reporting one more than the actual number of instruments? What is the maximum number of instruments that should be expected in v1 files?

Whatever the max number of ins expectation, a sanity check for mod->ins should
be added before the `for (i = 0; i < mod->ins; i++)` loop.

src/loaders/stm_load.c
viiri: I forgot to remove these lines from previous 3 years old patch.

cmatsuoka:
Should we assume a maximum number of 32 instruments for v1 for now? Given our example file and how other parameters are locked into fixed values I wouldn't be surprised if v1 has always 32 instruments. Also it seems that ST2 strips off the last instrument when loading/converting, but we can keep the last one. I'll propose a change increasing the number of instruments to 32, let's see if it fixes the problem.

It's also interesting that this problem was not detected when testing on Linux, I'll try ElectricFence later -- if it works we can add it to the CI tests.

sezero:
> It's also interesting that this problem was not detected when testing on Linux, I'll try ElectricFence later -- if it works we can add it to the CI tests.

It actually _is_ detected, but only if you build without optimizations.
Here is the result with such a build running under valgrind (of course
with your last patch removed to see the issue) :
```
==18089== Memcheck, a memory error detector
==18089== Copyright (C) 2002-2012, and GNU GPL'd, by Julian Seward et al.
==18089== Using Valgrind-3.8.1 and LibVEX; rerun with -h for copyright info
==18089== Command: ./xmp -dnull fcslide1.sts
==18089== 
Extended Module Player 4.1.0
Copyright (C) 1996-2016 Claudio Matsuoka and Hipolito Carraro Jr
Using null output
Mixer set to 44100 Hz, 16bit, cubic spline interpolated stereo
Press 'h' for help

Loading fcslide1.sts (1 of 1)
xmp_load_module [src/load.c:494] path = fcslide1.sts
xmp_load_module [src/load.c:512] decrunch
decrunch [src/load.c:188] Not packed
load_module [src/load.c:388] load
load_module [src/load.c:397] test Fast Tracker II
load_module [src/load.c:397] test Amiga Protracker/Compatible
load_module [src/load.c:397] test Startrekker
load_module [src/load.c:397] test Soundtracker
load_module [src/load.c:397] test Impulse Tracker
load_module [src/load.c:397] test Scream Tracker 3
load_module [src/load.c:397] test Scream Tracker 2
load_module [src/load.c:401] load format: Scream Tracker 2
==18089== Invalid read of size 4
==18089==    at 0x40363AB: stm_load (stm_load.c:210)
==18089==    by 0x4029818: load_module (load.c:402)
==18089==    by 0x4029D0E: xmp_load_module (load.c:547)
==18089==    by 0x804B170: main (main.c:399)
==18089==  Address 0x8c is not stack'd, malloc'd or (recently) free'd
==18089== 
==18089== 
==18089== HEAP SUMMARY:
==18089==     in use at exit: 9,566 bytes in 5 blocks
==18089==   total heap usage: 7 allocs, 2 frees, 10,270 bytes allocated
==18089== 
==18089== LEAK SUMMARY:
==18089==    definitely lost: 0 bytes in 0 blocks
==18089==    indirectly lost: 0 bytes in 0 blocks
==18089==      possibly lost: 0 bytes in 0 blocks
==18089==    still reachable: 9,566 bytes in 5 blocks
==18089==         suppressed: 0 bytes in 0 blocks
==18089== Rerun with --leak-check=full to see details of leaked memory
==18089== 
==18089== For counts of detected and suppressed errors, rerun with: -v
==18089== ERROR SUMMARY: 1 errors from 1 contexts (suppressed: 42 from 8)
Segmentation fault (core dumped)
```

cmatsuoka:
Ah interesting! Yeah, I think I just ran a regular build.

Also I must see why you approved the commit but it didn't clear the merge requirements.

cmatsuoka:
Reconfigured to accept approvals from all devel team members and merging PR now that's green.

sezero:
> Also I must see why you approved the commit 

I didn't

cmatsuoka:
Right, it was @AliceLR but it was not enough to enable the merge button. It was a team membership configuration, and it's hopefully fixed now.

AliceLR:
FWIW I was able to detect it on Linux but only after I fixed my local test-dev so it built the new test with ASan—I didn't try valgrind. I tried `-fstack-protector-all` to see if it would also catch this but it didn't (might be worth building the CI checks with it enabled anyway).

PR #187, opened 2020-11-04
Ignore MED volume/slide effects with param of 0, fix speed bound.

AliceLR:
This fixes a few minor OctaMED effects bugs.

* The slide up and slide down commands `1xx` and `2xx` should ignore a param of 0.
* The fine volume up and down commands `1Axx` and `1Bxx` should ignore a param of 0.
* The speed effect had faulty bounding (it is bounded between `0x01` and `0x20`, not 1 and 20).

There are some other missing or buggy commands here that I left alone because I'm not sure how to implement several of them right now.

* MMD0 `5xx` (old vibrato) isn't implemented (a small number of modules use this).
* `8xx` (set hold and decay) does nothing. There's an effect reserved that does nothing but also a second hold effect implemented in med_extras.c that sets the hold (but not decay).
* `Exx` (synth jump) does nothing.
* `FF4` and `FF5` (delay note one third of a line, delay note two thirds of a line) do nothing.
* `FF8` and `FF9` (filter) does nothing.
* `FFD` (set pitch) has an effect number reserved but does nothing.
* `FFE` (stop playing) does nothing.
* `16xx` should be able to use the whole param range.
* `18xx` should be able to use the whole param range.
* `1Fxy` should be able to both delay and then retrigger a note.

### DIFF
--- a/src/loaders/mmd_common.c
+++ b/src/loaders/mmd_common.c
@@ -86,20 +86,26 @@ void mmd_xlat_fx(struct xmp_event *event, int bpm_on, int bpmlen, int med_8ch,
 		 * chord sound or other special effect. Arpeggio works better
 		 * with some instruments than others.
 		 */
-		/* fall-through */
+		break;
 	case 0x01:
 		/* SLIDE UP 01
 		 * This slides the pitch of the current track up. It decreases
 		 * the period of the note by the amount of the argument on each
 		 * timing pulse. OctaMED-Pro can create slides automatically,
 		 * but you may want to use this function for special effects.
+		 * Note: a param of 0 does nothing and should be ignored.
 		 */
-		/* fall-through */
+		if (!event->fxp)
+			event->fxt = 0;
+		break;
 	case 0x02:
 		/* SLIDE DOWN 02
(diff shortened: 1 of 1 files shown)


sezero:
> There are some other missing or buggy commands here that I left alone because I'm not sure how to implement several of them right now.
> 
>     * MMD0 `5xx` (old vibrato) isn't implemented (a small number of modules use this).
> 
>     * `8xx` (set hold and decay) does nothing. There's an effect reserved that does nothing but also a second hold effect implemented in med_extras.c that sets the hold (but not decay).
> 
>     * `Exx` (synth jump) does nothing.
> 
>     * `FF4` and `FF5` (delay note one third of a line, delay note two thirds of a line) do nothing.
> 
>     * `FF8` and `FF9` (filter) does nothing.
> 
>     * `FFD` (set pitch) has an effect number reserved but does nothing.
> 
>     * `FFE` (stop playing) does nothing.
> 
>     * `16xx` should be able to use the whole param range.
> 
>     * `18xx` should be able to use the whole param range.
> 
>     * `1Fxy` should be able to both delay and then retrigger a note.

You can create a new bug entry for these if you want (so that they don't get lost.)

PR #189, opened 2020-11-05
Improve MOD loader checks for Mod's Grave WOW files, add tests.

AliceLR:
Mod's Grave WOW is a weird MOD variant with an M.K. signature but 8 channels instead of 4. The canonical way to detect these is to check the length of M.K. MODs. Unfortunately, that doesn't always work.

* If a MOD file is appended with junk data/greetz/whatever to exactly the right size, this will fool libxmp (see ponylips.mod).
* Some WOW files have an extra byte for some reason. (see acidfunk.wow).

I added a couple of basic checks that can be used to prevent basic false positives like ponylips.mod from happening. If it turns out there are somehow more of these, a pattern validation check (like the one Saga Musix added to OpenMPT) should pretty thoroughly rule out most regular MODs, but so far ponylips.mod is the only one of these I've seen.

I also added a couple of test files for these edge cases. This required tweaking `dump_module` again since I noticed other MODs with empty samples have different hashes than the weird 1 byte samples in fracture.stm (not that these get checked...).


sezero:
OK, thanks.

PR #190, opened 2020-11-05
Fix Schism Tracker version date calculation.

### LINKED ISSUE #154: Extended Schism Tracker version information
On 2020-10-28, the current Schism Tracker version format as found in IT and S3M headers will overflow. This was of course known beforehand so 0xFFF was declared to be a reserved value for any date starting from that day.

Since it's only one year until this format will overflow, I went ahead and prepared Schism Tracker for that day: If the version is 0xFFF, the *real* version is now stored in IT's reserved header field, or in the first two bytes of the unused 8 consecutive bytes in the S3M header. The data in these reserved fields follows the same as the value in `cwtv`, but without the tracker ID in the highest nibble (obviously), and 0x050 is not added to the value.

The attached files can be used to test the new version storage. They should decode as SchismTracker 2020-10-29.

[20201029.zip](https://github.com/schismtracker/schismtracker/files/3763851/20201029.zip)

AliceLR:
Fixes several issues with the calculation of Schism Tracker dates.

* The date calculation was using a UTC epoch time but calculating the version date with localtime_r instead of gmtime_r (maybe because Schism uses localtime_r... the difference is Schism generates its epoch time with mktime).
* The UTC epoch time was wrong (2009-10-01 instead of 2009-10-31).
* The S3M loader was not calculating Schism Tracker dates at all.
* Added detection for extended Schism Tracker dates (>2020-10-28) to both the S3M and IT loaders.

Fixes #154. I think I got this right, but this should *really* be reviewed by @sagamusix before it's merged.

edit: this sure doubles the number of spurious format truncation warnings for Schism Tracker dates though. 🙄

### DIFF
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -364,3 +364,54 @@ void libxmp_set_type(struct module_data *m, const char *fmt, ...)
 	vsnprintf(m->mod.type, XMP_NAME_SIZE, fmt, ap);
 	va_end(ap);
 }
+
+static int schism_tracker_date(int year, int month, int day)
+{
+	int mm = (month + 9) % 12;
+	int yy = year - mm / 10;
+
+	yy = yy * 365 + (yy / 4) - (yy / 100) + (yy / 400);
+	mm = (mm * 306 + 5) / 10;
+
+	return yy + mm + (day - 1);
+}
+
+/* Generate a Schism Tracker version string.
+ * Schism Tracker versions are stored as follows:
+ *
+ * s_ver <= 0x50:		0.s_ver
+ * s_ver > 0x50, < 0xfff:	days from epoch=(s_ver - 0x50)
+ * s_ver = 0xfff:		days from epoch=l_ver
+ */
+void libxmp_schism_tracker_string(char *buf, size_t size, int s_ver, int l_ver)
+{
+	if (s_ver >= 0x50) {
+		/* time_t epoch_sec = 1256947200; */
+		int t = schism_tracker_date(2009, 10, 31);
+		int year, month, day, dayofyear;
+
+		if (s_ver == 0xfff) {
+			t += l_ver;
+		} else
+			t += s_ver - 0x50;
+
+		/* Date algorithm reimplemented from OpenMPT.
+		 */
+		year = (int)(((int64)t * 10000L + 14780) / 3652425);
+		dayofyear = t - (365 * year + (year / 4) - (year / 100) + (year / 400));
+		if (dayofyear < 0) {
+			year--;
+			dayofyear = t - (365 * year + (year / 4) - (year / 100) + (year / 400));
+		}
+		month = (100 * dayofyear + 52) / 3060;
+		day = dayofyear - (month * 306 + 5) / 10 + 1;
+
+		year += (month + 2) / 12;
+		month = (month + 2) % 12 + 1;
+
+		snprintf(buf, size, "Schism Tracker %04d-%02d-%02d",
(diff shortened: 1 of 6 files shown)


sagamusix:
I didn't look in detail yet but in general the code looks sane. In OpenMPT we stopped using time functions from the standard library for this functionality because at least in some glibc versions they crashed in low-memory situations where according to API contracts they should return an error instead. It's unlikely that you would hit such a crash in a real-world situation but we did find it in our fuzzing process. Doing the same in libxmp would remove one dependency check from autotools and remove some ugly per-platform workarounds, and maybe even generate a smaller binary, as the manual date calculation is a very simple formula.

If you are interested in doing the same for libxmp in this commit, feel free to look into [OpenMPT's date calculation](https://github.com/OpenMPT/openmpt/blob/0641cc4b22efc881f6000075100db428861bdb09/soundlib/Load_it.cpp#L353).

sagamusix:
Oh, and I think it might make sense to unify the S3M/IT code into a single helper function somehow :)

AliceLR:
I think replacing gmtime_r there altogether is reasonable. I do not particularly like those localtime/gmtime platform hacks either (or the copypaste job for generating the tracker name for that matter).

Also, kind of a cheesy hack, but the module type field is big enough to expand `tracker_name` to a big enough size to shut up the warnings and still fit within bounds with the format and version.

sezero:
> I think replacing gmtime_r there altogether is reasonable.

Go for it, e.g. in a new gmtime.c source or something.

j-schultz:
Since the calculations here can be done in-line, I doubt it requires adding a new .c file.

sagamusix:
Since the calculations here can be done in-line, I doubt it requires adding a new .c file.

AliceLR:
I've had a look at the date algorithm OpenMPT uses and for the most part I think I'm satisfied with it. Just to cover myself, are you explicitly giving permission for me to create a derivative implementation of this algorithm under the MIT license? (I can't imagine anyone will try to promote a product off of libxmp having a date algorithm, BUT)

sagamusix:
Sure, but it's a well-known algorithm anyway, not mine. :)

AliceLR:
I don't think the algorithm can be copyrighted anyway, but some of the implementation details seemed specific enough to ask anyway. Thanks, I will update this branch soon.

AliceLR:
OK, I've updated the commit to use the date algorithm(s) from OpenMPT and to get rid of the localtime_r/gmtime_r dependency altogether.

src/loaders/common.c
```diff
@@ -364,3 +364,58 @@ void libxmp_set_type(struct module_data *m, const char *fmt, ...)
 	vsnprintf(m->mod.type, XMP_NAME_SIZE, fmt, ap);
 	va_end(ap);
 }
+
+static int schism_tracker_epoch(void)
```
sagamusix, line 368: If you are anyway going to put this in its own function rather than just keeping the epoch as a magic number as before, I would suggest to make this function take parameters year, month and day. Then it can be re-used to check against readable Schism Tracker versions when toggling tracker quirk flags (xmp doesn't currently do this but e.g. OpenMPT does it)

AliceLR: Yes, I think that would be a better solution. I've updated this function to take a date instead in case it ever needs to be reused (worst case it should optimize exactly the same as before).

sagamusix:
Looks fine now :)

sezero:
This is in now, thanks.

PR #193, opened 2020-11-06
Change return type of xmp_get_format_list to const char *const *.

### LINKED ISSUE #192: xmp_get_format_list() should return const?
In format.c, _farray is `const char *[]` whereas format_list() returns `(char **)_farray`
What if someone tries to modify the result of `xmp_get_format_list()`?

AliceLR:
This function returns a static array of `const char *` pointers to static data, but `format_list()` and `xmp_get_format_list()` both return this array as a `char **`, which doesn't really communicate that none of this data should be modified/freed. Not sure if this should actually be merged right now (it may be a breaking API change), but I think it should be addressed at some point.

Fixes #192.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -233,7 +233,7 @@ const unsigned int xmp_vercode
 
 .. _xmp_get_format_list():
 
-char \*\*xmp_get_format_list()
+const char \*const \*xmp_get_format_list()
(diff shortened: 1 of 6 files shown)


sezero:
Oops, didn' notice your patch when I pushed mine.

sezero:
Pushed the interdiff as ff818d7

PR #195, opened 2020-11-06
Increase maximum IFF chunk size to 8 MiB.

### LINKED ISSUE #166: Digibooster module not recognized

AliceLR:
This fixes the loading of several DigiBooster Pro modules with sample chunks larger than 4 MiB, of which I found >=22 in the ModLand collection. Fixes #166.

### DIFF
--- a/src/loaders/iff.h
+++ b/src/loaders/iff.h
@@ -13,7 +13,7 @@
 #define IFF_SKIP_EMBEDDED	0x10
 #define IFF_CHUNK_TRUNC4	0x20
 
-#define IFF_MAX_CHUNK_SIZE	0x400000
+#define IFF_MAX_CHUNK_SIZE	0x800000
 
 typedef void *iff_handle;
 


sagamusix:
I would suggest to remove the limit completely in another PR. It just makes no sense and gives a false sense of safety. If a file really contains 100 megabytes worth of samples, there is nothing wrong with loading them. If a library user is afraid of allocating 100MB of RAM, they should not pass a 100MB file to their decoder. Of course the story is different if the IFF chunk claims to be big but the file is truncated, but an upper limit for chunk sizes is not the answer to that problem.

sezero:
> I would suggest to remove the limit completely in another PR. It just makes no sense and gives a false sense of safety. If a file really contains 100 megabytes worth of samples, there is nothing wrong with loading them. 

That's reasonable.  @AliceLR, @cmatsuoka?

> Of course the story is different if the IFF chunk claims to be big but the file is truncated

Do the iff users check properly for EOF returns from hio?

PR #199, opened 2020-11-07
build: use configure flag to disable depackers

cmatsuoka:
Use a flag to the configure script to enable and disable depackers
instead of editing the variable directly on the makefile in systems
where the Makefile is generated using autoconf.

OS/2 and VC++ makefiles retain USE_DEPACKERS.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -7,16 +7,10 @@ VERSION		= $(VERSION_MAJOR).$(VERSION_MINOR).$(VERSION_RELEASE)
 # This dylib will support anything linked against COMPAT_VERSION through VERSION
 COMPAT_VERSION	= $(VERSION_MAJOR)
 
-# Module depacker functionality. To disable: make USE_DEPACKERS=0
-USE_DEPACKERS	= 1
-
 CC		= @CC@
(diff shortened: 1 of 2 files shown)


sezero:
This is good (I was lazy to do it this way.)

cmatsuoka:
Thanks for verifying it, could you mark it as approved or make any additional comments (in the "Files changed" tab above, and then "Review changes")?

I'm also fixing the CI tests in PR #202 so after that lands we can merge master into this PR so it can also be ready to be merged.

sezero (approved):
OK

sezero:
Out of curiosity: Why is 'Rebase and merge' option not enabled?

cmatsuoka:
It's just to make it more visible which set of commits came from each PR (and then you can see all the PR merges with `git log --first-parent`). If all PRs are stacked on top of each other in a linear way you can't do that.

sezero:
This can be merged now, no?

sezero:
PR #202 is done, and something is wrong with travis (takes eons to start -- or finalize?)
Manually merging this one so that we can go on.

PR #201, opened 2020-11-07
dylib: change undefined symbols to be error rather than dynamic lookup.

sezero:
I don't know why it has always been setup to dynamic_lookup,
but macOS dylib builds must not have any undefined symbols.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -135,7 +135,7 @@ lib/$(DLL): $(LOBJS)
 
 lib/$(DYLIB): $(LOBJS)
 	@mkdir -p lib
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
Thanks, this looks good, I don't know why it was set to dynamic_lookup.

sezero:
Obviously I can't delete the obsolete branch. Should I do PRs from personal forks?

cmatsuoka:
Yes, it's always better to do PRs from a fork and avoid creating temporary branches on the upstream repo. Don't worry about the branch, we can configure the repository to allow us to delete branches, I'll do it and protect the branch again afterwards to avoid deleting an important branch unintentionally.

sezero:
> Yes, it's always better to do PRs from a fork and avoid creating temporary branches on the upstream repo.

OK

> Don't worry about the branch, we can configure the repository to allow us to delete branches, I'll do it and protect the branch again afterwards to avoid deleting an important branch unintentionally.

Done deleting the obsolete branch

PR #203, opened 2020-11-07
Add LIBXMP_AMIGA to identify amiga[-like] platforms at compile time

sezero:


AliceLR:
I noticed there are checks on `__amigaos4__` in src/depackers/xfd_link.c, src/loaders/med.h, and src/loaders/rtm_load.c. Should these also be changed to `LIBXMP_AMIGA`?

sezero:
> I noticed there are checks on `__amigaos4__` in src/depackers/xfd_link.c, src/loaders/med.h, and src/loaders/rtm_load.c. Should these also be changed to `LIBXMP_AMIGA`? (not a recommendation, I'm just not sure.)

Nope, they're good. (They're just OS and/or SDK specific checks for typedefs)

> edit: there's also one in src/common.h line 23/25.

Nope (same)

AliceLR (approved):
Looks good to me, then. (Don't know if @cmatsuoka wants to comment though.)

sezero:
@cmatsuoka: waiting for your approval

PR #204, opened 2020-11-08
Fix test-dev sample, depack, prowizard, and api leaks.

AliceLR:
First of many. Fixes any trivial leaks in the test_sample_\*, test_depack_\*, test_prowizard_\*, and test_api_\* test files. There are some test_api_* leaks that this doesn't fix (see #110) but since those may be actual bugs they should probably be handled separately.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -36,6 +36,7 @@
 #include "list.h"
 #include "hio.h"
 #include "tempfile.h"
+#include "loaders/loader.h"
 
 #ifndef LIBXMP_NO_DEPACKERS
 #if !defined(HAVE_POPEN) && defined(_WIN32)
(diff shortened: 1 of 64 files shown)


AliceLR:
Looking into why this happened a few times...:
```
test 22: test_sample_load_delta: at test_sample_load_delta.c:24: Invalid 8-bit conversion: **fail**
test 23: test_sample_load_signal: at test_sample_load_signal.c:32: Invalid 8-bit conversion: **fail**
```

AliceLR:
So that's why those were static! Oof... :(

sezero:
Personally I don't like those `free(s.data - 4)`.  We need something,
say libxmp_free_sample() or free_sample_data() for this. Even a macro
would do.  That way, if internals ever change the users of this macro
won't be affected.

AliceLR:
Yeah, I'm not a fan of those either.

AliceLR:
Updated to add a `libxmp_free_sample` function. I've updated src/load.c and src/smix.c to use the new function instead of the same `data - 4` hack.

sezero (approved):
This is OK.

PR #109, opened 2017-11-03
Testing Travis CI

NoSuck:
Clang doesn't like something.

https://travis-ci.org/cmatsuoka/libxmp/builds/296784287

PR #206, opened 2020-11-08
Fix test-dev player, mixer, and libxmp_prepare_scan leaks.

AliceLR:
Third of ~~four~~ THREE patches. This fixes trivial leaks in the `test_player_*` and `test_mixer_*` tests, as well as the `libxmp_prepare_scan`-related leaks documented in #110. The first two sets were the same `xmp_release_module`/`xmp_free_context` copypaste as everywhere else, but the `libxmp_prepare_scan` fix required a new function.

This should be the last of the patches for trivial leaks in the tests. No promises these patches will get the CI ASan checks working again, but this might significantly reduce the log spam (going off of what I saw in the logs for #109). ~~The only problem left when I run `make devcheck` with ASan enabled is `test_api_test_module` leaking hio handles, which may or may not be related to #178.~~ It was a single wrong function call unrelated to that PR and I amended this patch to also fix it.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -416,6 +416,7 @@ struct context_data {
 char	*libxmp_adjust_string	(char *);
 int	libxmp_exclude_match	(const char *);
 int	libxmp_prepare_scan	(struct context_data *);
+void	libxmp_free_scan	(struct context_data *);
 int	libxmp_scan_sequences	(struct context_data *);
 int	libxmp_get_sequence	(struct context_data *, int);
 int	libxmp_set_player_mode	(struct context_data *);
--- a/src/load.c
+++ b/src/load.c
@@ -347,7 +347,7 @@ int xmp_test_module(const char *path, struct xmp_test_info *info)
 			}
 #endif
 
-			fclose(h->handle.file);
+			hio_close(h);
 
 #ifndef LIBXMP_NO_DEPACKERS
 			unlink_temp_file(temp);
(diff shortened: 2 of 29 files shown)


sezero (approved):
This looks good.

Travis, again, is taking eons to do its job. Will merge when it finishes.

sezero:
> `make devcheck` with ASan enabled is `test_api_test_module` leaking hio handles, 

Does it point to anywhere specific in code?

AliceLR:
> Does it point to anywhere specific in code?

I copied a dump of the output to #110 but there aren't any line numbers, just function names (I should make sure `-fno-emit-frame-pointer` is enabled and try again...).

AliceLR:
> Does it point to anywhere specific in code?

The issue may be load.c line 350, where `xmp_test_module` closes the underlying file instead of using `hio_close` on the HIO_HANDLE... not sure why that is.

sezero:
> > Does it point to anywhere specific in code?
> 
> The issue may be load.c line 350, where `xmp_test_module` closes the underlying file instead of using `hio_close` on the HIO_HANDLE... not sure why that is.

Can't think of anything reasonable for that. xmp_load_module does call hio_close as test_module should have.

AliceLR:
I've confirmed that changing that line fixes the `xmp_test_module` test. Since it's a 1-line patch and Travis is still thinking about running the checks I'll just amend this patch to add that.

sezero:
Good catch. That line came into being with commit d7c0954695 and stayed that way since.

sezero:
Something is wrong with travis (takes eons to start -- or finalize?)
Manually merging this one so that we can go on.

NoSuck:
`libxmp_free_scan` should be a nice addition.  Thank you for the work.

PR #208, opened 2020-11-08
test-dev: update data generation script to use python3

cmatsuoka:
Our module loading tests currently compare the data loaded from
the module file with reference data prepared by a python script.
Update this script to use python3.

### DIFF
--- a/test-dev/gen_module_data.py
+++ b/test-dev/gen_module_data.py
@@ -1,9 +1,9 @@
-#!/usr/bin/python
+#!/usr/bin/python3
 # vim: tabstop=8 expandtab shiftwidth=4 softtabstop=4
 
 import sys
 import os
-import md5
+import hashlib
 
 
 ROOT_PATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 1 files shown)


sezero:
I can't help with this one (not a py guy). Called Alice to review and/or test.

cmatsuoka:
This needs the old python bindings for libxmp re-generated for python3 using [ctypesgen](https://github.com/davidjamesca/ctypesgen) and some manual tweaking. This will be easier after the bindings are properly fixed/updated.

AliceLR:
The changes here look correct but I don't have much experience with python3.

FWIW I wasn't sure how to get the libxmp-python bindings updated when I was fixing tests as part of #181 so I added a C [reimplementation](https://github.com/libxmp/libxmp/blob/master/test-dev/gen_module_data.c) [here](https://github.com/libxmp/libxmp/blob/master/test-dev/util.c#L374) as a workaround.

AliceLR (approved):

sezero:
Is this not ready for merge?

cmatsuoka:
Since we have a C implementation this script is not really necessary, but it's not included in the distribution package and having it updated doesn't hurt, so I think we can merge it and then decide later if we drop it or not.

cmatsuoka:
(in fact I think this shouldn't be in the 4.5.0 milestone, because the script is not distributed with the player package)

AliceLR:
The C implementation still has bugs (example: I used a lot of `%u` where `%d` should have been and I had to patch a few of them as part of #218) and at the very least this script should be useful for testing the Python bindings, so I think it's definitely a good thing to have around. Since it's OK to merge this as-is, I'll go ahead and merge it.

PR #209, opened 2020-11-08
loaders,test-dev: address clang warnings

cmatsuoka:
Address code indent and typecast warnings issued when building with
clang on Ubuntu Groovy.

### DIFF
--- a/src/loaders/amf_load.c
+++ b/src/loaders/amf_load.c
@@ -205,13 +205,12 @@ static int amf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	}
 
 	for (i = 0; i < mod->ins; i++) {
-		/*uint8 b;*/
 		int c2spd;
 
 		if (libxmp_alloc_subinstrument(mod, i, 1) < 0)
 			return -1;
 
-		/*b =*/ hio_read8(f);
+		hio_read8(f);
 
 		hio_read(buf, 1, 32, f);
 		libxmp_instrument_name(mod, i, buf, 32);
--- a/test-dev/util.c
+++ b/test-dev/util.c
@@ -507,7 +507,7 @@ void dump_module(struct xmp_module *mod, FILE *f)
 			MD5Update(&ctx, xxs->data, len);
 		} else if (len > 0) {
 			/* The data file for fracture.stm had this hash for empty samples... */
-			MD5Update(&ctx, "1", 1);
+			MD5Update(&ctx, (unsigned char *)"1", 1);
 		}
 		MD5Final(d, &ctx);
 		fprintf(f, " ");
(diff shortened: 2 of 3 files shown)


src/loaders/amf_load.c
```diff
@@ -205,13 +205,12 @@ static int amf_load(struct module_data *m, HIO_HANDLE *f, const int start)
...
 
 	for (i = 0; i < mod->ins; i++) {
-		/*uint8 b;*/
 		int c2spd;
 
 		if (libxmp_alloc_subinstrument(mod, i, 1) < 0)
 			return -1;
 
-		/*b =*/ hio_read8(f);
+		hio_read8(f);
```
sezero, line 213: Clang is warning for this? Crazy.

sezero (approved):
OK with me.

test-dev/util.c
```diff
@@ -507,7 +507,7 @@ void dump_module(struct xmp_module *mod, FILE *f)
 			MD5Update(&ctx, xxs->data, len);
 		} else if (len > 0) {
 			/* The data file for fracture.stm had this hash for empty samples... */
-			MD5Update(&ctx, "1", 1);
+			MD5Update(&ctx, (unsigned char *)"1", 1);
```
sezero, line 510: This is ugly.  No other choice I guess?

src/loaders/amf_load.c
cmatsuoka: Apparently it considered that it could be intended to be part of the if. I think I should add braces there as well to avoid cases like this.

test-dev/util.c
cmatsuoka: I don't see any other obvious way to address that. gcc doesn't care about it, it's just clang but it's good to keep the compilation with no warnings. I think we should test with Visual C as well (and do it in Travis if possible).

AliceLR: I'm not even sure this is the correct thing to do here, this is just what I had to add to reproduce the hashes in the fracture.stm data file. It's not clear to me how the Python version of this util even came up with this hash (maybe it's hashing a string representation of the length when no data is available?).

sezero: Something like this should do:
```
diff --git a/test-dev/util.c b/test-dev/util.c
index b9bbba5..dd6a1a2 100644
--- a/test-dev/util.c
+++ b/test-dev/util.c
@@ -507,7 +507,8 @@ void dump_module(struct xmp_module *mod, FILE *f)
 			MD5Update(&ctx, xxs->data, len);
 		} else if (len > 0) {
 			/* The data file for fracture.stm had this hash for empty samples... */
-			MD5Update(&ctx, (unsigned char *)"1", 1);
+			static const unsigned char one_str[] = {'1',0};
+			MD5Update(&ctx, one_str, 1);
 		}
 		MD5Final(d, &ctx);
 		fprintf(f, " ");
```

PR #210, opened 2020-11-08
build: better handling of no depackers configuration flag

cmatsuoka:
Under some circumstances building with depackers disabled was
failing. This should make the flag parsing more robust.


sezero (approved):
OK

PR #211, opened 2020-11-09
Add LDFLAGS to libxmp-tests and utilities rules.

AliceLR:
This adds `$(LDFLAGS)` to the test-dev linking rules that were missing it (`libxmp-tests`, `gen_mixer_data`, and `gen_module_data`). I kept running into issues with this because I've been building with ASan by using `CFLAGS=-fsanitize=address LDFLAGS=-fsanitize=address ./configure` and the lack of `$(LDFLAGS)` on the `libxmp-tests` rule prevents it from linking correctly.

edit: placing `$(LDFLAGS)` after `-o $@` was for consistency with `libxmp-covertest` but I don't really like that and it can be moved to right after `$(LD)` if that's preferable.

### DIFF
--- a/test-dev/Makefile.in
+++ b/test-dev/Makefile.in
@@ -382,12 +382,12 @@ clean:
 utilities: gen_mixer_data gen_module_data
 
 gen_mixer_data: gen_mixer_data.o
-	@CMD='$(LD) -o $@ gen_mixer_data.o -L../lib -lxmp'; \
+	@CMD='$(LD) $(LDFLAGS) -o $@ gen_mixer_data.o -L../lib -lxmp'; \
 	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo LD $@ ; fi; \
 	eval $$CMD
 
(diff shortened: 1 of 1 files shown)


sezero (approved):
OK by me.

sezero:
> edit: placing `$(LDFLAGS)` after `-o $@` was for consistency with `libxmp-covertest` but I don't really like that and it can be moved to right after `$(LD)` if that's preferable.

Yes, moving it is better, and it can be done to libxmp-covertest too.

AliceLR:
OK, I updated it to move `$(LDFLAGS)` to just after `$(LD)`. There was also a missing `-lm` for gen_module_data.

sezero:
> There was also a missing `-lm` for gen_module_data (should maybe be handled with `$(LIBS)`?).

Would be better, yes. (BTW, it's a blah but we can exclude `AC_CHECK_LIB(m,pow)`
from mingw and darwin like in top-level configure.ac)

PR #212, opened 2020-11-09
Allow up to 512 rows in X-Tracker patterns.

### LINKED ISSUE #197: Verify upper pattern length limits for DigiBooster Pro, X-tracker, Graoumf, MASI, and Oktalyzer.
Split from #185. These formats need to have their pattern length limits verified since they are all subject to the 256 row limit of `libxmp_alloc_pattern_tracks`, but contain no format-specific sanity checks or documentation otherwise indicating a higher limit. 
```
dbm_load.c:206:         if (libxmp_alloc_pattern_tracks(mod, i, hio_read16b(f)) < 0)
dmf_load.c:215:         if (libxmp_alloc_pattern_tracks(mod, i, hio_read16l(f)) < 0)
gtk_load.c:248:         if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
masi_load.c:267:        if (libxmp_alloc_pattern_tracks(mod, i, rows) < 0)
okt_load.c:245: if (libxmp_alloc_pattern_tracks(mod, data->pattern, rows) < 0)
```
I did run checks on the ModLand 2014 torrent collection for DigiBooster Pro (DBM) files and found no modules with patterns over length 256, but ideally these should all be verified with their original trackers (aside from MASI...).

AliceLR:
X-Tracker modules can have up to 512 rows in a pattern. This is the only remaining pattern length limit issue I found in #197.  Unfortunately, this patch doesn't do anything to fix whatever other issues X-Tracker support has, so this loader is still disabled.

Fixes #197.

### DIFF
--- a/src/loaders/dmf_load.c
+++ b/src/loaders/dmf_load.c
@@ -194,6 +194,7 @@ static int get_patt(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 {
 	struct xmp_module *mod = &m->mod;
 	int i, j, r, chn;
+	int patrows;
 	int patsize;
 	int info, counter, data;
 	int track_counter[32];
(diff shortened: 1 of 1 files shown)


sezero (approved):
OK

PR #213, opened 2020-11-10
Update changelog

sezero:
Also update libxmp-lite changelog with relevant entries.

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -17,6 +17,8 @@ Stable versions
 	- allow up to 512 rows in X-Tracker patterns
 	- fix the loading of several DigiBooster Pro modules with large
 	  sample chunks
+	- fix IT bug where Cxx on same row as SBx would not be ignored
+	- fix IT bug where Qxy would ignore the volume parameter
 	- make it possible to disable module depacker functionality
 	Changes by viiri:
 	- fix samples corruption in STM modules
@@ -47,7 +49,14 @@ Stable versions
 	Changes by Carsten Teibes:
 	- fix lite build mod loader symbols
 	Changes by NoSuck:
+	- add new xmp_set_row() call to skip replay to the given row
+	- IT: T00 now repeats previous slide
 	- fix IT message length
+	- prevent clobbering of muted channels' volumes in IT modules
+	- clamp number of IT envelope nodes at load time
+	- fix IT message (comment) length miscalculation
+	- fix IT volume panning effect
+	- fix mute status on player creation
 	Changes by Cameron Cawley:
 	- fix sanity check in Digital Symphony loader
 	Fix problems reported by Ralf Hoffmann:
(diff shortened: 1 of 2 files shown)


lite/Changelog
```diff
@@ -1,6 +1,38 @@
...
+	  xmp_smix_load_sample() now accept const char* path parameters
+	  (no ABI change)
+	- explicitly document that callers of xmp_load_module_from_file()
+	  are responsible for closing their own file.
+	- remove nonportable use of fdopen in xmp_load_module_from_file()
+	- fix a seek issue with xmp_load_module_from_memory
+	- fix memory-io functions' error handling
+	- fix lite build mod loader symbols
+	- fix IT message length
+	- fix strictness of MOD pattern data tester
```
AliceLR, line 19: I'm not sure this change is applicable to libxmp-lite; the commit in question (https://github.com/libxmp/libxmp/commit/fa8ede35b76f23ea0d93f8ab22b2162469d49bc0) only changes the regular MOD loader. The lite MOD loader also appears to omit the pattern validation checks the regular MOD loader has.

lite/Changelog
```diff
@@ -1,6 +1,38 @@
 Stable versions
 ---------------
 
+4.5.0 ():
+	- make xmp_set_position() consistently clear pattern break/jump vars
+	- fix S3M ADPCM4 samples
```
AliceLR, line 6: This one also appears to be non-applicable to libxmp-lite: the ADPCM sample loader is disabled when `LIBXMP_CORE_PLAYER` is defined: https://github.com/libxmp/libxmp/blob/master/src/loaders/sample.c#L283.

lite/Changelog
```diff
@@ -1,6 +1,38 @@
...
+	- fix a seek issue with xmp_load_module_from_memory
+	- fix memory-io functions' error handling
+	- fix lite build mod loader symbols
+	- fix IT message length
+	- fix strictness of MOD pattern data tester
+	- fix loading of XMLiTE XM modules
+	- handle XM 16-bit samples with odd in-file data
+	- fix smix sample allocation
+	- force reset of buffer state on player start
+	- fix double free in case of ADPCM sample load error
```
AliceLR, line 24: See the above comment re: ADPCM loading being disabled for libxmp-lite.

lite/Changelog
AliceLR: Actually, why is the ADPCM loader disabled for libxmp-lite? It looks like a fairly small amount of code.

lite/Changelog
sezero: Updated the lite-side of the patch now.

AliceLR (approved):
The updates to both changelogs appear correct to me (I don't know if I missed anything else that libxmp-lite disables though). @cmatsuoka should probably also look at this.

sezero:
Just added some missed entries to changelog

docs/Changelog
```diff
@@ -30,10 +45,18 @@ Stable versions
...
+	- several code clean-ups
 	Changes by Carsten Teibes:
 	- fix lite build mod loader symbols
 	Changes by NoSuck:
+	- add new xmp_set_row() call to skip replay to the given row
+	- IT: T00 now repeats previous slide
 	- fix IT message length
+	- prevent clobbering of muted channels' volumes in IT modules
+	- clamp number of IT envelope nodes at load time
+	- fix IT message (comment) length miscalculation
```
AliceLR, line 57: Are both of these IT message length entries for the same fix or for different things?

sezero: Yeah. Removed dups now.

PR #216, opened 2020-11-10
fix Makefile.vc and Makefile.os2 generation after commit 79f13b5

sezero:
because configure substitutes @DEPACKER_OBJS@ with $(DEPACKER_OBJS)

also change OBJ to OBJS in Makefile.os2 for consistency with Makefile.vc


AliceLR (approved):
This patch looks good.

sezero:
> This patch looks good.

Merged, thanks

Issue #152, opened 2019-10-20
Keyoff+Instrument change bug with xyce-dans_la_rue.xm


vlad54rus:
Tried to play xyce-dans_la_rue.xm (https://modarchive.org/index.php?request=view_by_moduleid&query=172032) with latest Xmp For Android and encountered this issue.
When keyoff note paired with instrument change command, previous instrument starts playing again at full volume.
xyce-dans_la_rue.xm, patterns 0E, 0F, 16,17.

sezero:
This doesn't sound like an Android-specific issue.

PR #217, opened 2020-11-11
Fix tempo assignment in pre-scan loop

### LINKED ISSUE #153: Crash at seek with xyce-dans_la_rue.xm
Another issue with xyce-dans_la_rue.xm - Xmp for Android crashes if i try to seek. Removing E60 command from pattern 0F channel 4 line 20 seems to solve this problem, but i want to know if it can be fixed on xmp side.

cmatsuoka:
Before playing a module we run a pre-scan to get timestamps for
each pattern, among other pieces of information. Update this data
if it's not set, meaning that we don't update e.g. inside a pattern
loop. This should fix seek in xyce-dans_la_rue.xm.
    
Fixes #153
    

sezero (approved):
OK.

sezero:
Does this affect #152 in any way?

cmatsuoka:
> Does this affect #152 in any way?

It doesn't seem to be related, I'm investigating it.

AliceLR (approved):
This fix looks correct as far as I can tell.

Issue #103, opened 2017-08-13
.it sample vibrato rate and depth are off


AliceLR:
This .it file uses a sample vibrato configuration that is immediately noticeable in Impulse Tracker.  When played with xmp, the vibrato is performed but is much harder to hear because the depth is not as deep as it should be.  Additionally, the vibrato sounds slower than when played in IT.

[sample_vibrato.it.zip](https://github.com/cmatsuoka/libxmp/files/1220279/sample_vibrato.it.zip)

AliceLR:
Reversing the direction of the bitshift when converting sample vibrato to instrument vibrato makes the depth sound closer to Impulse Tracker. (it_load.c, line 817)
```C
				sub->vde = ish.vid << 1;
```
The rate is still too slow, though.

AliceLR:
The other thing I ended up doing was to change line 816 to add 1 to the rate:
```
				sub->vra = ish.vis + 1;	/* sample to sub-instrument vibrato */
```
I don't fully understand how the instrument vibrato works in libxmp, though, or whether or not this is actually a good idea.  It did make the .it files I was using to test sound closer to IT, though.

AliceLR:
From some experimentation in Impulse Tracker the vibrato properties listed [here](https://modland.com/pub/documents/format_documentation/Impulse%20Tracker%20v2.04%20(.it).html) seem to function:

* `ViS`/Vibrato Speed: how fast the vibrato cycles (steps through the table per tick?). When this is 0, nothing happens. When at 1, the vibrato is very slow. At 64, the vibrato cycles quite fast. (range: 0-64)
* `ViD`/Vibrato Depth: magnitude of vibrato in fine linear slide steps. (range: 0-32, not 64 like the doc claims)
* `ViT`/Vibrato Waveform: not relevant to this issue.
* `ViR`/Vibrato Rate: sweep rate of vibrato in 256ths of a fine linear slide step. The depth begins at 0 and increases by this value until it reaches `ViD`. (range: 0-255, not 64 like the doc claims)

How they are translated to xmp subinstrument values (master):
```C
				sub->vra = ish.vis;	/* sample to sub-instrument vibrato */
				sub->vde = ish.vid >> 1;
				sub->vwf = ish.vit;
				sub->vsw = (0xff - ish.vir) >> 1;
```

How these values are actually used in libxmp (read_event.c, 146):
```C
		libxmp_lfo_set_depth(&xc->insvib.lfo, sub->vde);
[...]

PR #218, opened 2020-11-11
Fix IT sample global volume and sample vibrato.

### LINKED ISSUE #102: .it global sample volume is broken
This .it file contains two copies of the same sample with different global sample volumes that are played at different pitches.  When played in Impulse Tracker, one low sample can be heard clearly; when it's played with xmp, both samples can be heard equally loud.  libxmp seems to treat the global sample volume as maximum.

[global_sample_volume.it.zip](https://github.com/cmatsuoka/libxmp/files/1220272/global_sample_volume.it.zip)

AliceLR:
Fixes a few bugs in IT sample global volume and sample vibrato handling:

* IT sample global volume was being ignored in sample mode due to `QUIRK_INSVOL` not being enabled. `QUIRK_INSVOL` is now unconditionally enabled for IT modules.
* IT sample global volume was also being ignored in instrument mode for IT 1.x modules due to `QUIRK_INSVOL` not being enabled for IT 1.x modules.
* Fixed an incorrect bitshift on IT sample vibrato depth that resulted in IT sample vibrato being very weak.
* Sample vibrato rate is now rounded instead of truncated in read_event.c. This is a bandaid fix and the LFO period size should probably eventually be expanded.

Fixes #102 and #103.

edit: 5 failing tests from this patch, will look into them.


AliceLR:
I fixed the one module data test that this broke (differences were due to the change in sample vibrato depth). Haven't looked too close at the four failing OpenMPT mixer tests yet.

AliceLR:
Remaining tests:
* `test_openmpt_it_autovibrato_reset.c`: the property it is testing (the sample vibrato sweep) is unaffected by these changes, so I've just updated the data file.
* `test_openmpt_it_filter_reset.c`: ditto (testing filter macros, changes are due to the sample global volume fix).
* `test_openmpt_it_panreset.c`: ditto (testing `X80`, changes are due to sample global volume fix).
* `test_openmpt_it_samplenumberchange.c`: ditto (testing sample numbers without notes, changes are due to sample global volume fix).

sezero:
Let's have @sagamusix and @cmatsuoka look at this.

cmatsuoka (approved):
IT support in libxmp is generally innacurate and problems are fixed as they're reported, so if this change fixes issues or makes replay sound "less wrong" it should be good to go. (There are many hacks in IT replaying caused by IT support being added after the common replay code was already in place -- in the future we should refactor the common player to be more IT friendly, including adding proper bidirectional sample  loops).

AliceLR:
One of the tests on the master branch failed after this was merged: https://travis-ci.org/github/libxmp/libxmp/builds/743831472

I did a local build of libxmp and the tests on my iMac with GCC with no issue so that's not the problem. Is this maybe an issue on Travis' end?

sezero:
> One of the tests on the master branch failed after this was merged: https://travis-ci.org/github/libxmp/libxmp/builds/743831472
> 
> I did a local build of libxmp and the tests on my iMac with GCC with no issue so that's not the problem. Is this maybe an issue on Travis' end?

Seems like it

PR #219, opened 2020-11-11
Fix xm keyoff with instrument

### LINKED ISSUE #152: Keyoff+Instrument change bug with xyce-dans_la_rue.xm

cmatsuoka:
When a XM keyoff event is used with an instrument, don't reset the
volume. This fixes xyce-dans_la_rue.xm channel 0 in patterns 0E/0F.
    
Fixes #152
    

AliceLR:
This definitely seems like an improvement over the previous behavior in pattern 0F but something still sounds off in pattern 0E, channels 0 and 11. In FastTracker 2 the notes in channels 0 and 11 retrigger when the keyoff+instrument is encountered in pattern 0E, but in libxmp this no longer occurs. So keyoff+instrument resets the volume in the middle of 0E for some reason, but not at the start of 0F.

cmatsuoka:
I misunderstood the behavior in FT2 and disabled the volume reset in the first keyoff -- it became clear that it's resetting the volume after I changed the volume slide before the keyoff to a larger value in FT2. It should play correctly now.

AliceLR:
Channels 0 and 11 sound correct now but I think there's still one issue. Notes in channels 6 and 10 that carry over from pattern 0D are louder in xmp than in FastTracker 2 (channel 10 particularly noticeable since it's a 4th, might be quieter in FT2 due to the `A00`s at the end of pattern 0D). I don't know if that's relevant to this particular issue or if it's a different problem.

sagamusix:
Please note that this XM file was made with a very old version of OpenMPT and relies on an MPT-specific bug. The bugfix should only be applied to such files. I'll have to double-check when it was fixed but any file that uses the new identification scheme ("OpenMPT 1.xx.yy.zz" in the XM header) should not use this quirk, and of course files made with most other trackers.

cmatsuoka:
It seems that the volume difference in channel 10 happens because we updated the instrument to 4 but keyoff + instrument should preserve the instrument that's currently playing according to what happens in FT2. @sagamusix, do you have a test module for this scenario?

sagamusix:
Test for the OpenMPT behaviour? I only really have test cases for the correct play behaviour.

cmatsuoka:
I mean a test for the correct FT2 behavior in the case mentioned by @AliceLR to see if keyoff+instr preserves the old instrument or switches to the new one. (libxmp for some reason was changing the instrument, but FT2 keeps the original one.)

sagamusix:
https://wiki.openmpt.org/Development:_Test_Cases/XM#keyoff.2Binstr.xm is the relevant test case, I think.

AliceLR (approved):
With the latest update, patterns 0E and 0F both sound correct to me and I haven't noticed anything else unusual. Going to go ahead and approve this.

> Please note that this XM file was made with a very old version of OpenMPT

I didn't realize MPT and older versions of OpenMPT report as FastTracker v2.00 in the tracker field somehow, despite literally having an XM I made in old MPT sitting around that reports as FastTracker v2.00 (not really a format I use much). Wow...

sezero:
Merging this manually. (Travis sanitizer failures are irrelevant - linkage issue to be resolved.)

PR #220, opened 2020-11-11
build: add experimental travis windows build

cmatsuoka:
Experimental Windows build on Travis. Documentation is a bit scarce, so
this was written on a trial-and-error basis. Variable substitution and directory
separator slashes/backslashes doesn't seem to work like in regular bash.

Clang sanitizer builds disabled because they failed consistently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,15 +39,28 @@ matrix:
             - gcc-mingw-w64-x86-64
             - gcc-mingw-w64
 
-    - os: linux
-      dist: focal
-      compiler: clang
-      env: CFLAGS="-fsanitize=address" LDFLAGS="-fsanitize=address"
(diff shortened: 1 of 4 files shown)


sezero:
Can you also add sanitizer checks back, so that this closes #109 and #110?

cmatsuoka:
Windows builds working for 32 and 64 bit. Only the quick test is executed, ideally we should add the full set of regression tests but the testing harness must be ported to windows first.

sezero (commented):
The sanitizer checks are commented oft because of clang linkage issue?
It works now since #229 is merged.

lite/Makefile.vc.in needs updating too, for symmetry.

cmatsuoka:
Thanks, I'll merge master and fix the lite Makefile.

sezero (approved):
OK

PR #221, opened 2020-11-11
Fix DigiBooster Pro volume envelope number of points.

### LINKED ISSUE #196: Some Digibooster modules don't play correctly
Check out the attached archive. The module in that archive doesn't play correctly with libxmp. The file `good.mp4` shows what it should sound like and the file `bad.mp4` shows what it currently sounds like with libxmp. The problems begin around 23 seconds into the song.
 
[digib.zip](https://github.com/cmatsuoka/libxmp/files/5502569/digib.zip)

AliceLR:
DigiBooster Pro envelopes store the number of sections in an envelope but libxmp was treating it like the number of points, so the last envelope point would be ignored. I also added some sanity checks to the envelope loader to prevent out-of-bounds writes due to invalid instrument numbers. Probably fixes #196 (unless something else is broken).


sezero:
Let's have @sagamusix and @cmatsuoka look at this.

src/loaders/dbm_load.c
```diff
@@ -321,22 +321,32 @@ static int get_smpl(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
...
 
 	for (i = 0; i < nenv; i++) {
 		ins = hio_read16b(f) - 1;
-		mod->xxi[ins].aei.flg = hio_read8(f) & 0x07;
-		mod->xxi[ins].aei.npt = hio_read8(f);
-		mod->xxi[ins].aei.sus = hio_read8(f);
-		mod->xxi[ins].aei.lps = hio_read8(f);
-		mod->xxi[ins].aei.lpe = hio_read8(f);
+		flg = hio_read8(f) & 0x07;
+		npt = hio_read8(f) + 1;
```
sagamusix, line 333: should probably do the addition later to avoid integer overflow. nothing bad happens in this specific case but would still be cleaner.

AliceLR: This is fixed now, thanks. I also applied this change to `ins`, where something bad *was* happening: it could resolve to -1 and crash due to another missing out-of-bounds write check.

AliceLR:
`npt` is an int so adding 1 would still put it over 32, which the sanity check should catch, unless I am still missing something.

j-schultz:
Ah, I completely missed the different types. It's okay then :)

sagamusix:
Ah, I completely missed the different types. It's okay then :)

sagamusix:
Looks good but I don't see a way to approve the PR.

AliceLR:
Yeah, I'm not 100% sure what the rationale is behind `ins` being an int (considering the whole indexing instrument -1 thing...) but I just added the declarations for the new variables to the same line since the expanded range is convenient for exactly that issue re: `npt`.

sagamusix (approved):

sezero:
Merging this. The travis sanitizer failures are irrelevant (it's a linkage issue to be resolved.)

PR #225, opened 2020-11-12
configure.ac: add -Wl,--no-undefined to LDFLAGS if supported.

sezero:
based on SDL configury.  move Darwin -Wl,-undefined,error to configure
for consistency.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -126,7 +126,7 @@ lib/$(DLL): $(LOBJS)
 
 lib/$(DYLIB): $(LOBJS)
 	@mkdir -p lib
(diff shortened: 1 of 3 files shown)


sezero:
@AliceLR: can you have a look at this?

AliceLR (approved):
This looks good to me. I wasn't sure about unconditionally adding `-Wl,-undefined,error` but since it's been around since Tiger or earlier it's probably a non-issue. I also didn't know `--no-undefined` is different/broken on OpenBSD but I'm not surprised.

sezero:
Thanks. Merging.

PR #227, opened 2020-11-12
test travis sanitizer checks with gcc as compiler

sezero:


AliceLR:
Could have sworn these were working last night... :|

GCC should be fine for ASan but it doesn't have support for MSan.

sezero:
> Could have sworn these were working last night... :|

-Wl,--no-undefined is in the way, because for whatever reason clang
doesn't link to asan for shared libs and it worked before because the
library's asan needs were satisfied by the exe.

> GCC should be fine for ASan but it doesn't have support for MSan.

Yeah. IIRC, closest thing is -fsanitize=address combined with -fsanitize=leak
(not sure at the moment.)

AliceLR:
I made a patch to turn off `--no-undefined` when the sanitizers are enabled, can you check it? Though Travis still hasn't finished checking it... #229

PR #229, opened 2020-11-12
Don't use --no-undefined when linking with -fsanitize.

AliceLR:
It appears that `--no-undefined` is completely broken when trying to use the sanitizers with clang. I added a check to configure.ac to test for `fsanitize` before trying to add it. (I tried putting various programs in the `AC_TRY_LINK` prior to this but none of them flagged the same issue, probably because it only happens when linking a library.)

This should hopefully fix the new Travis sanitizer issues.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,13 +54,20 @@ cygwin* | *djgpp | mint* | amigaos* |aros* | morphos*)
   ;;
 esac
 
+AC_CHECK_DEFINED(__clang__)
+
 case "${host_os}" in
 dnl Skip this on platforms where it is just simply busted.
 openbsd*) ;;
  darwin*) LDFLAGS="$LDFLAGS -Wl,-undefined,error" ;;
-       *) save_LDFLAGS="$LDFLAGS"
-          LDFLAGS="$LDFLAGS -Wl,--no-undefined"
-          AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          dnl For whatever reason, the Clang sanitizers and --no-undefined for
+          dnl shared libraries are incompatible.
+       *) if test "$ac_cv_defined___clang__" = "no" || test "${LDFLAGS#*fsanitize}" = "$LDFLAGS"
+          then
+            save_LDFLAGS="$LDFLAGS"
+            LDFLAGS="$LDFLAGS -Wl,--no-undefined"
+            AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          fi
           ;;
 esac
 
--- a/test/test.c
+++ b/test/test.c
@@ -110,9 +110,15 @@ int main()
 
 	printf(" pass\n");
 
+	xmp_release_module(c);
+	xmp_free_context(c);
 	exit(0);
 
     err:
 	printf(" fail\n");
+	if (c) {
+		xmp_release_module(c);
+		xmp_free_context(c);
+	}
 	exit(1);
 }


AliceLR:
Wow, still running...

sezero:
We can run sed on the generated Makefile, too.

Another option is reverting --no-undefined although I really don't like it.

No other option?

Do we know why clang isn't linking to asan etc libs when generating a shared lib?

AliceLR:
I linked it in #224 instead of somewhere it'd be useful, but here's a relevant issue: https://github.com/google/sanitizers/issues/380. Apparently these flags just do not work together.

I don't really like the idea of reverting --no-undefined. Is there a specific problem you had in mind that wouldn't be caught with the check I added but could be avoided with the sed solution (not doubting you, just curious—I do not have much experience at all with autoconf)? I don't think this check is perfect but it's enough to get the Travis checks and any local sanitizer builds I've done working.

sezero:
> I linked it in #224 instead of somewhere it'd be useful, but here's a relevant issue: [google/sanitizers#380](https://github.com/google/sanitizers/issues/380). Apparently these flags just do not work together.

OK, missed the link. (Will look at it, thanks.)

> I don't really like the idea of reverting --no-undefined.

OK

> Is there a specific problem you had in mind that wouldn't be caught with the check I added but could be avoided with the sed solution (not doubting you, just curious—I do not have much experience at all with autoconf)?

We don't need dirtying configury (although it's debatable which one would be more ugly or dirty)

> I don't think this check is perfect but it's enough to get the Travis checks and any local sanitizer builds I've done working.

That's good. Although my _personal_ preference would be editing the generating makefile. If you insist though, we can pull this in as is.

AliceLR:
I'm not insisting. I think it'd be better to do this right the first time, I'm just not sure what doing it "right" is (maybe adding configure flags to enable asan/msan/etc. that disable --no-undefined).

sezero:
> I'm not insisting. I think it'd be better to do this right the first time, I'm just not sure what doing it "right" is (maybe adding configure flags to enable asan/msan/etc. that disable --no-undefined).

Hm, maybe detect clang in configury and only then disabling --no-undefined when sanitizer flags are present?

sezero:
I'm also curious whether the sanitize builds link correctly as a dylib with -Wl,-undefined,error

AliceLR:
> Hm, maybe detect clang in configury and only then disabling --no-undefined when sanitizer flags are present?

Just to clarify, do you mean both detecting clang and and adding new sanitizer configure flags, or detecting clang and checking LDFLAGS for sanitizer flags like in the current patch? (the first seems better...)

> I'm also curious whether the sanitize builds link correctly as a dylib with -Wl,-undefined,error

I can check this when I get a chance.

sezero:
> > Hm, maybe detect clang in configury and only then disabling --no-undefined when sanitizer flags are present?
> 
> Just to clarify, do you mean both detecting clang and and adding new sanitizer configure flags, or detecting clang and checking LDFLAGS for sanitizer flags like in the current patch? (the first seems better...)

Whichever

> > I'm also curious whether the sanitize builds link correctly as a dylib with -Wl,-undefined,error
> 
> I can check this when I get a chance.

OK

AliceLR:
> I'm also curious whether the sanitize builds link correctly as a dylib with -Wl,-undefined,error

I'm able to link libxmp and run the tests just fine with both ASan and `-Wl,-undefined,error` enabled on my Mac Mini (High Sierra, clang 10). MSan appears to not be supported by this version of clang on Mac.

The MacPorts builds of clang for Snow Leopard unfortunately do not support any of the sanitizers.

AliceLR:
Updated patch for detecting Clang, does this look about right or is there a better way to detect it? (Not pushing an update yet since doing that too much seems to make Travis even slower and I haven't added sanitizer config flags yet.)
```diff
diff --git a/configure.ac b/configure.ac
index 469d6afc..9dc7acf2 100644
--- a/configure.ac
+++ b/configure.ac
@@ -54,13 +54,24 @@ cygwin* | *djgpp | mint* | amigaos* |aros* | morphos*)
   ;;
 esac
 
+AC_CHECK_DEFINED(__clang__,[
+  is_clang=yes
+],[
+  is_clang=no
+])
+
 case "${host_os}" in
 dnl Skip this on platforms where it is just simply busted.
 openbsd*) ;;
  darwin*) LDFLAGS="$LDFLAGS -Wl,-undefined,error" ;;
-       *) save_LDFLAGS="$LDFLAGS"
-          LDFLAGS="$LDFLAGS -Wl,--no-undefined"
-          AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          dnl For whatever reason, the Clang sanitizers and --no-undefined for
+          dnl shared libraries are incompatible.
+       *) if test "$is_clang" = "no" || test "${LDFLAGS#*fsanitize}" = "$LDFLAGS"
+          then
+            save_LDFLAGS="$LDFLAGS"
+            LDFLAGS="$LDFLAGS -Wl,--no-undefined"
+            AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          fi
           ;;
 esac
 
```

AliceLR:
Update on adding sanitizer flags: I have them working on the main configure.ac, but adding them to test-dev looks like it'd be a copypaste job. I think it might be better manually setting CFLAGS/LDFLAGS for now...

sezero:
This is looking good,

> Update on adding sanitizer flags: I have them working on the main configure.ac,

Slightly simplified patch:
```
diff --git a/configure.ac b/configure.ac
index 469d6af..38dc243 100644
--- a/configure.ac
+++ b/configure.ac
@@ -54,13 +54,20 @@ cygwin* | *djgpp | mint* | amigaos* |aros* | morphos*)
   ;;
 esac
 
+AC_CHECK_DEFINED(__clang__)
+
 case "${host_os}" in
 dnl Skip this on platforms where it is just simply busted.
 openbsd*) ;;
  darwin*) LDFLAGS="$LDFLAGS -Wl,-undefined,error" ;;
-       *) save_LDFLAGS="$LDFLAGS"
-          LDFLAGS="$LDFLAGS -Wl,--no-undefined"
-          AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          dnl For whatever reason, the Clang sanitizers and --no-undefined for
+          dnl shared libraries are incompatible.
+       *) if test "$ac_cv_defined___clang__" = "no" || test "${LDFLAGS#*fsanitize}" = "$LDFLAGS"
+          then
+            save_LDFLAGS="$LDFLAGS"
+            LDFLAGS="$LDFLAGS -Wl,--no-undefined"
+            AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
+          fi
           ;;
 esac
 
```
> adding them to test-dev looks like it'd be a copypaste job.

test-dev side doesn't even need the patch (or am I missing something?)

AliceLR:
> Slightly simplified patch:

Nice catch, I will update my local patch to match.

> test-dev side doesn't even need the patch (or am I missing something?)

Specifically re: test-dev I mean adding the options for the sanitizers (since both libxmp and the regression tests should be built with the same sanitizer options). This can probably be simplified (have never edited a configure.ac before this PR), but the boilerplate I came up with that would have to be copied is something like:
```m4
AC_ARG_ENABLE(asan,      [  --enable-asan           Build with AddressSanitizer])
AC_ARG_ENABLE(msan,      [  --enable-msan           Build with MemorySanitizer])
AC_ARG_ENABLE(ubsan,     [  --enable-ubsan          Build with UndefinedBehaviorSanitizer])
```
```m4
dnl XMP_TRY_LINK(<message>,<cache-var>,<cflags>,<ldflags>,<program>,<ifyes>,<ifno>)
AC_DEFUN([XMP_TRY_LINK],[
  AC_CACHE_CHECK([$1],[$2],[
    oldcflags="${CFLAGS}"
    oldldflags="${LDFLAGS}"
    CFLAGS="${CFLAGS} $3"
    LDFLAGS="${LDFLAGS} $4"
    AC_LINK_IFELSE([AC_LANG_SOURCE([[$5]])],[$2=yes],[$2=no],[true])
    CFLAGS="${oldcflags}"
    LDFLAGS="${oldldflags}"])
  AS_IF([test "x$$2" = xyes], [$6], [$7])])
```
```m4
using_sanitizers=no

if test "$enable_asan" = "yes"
then
  asan_flags="-fsanitize=address -fno-omit-frame-pointer -g"
  XMP_TRY_LINK(whether compiler understands -fsanitize=address,
    ac_cv_c_flag_fsanitize_address,
    ${asan_flags},
    ${asan_flags},
    [int main(void) { return 0; }],
    [CFLAGS="${CFLAGS} ${asan_flags}"
    LDFLAGS="${LDFLAGS} ${asan_flags}"
    using_sanitizers=yes],
    AC_MSG_ERROR([Failed to link AddressSanitizer.]))
fi
(more or less the same thing for MSan and UBSan...)
```
...I think I'd rather just stick with that LDFLAGS check unless there's ever a way to configure test-dev with the same configure script as libxmp.

AliceLR:
I went ahead and pushed an updated patch with the clang check. I also got LeakSanitizer spam from one of my local tests with `make check` earlier so I added `xmp_release_module`/`xmp_free_context` to test.c.

sezero (approved):
OK.

AliceLR:
Wow, the Travis tests already passed? I was sure they'd be going for several more hours...

PR #233, opened 2020-11-16
Enable fine effects for DigiBooster Pro modules.

AliceLR:
DigiBooster Pro modules support fine volume slide and portamento but the quirk to enable that wasn't set in the DigiBooster Pro loader. This fixes multiple instances of loud notes in "Krashan - The Call of Quantum Field.dbm" caused by `AF1` being interpreted as a slide up instead of as a fine slide down (reported in #196).


sezero (approved):
OK

sezero:
Can #196 be closed now?

AliceLR:
I'm not sure, I was going to wait on it for further feedback in case there are more issues with that module.

sezero:
OK, let's wait and see. Will drop a note at #196.

PR #234, opened 2020-11-16
Update changelog

sezero:


AliceLR:
The only thing I noticed missing from these is 0a6f4cc1/"configure.ac: add -Wl,--no-undefined to LDFLAGS if supported."; should that also be in the changelog? This looks good otherwise.

sezero:
> configure.ac: add -Wl,--no-undefined to LDFLAGS if supported."; should that also be in the changelog?

No need

PR #235, opened 2020-11-17
Add support for DigiBooster Pro pan envelopes.

AliceLR:
I noticed that xmp doesn't support DigiBooster Pro pan envelopes or DSP effects while fixing other issues with DigiBooster Pro modules. Since the pan envelopes are fairly trivial I went ahead and preemptively added them before that gets reported in #196 :-)

Minor compatibility point: the [DigiBooster 3 format specification](http://www.digibooster.de/en/format.php) indicates that pan envelope values are in the range [-128,128] (which can be verified by looking at one of the [demo modules](http://krashan.ppa.pl/music/Krashan%20-%20M2.dbm)), but every DBM module in my local copy of the ModLand collection uses values from 0-64, with 32 as the center. This appears to be a change specifically for DigiBooster 3.


sezero (approved):
This seems OK.

PR #236, opened 2020-11-17
Fix GDM set pan position effect.

AliceLR:
Adds an extended GDM effect I missed when fixing support for the fine slides. None of the other special extended GDM effects in the MenTaLguY documentation are actually implemented in BWSB, so I think that should be every GDM effect working now.

Test S3M and its converted GDM: [s3m_pan.zip](https://github.com/libxmp/libxmp/files/5551981/s3m_pan.zip)


sezero (approved):
OK.  Will merge as soon as Travis completes.

PR #238, opened 2020-11-17
minor adjustments for 4.5.0:

sezero:
- load.c (load_module): remove empty if construct from hio_error()
  when clearing error flag.
- prowiz.c (pw_wizardry): ditto.
- lite/test.c: release module and free context upon exit.
- lite/configure.ac: apply clang -fsanitize/--no-undefined patch to lite, too.
- README: change version to 4.5.
- lite/README: extend copyright years to 2020.
- docs/manpage-header.rst: change version to 4.5 and date to 2020.
- docs/libxmp.rst: whitespace fixes.

### DIFF
--- a/src/load.c
+++ b/src/load.c
@@ -451,10 +451,7 @@ static int load_module(xmp_context opaque, HIO_HANDLE *h)
 	test_result = load_result = -1;
 	for (i = 0; format_loaders[i] != NULL; i++) {
 		hio_seek(h, 0, SEEK_SET);
-
-		if (hio_error(h)) {
-			/* reset error flag */
-		}
+		hio_error(h); /* reset error flag */
 
 		D_(D_WARN "test %s", format_loaders[i]->name);
 		test_result = format_loaders[i]->test(h, NULL, 0);
--- a/src/loaders/prowizard/prowiz.c
+++ b/src/loaders/prowizard/prowiz.c
@@ -118,9 +118,7 @@ int pw_wizardry(HIO_HANDLE *file_in, FILE *file_out, const char **name)
 	}
 
 
-  /********************************************************************/
   /**************************   SEARCH   ******************************/
-  /********************************************************************/
 
 	for (i = 0; pw_formats[i] != NULL; i++) {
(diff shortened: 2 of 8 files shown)


src/load.c
```diff
@@ -451,10 +451,7 @@ static int load_module(xmp_context opaque, HIO_HANDLE *h)
 	test_result = load_result = -1;
 	for (i = 0; format_loaders[i] != NULL; i++) {
 		hio_seek(h, 0, SEEK_SET);
-
-		if (hio_error(h)) {
-			/* reset error flag */
-		}
+		hio_error(h); /* reset error flag */
```
AliceLR, line 454: This appears to have been intentional (don't know what purpose it served though). I think this change is the correct thing to do but I just thought that was interesting.
https://github.com/libxmp/libxmp/commit/42d4717802a9e10c41b117cd692de902e58ef6d5

src/loaders/prowizard/prowiz.c
```diff
@@ -132,10 +130,7 @@ int pw_wizardry(HIO_HANDLE *file_in, FILE *file_out, const char **name)
 		goto err2;
 	}
 
-	if (hio_error(file_in)) {
-		/* reset error flag */
-	}
-
+	hio_error(file_in); /* reset error flag */
```
AliceLR, line 133: Same with this one. https://github.com/libxmp/libxmp/commit/4781ec44162a6ed548a96386f92cb0b27f85498e

AliceLR (approved):
This looks good to me.

sezero:
https://github.com/libxmp/libxmp/blob/master/src/loaders/iff.c#L77 is `(void)`ing hio_error().

There are many other uses of hio calls which return something but the return value is ignored,
e.g. hio_seek(), so I saw no value in using confusing empty if bodies.

AliceLR:
That's strange. hio_error never seems to have been marked with `__attribute__((warn_unused_result))` so I'm not sure what the purpose of either of those hacks is (and I don't think `(void)` suppresses that anyway...).

Issue #145, opened 2019-07-18
Sagamusix "atomic_playboy.it" doesn't play correctly


knight-ryu12:
Wether it uses ~~OMPT specific hacks~~ or not, it doesn't play correctly.
(EDIT: It was somewhat obscure IT effect behavior)

1) Offset handling is buggy
"Am I a playboy" voice with offset doesn:t play correctly

2) Lower note doesn't trigger the voice

sagamusix:
It doesn't use any OpenMPT-specific hacks, but it exploits IT's behaviour of portamento + offset on the same row.

knight-ryu12:
seems like that is the case.
Hmm… wonder where the line is those behaivor critical source…

knight-ryu12:
Maybe this can be referenced with recent IT commit?

AliceLR:
This bug still exists as far as I can tell.

1) occurs at order 16h / pattern 9 and might be due to the initial note not having an offset but it's hard to tell because there's a bit going on there. A similar issue occurs in order 23h / pattern 24 and there's less going on (initial note there also lacks the offset).

2) occurs at order 19h / pattern 12 and is unrelated to portamento+offset abuse. The offender might be this (track 12, row 27h):
```
A-4 01 v00 ...
=== .. v54 ...
```

AliceLR:
Relevant portion of read_event.c for the second bug: https://github.com/libxmp/libxmp/blob/master/src/read_event.c#L1312

The cited test is not completely silent in OpenMPT and I suspect it's silent in xmp for the wrong reason (row 4 should be quiet because of two inverted chip samples playing simultaneously, not because the volume should be ignored in that situation).

AliceLR:
The bugs at order 19h / pattern 12 and order 23h / pattern 24 seem to be fixed with patch #239. This patch does nothing about the first bug, which needs more investigation (and possibly input from someone with a degree in obscure Impulse Tracker behavior).

AliceLR:
Well, here's the first issue that needs to be resolved with IT sample sustain loops: xmp overwrites the regular loop with it (this track uses a small sustain loop to create instruments from a voice sample and uses a separate bidirectional main loop to repeat the full voice sample). https://github.com/libxmp/libxmp/blob/master/src/loaders/it_load.c#L740

AliceLR:
[...]

PR #239, opened 2020-11-24
Fix NoteOffInstr.it note off volume bugs.

AliceLR:
Fixes two bugs related to note off and volume handling both found in NoteOffInstr.it (and in a couple of places in atomic_playboy.it: #145):

* Lines with both an instrument and a note off should allow the instrument volume to be used. Fixes row 4, channel 1.
* Volume events should not be ignored on a line with a note off. Fixes row 4, channel 2.


sezero:
@sagamusix, @cmatsuoka ?

sezero:
PING @sagamusix, @cmatsuoka ?

sagamusix:
Sorry, I currently don't really have the time to dive deep into the xmp replayer to understand if the fix is correct. If it plays the test case correctly, feel free to approve it!

sezero (approved):
> If it plays the test case correctly, feel free to approve it

I assume @AliceLR done thorough play-testing. Approving.

cmatsuoka (approved):
As I mentioned before, the existing implementation grew organically based on trial and error which causes the IT replay to be flawed in many aspects. To fix some of the problems we need structural changes (such as rewriting bidirectional sample processing since unrolling them beforehand doesn't work for sustain loops), but other problems can be addressed using simple yet functional fixes like this one. Generally, if we can fix existing problems without introducing regressions or increasing architectural chaos, I'm all for it.

PR #240, opened 2020-11-26
Fix xmp_set_position et al. when used during loops, pattern delay.

AliceLR:
An update/followup to #169, since we decided it'd be better to patch these separately if they were affected (which they were). When `xmp_set_position`, `xmp_next_position`, or `xmp_prev_position` (or anything else using `set_position` or these functions internally) is used, the flow variables for looping and pattern delay are now cleared. This prevents state from those effects carrying over after the position change if the position change occurs mid-row. I also added a test.

I think all of the effects that can cause flow bugs when interrupted by `xmp_set_position` should be accounted for now, but if I missed anything, let me know.


sezero:
@cmatsuoka ?

sezero:
PING @cmatsuoka ?

cmatsuoka:
Checking this. Sorry for the delay, I had kind of a tough week here.

cmatsuoka (approved):
Thanks. Resetting to a consistent and predictable state seems to be the right thing to do in such situations.

PR #241, opened 2020-11-26
Add misc. loader tests for previous pull requests.

AliceLR:
Adds loader tests for various patches I submitted without tests:

* Fix S3M ADPCM4 samples. (#163, f11ebc67)
* Fix OctaMED 'tracker compatibility' tempos. (#177, 29b5d06f)
* Fix MEDs with pattern lengths >256, MED instrument corruption. (#185, a07c760a)
* Fix Schism Tracker version date calculation. (#190, 7e20a828)

Also, I reformatted the test-dev Makefile.in loader tests list because it was getting fairly messy. I ordered them by {MOD,S3M,XM,IT}, then other loaders, then ProWizard loaders.


cmatsuoka (approved):
Thanks. I'm also planning to add test coverage reports to github so we can see what we're missing in the tests (probably a lot).

PR #243, opened 2020-11-27
Fix GDM format name ('General Digital Music', not 'Generic').

AliceLR:
This format is called "General Digital Music" but it was incorrectly named in the formats list and in the loader.

Sources:
* [GDMFORM.TXT](https://github.com/libxmp/libxmp/files/5609681/GDMFORM.TXT) packaged with Bells, Whistles, and Sound Boards.
* [MenTaLguY GDM format documentation.](https://github.com/libxmp/libxmp/blob/master/docs/formats/gdm.txt)
* [15 lines above the loader struct...](https://github.com/libxmp/libxmp/blob/master/src/loaders/gdm_load.c#L24)

(I've also seen it referred to as "General DigiMusic" before, but not in any official source...)

### DIFF
--- a/docs/formats.txt
+++ b/docs/formats.txt
@@ -85,7 +85,7 @@ XM		Fast Tracker II		1.02, 1.03, 1.04
 PC packed formats:
 
 AMF		DSMI (DMP)		0.9, 1.0, 1.1, 1.2, 1.3, 1.4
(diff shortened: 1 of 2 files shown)


cmatsuoka (approved):
Thanks Alice.

PR #247, opened 2020-12-01
Fix out-of-bounds read in IT loader caused by invalid effects.

AliceLR:
The IT loader bounds effects values from 0 to 31 but the effect translation table is shorter than that. Some IT modules use non-standard effects that were reading this array out of bounds. This patch just extends the array to size 32 (in the off chance these ever get supported...). Fixes #246.

PR #249, opened 2020-12-02
Fix effects array bounding in LIQ/IMF loaders.

AliceLR:
Fixes some more loader effects bounding issues found in the comments of #247, specifically out-of-bounds reads in `liq_load.c` and `imf_load.c`. I also added added explicit array sizes for most loaders using effects arrays (since in these loaders the array size is a functional part of the loader) and an `ARRAY_SIZE` macro so the relationship between the bounding and the effects array size is more obvious.

I'm not very familiar with Liquid Tracker or Imago Orpheus so if @debrouxl could verify that these loaders are correctly patched now that would be great.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -77,6 +77,7 @@ typedef signed long long int64;
 } while (0)
 #define MIN(x,y) ((x) < (y) ? (x) : (y))
 #define MAX(x,y) ((x) > (y) ? (x) : (y))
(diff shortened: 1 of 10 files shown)


sezero (approved):

sezero:
Applied. Thanks.

debrouxl:
I'm even less familiar with these formats than you are :)

I rebuilt with your patch, and xmp is no longer aborted by the AddressSanitizer checks when loading the previously offending samples.

PR #252, opened 2020-12-03
Fix misc. fuzzing crashes.

AliceLR:
Fixes the following bugs found in #251:
* The FNK loader could attempt to load more than 128 patterns despite having a fixed maximum pattern count of 128.
* The IT loader allows duplicate check actions in the range 0 to 3 but the array to convert DCAs to NNAs was only size 3.
* The MDL loader was missing checks to prevent loading modules with duplicate chunks.
* The MED4 loader could attempt to load instrument names of up to length 255 into an array of size 40.
* The MTM loader could attempt to load more than 32 channels despite having a fixed maximum channel count of 32.
* The PSM loader allows up to 255 samples but the sample offsets array was only size 64.
* The ULT loader was missing bounds checks on the channel count and track RLE unpacking.
* The ProWizard DI, Pha, and TP2/3 unpackers could attempt to load more than 128 patterns despite having fixed size arrays of 128.

Also fixes the following bugs I found while fixing those:
* Fix double frees when `xmp_release_module` is called twice due to dangling pointers in the xmp context not being cleared after they are freed.

### DIFF
--- a/src/hmn_extras.c
+++ b/src/hmn_extras.c
@@ -111,6 +111,7 @@ void libxmp_hmn_reset_channel_extras(struct channel_data *xc)
 void libxmp_hmn_release_channel_extras(struct channel_data *xc)
 {
 	free(xc->extra);
+	xc->extra = NULL;
 }
 
 int libxmp_hmn_new_module_extras(struct module_data *m)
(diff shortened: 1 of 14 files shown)


AliceLR:
I think some of these should probably also be added to test-dev (maybe as a new test category specifically for loader/player fuzzing results?) but considering the 4 hour turnaround on Travis, unless there are changes required here I'll make that a separate PR.

sezero (approved):

PR #253, opened 2020-12-04
Fix more fuzzing bugs, add fuzzing tests.

AliceLR:
Fixes the second round of bugs found by fuzzing in issue #251:

* ABK: reject modules with 0 instruments.
* DBM: fix out-of-bounds module name copy; fix buffer overflow in the INST handler; add duplicate chunk checks and chunk size checks where appropriate.
* Galaxy 4 J2B: fix instrument envelope point count bounds checks.
* Galaxy 5 J2B: add channel count bounds checking.
* MDL: reject tracks that attempt to reference a "previous" event when unpacking the first event.
* MED3: fix faulty pattern unpacking bounds checks.
* MMD1: fix channel count bounds checking.

Also adds test cases for these fixes and for the fixes from #252.


AliceLR:
(I'll look into why one of the ULT tests fails for MSan soon...)

AliceLR:
It was caused because the sample input has a truncated order list (the invalid channels count was probably from EOF...).

sezero (approved):

Issue #256, opened 2020-12-05
Memory errors revealed by AFL++, batch 5


debrouxl:
New issue, as requested in #251 .

**OOB reads in Prowizard loaders**:

```
Loading output/3/crashes/id:000506,sig:06,src:003211+014406,time:349403058,op:splice,rep:64 (1 of 1)
=================================================================
==2646623==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7f0b34e06ea2 at pc 0x7f0b34d400e5 bp 0x7ffd01362790 sp 0x7ffd01362788
READ of size 1 at 0x7f0b34e06ea2 thread T0
    #0 0x7f0b34d400e4 in depack_ac1d .../xmp/libxmp/src/loaders/prowizard/ac1d.c:110:16
    #1 0x7f0b34d3ddc5 in pw_wizardry .../xmp/libxmp/src/loaders/prowizard/prowiz.c:135:6
    #2 0x7f0b34d29d19 in pw_load .../xmp/libxmp/src/loaders/pw_load.c:105:6
    #3 0x7f0b34c772e0 in load_module .../xmp/libxmp/src/load.c:461:18
    #4 0x7f0b34c7664c in xmp_load_module .../xmp/libxmp/src/load.c:609:8
    #5 0x4cee43 in main .../xmp/xmp-cli/src/main.c:399:9
    #6 0x7f0b34790d09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #7 0x420679 in _start (.../bin/xmp+0x420679)

0x7f0b34e06ea2 is located 24 bytes to the right of global variable 'ptk_table' defined in 'src/loaders/prowizard/ptktable.c:4:13' (0x7f0b34e06e40) of size 74
SUMMARY: AddressSanitizer: global-buffer-overflow .../xmp/libxmp/src/loaders/prowizard/ac1d.c:110:16 in depack_ac1d
```

```
Loading output/1/crashes/id:000117,sig:06,src:004440,time:5801435,op:havoc,rep:8 (1 of 1)
=================================================================
==2540979==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7f130d1abe9e at pc 0x7f130d0fa209 bp 0x7fff33e7efd0 sp 0x7fff33e7efc8
READ of size 1 at 0x7f130d1abe9e thread T0
    #0 0x7f130d0fa208 in depack_np3 .../xmp/libxmp/src/loaders/prowizard/np3.c:114:35
[...]

PR #257, opened 2020-12-06
Add --disable-prowizard configure flag.

AliceLR:
This was requested as part of #251 (relevant portion has been split off to #256). Since there are a lot of ProWizard loaders, they're fairly niche, and they haven't been tested as thoroughly as most other loaders, it might be desirable to turn them off.

I don't feel particularly strongly about this either way (and this shouldn't be taken as an excuse to ignore vulnerabilities in the ProWizard loaders).

If this patch seems useful enough to merge, it will probably need some revision (also, I have not tested the changes to the MSVC and Watcom Makefiles...).


sezero (approved):

sezero:
Can we not do something like the following, too?
```
diff --git a/src/format.c b/src/format.c
index 398c30d..6859989 100644
--- a/src/format.c
+++ b/src/format.c
@@ -171,17 +171,17 @@ const char *const *format_list(void)
 
 	if (_farray[0] == NULL) {
 		for (count = i = 0; format_loaders[i] != NULL; i++) {
-			if (strcmp(format_loaders[i]->name, "prowizard") == 0) {
 #ifndef LIBXMP_NO_PROWIZARD
+			if (strcmp(format_loaders[i]->name, "prowizard") == 0) {
 				int j;
 
 				for (j = 0; pw_formats[j] != NULL; j++) {
 					_farray[count++] = pw_formats[j]->name;
 				}
-#endif
-			} else {
-				_farray[count++] = format_loaders[i]->name;
+				continue;
 			}
+#endif
+			_farray[count++] = format_loaders[i]->name;
 		}
 
 		_farray[count] = NULL;
```

AliceLR:
Yes, I think that's a better way to handle that. That's the main part I think needs revision... I just did the bare minimum to get the patch working there and didn't clean it up, so thank you.

PR #262, opened 2020-12-07
player: fix volume, pitch and pan slide lag

### LINKED ISSUE #242: Portamento, volume slide effects lag behind 1 frame.
I've noticed in some test modules that portamento and volume slide effects lag behind a frame (including "fine" effects that should be immediately executed during the first frame). This happens because the virtual channel volume/frequency data is configured prior to these effects being processed for the frame. https://github.com/libxmp/libxmp/blob/master/src/player.c#L1297

[vslide_lag.it.zip](https://github.com/libxmp/libxmp/files/5609611/vslide_lag.it.zip): in Impulse Tracker and OpenMPT, this module is silent. In xmp, the note plays at volume 15 for a single tick because the volume slide is applied after data is provided to the virtual channels. This also happens in MikMod (edit: unrelated. MikMod doesn't execute new-effects IT volslides on the first tick for some reason).

[portamento_offset_after_keyoff.it.zip](https://github.com/libxmp/libxmp/files/5609616/portamento_offset_after_keyoff.it.zip): in Impulse Tracker, OpenMPT, and MikMod, the notes in this module that use portamento without offset play at roughly the same time as the notes that do have an offset. In xmp they lag behind, creating a noticeable "swing" that shouldn't be there.

Panbrello is probably also affected since `update_pan` is called after `process_pan`.

I'm sure there's a reason it was implemented like this but I've confirmed that at least two players do not behave this way (Impulse Tracker, BWSB), and I'm sure there are more. Calling the update functions before the process functions improves these two test cases but predictably makes 76 tests fail. This might also subtly break other things that worked correctly before.

This seems like a serious issue, but how should we go about fixing it?

edit: I'll update this as I test different players.

| Format / tested player | Update first? | Example file |
|-|-|-|
| IT / Impulse Tracker 2.14 | Yes. | see above. |
| MOD / ProTracker 2.3, 3.15 | Yes. | [vslide_lag.mod.zip](https://github.com/libxmp/libxmp/files/5610258/vslide_lag.mod.zip) |
| S3M / Scream Tracker 3.21 | Yes. | [vslide_lag.s3m.zip](https://github.com/libxmp/libxmp/files/5609957/vslide_lag.s3m.zip) |
| XM / Fasttracker 2.09a | Yes. | [vslide_lag.xm.zip](https://github.com/libxmp/libxmp/files/5609973/vslide_lag.xm.zip) |
| DBM / DigiBooster Pro 2.13 | Yes. | Import the example MOD. |
| FAR / Farandole Composer 1.00 | Yes. | [VSLIDELG.FAR.zip](https://github.com/libxmp/libxmp/files/5610386/VSLIDELG.FAR.zip) (note: xmp handling of FAR effects also causes volslides to be much weaker than they should be, but it's enough to verify that xmp is applying it on the wrong tick.) |
| GDM / BWSB 1.21 GDMPLAY.EXE | Yes. | Use BWSB's SB16 driver since some effects are broken in the GUS driver: [VSLIDE~1.GDM.zip](https://github.com/libxmp/libxmp/files/5609959/VSLIDE.1.GDM.zip) |
| MED / OctaMED 2.00, 4.03, 5.00, SS 1.03 | Yes. | [vslide_lag.med.zip](https://github.com/libxmp/libxmp/files/5610316/vslide_lag.med.zip) (note: MED slides) |
| OKT / Oktalyzer 1.57 | Yes. | [vslide_lag.okt.zip](https://github.com/libxmp/libxmp/files/5610302/vslide_lag.okt.zip) |
| ULT / UltraTracker 1.6 | ??? | [VSLIDEX1.ULT.zip](https://github.com/libxmp/libxmp/files/5611302/VSLIDEX1.ULT.zip) and [VSLIDEX2.ULT.zip](https://github.com/libxmp/libxmp/files/5611303/VSLIDEX2.ULT.zip). Hard to tell because the volume is on a scale from 0-255. Volslide down and fine volslide down both click identically in UltraTracker. With one fine slide down, xmp clicks comparably to UltraTracker. xmp can't handle two slides down on the same frame. |

### LINKED ISSUE #260: Very high frequencies in S3M not being cut
The OpenMPT PeriodLimit.s3m test case states that: "ScreamTracker 3 limits the final output period to be at least 64, i.e. when playing a note that is too high or when sliding the period lower than 64, the output period will simply be clamped to 64. _However, when reaching a period of 0 through slides, the output on the channel should be stopped._"

Libxmp currently enforces the period limit at 64 when playing in ST3 compatible mode, but period 0 is not cutting the note as it should.

cmatsuoka:
Alice reports that effect processing for volume and pitch slides are
lagging behind one frame, and that's caused by the relevant control
variables being updated after the current volume and pitch values are
already sent to the sample mixer. This results in changes that should
happen in the first frame of each row actually becoming effective in
the second frame.
    
To fix that, we're now updating the slide variables before using them.
This caused a number of regression tests to fail, and those were
updated to use the correct reference values.
    
Fixes #242
Fixes #260
    

AliceLR:
The player.c changes look good at a glance and the examples from #242 work now. (I still need to check the tests...)

AliceLR:
I've noticed something strange with `openmpt/it/FineVolColSlide.it` that might need a followup patch: the volume of the left channel resets at the start of the first repeat of the first row. (This patch does correct the behavior on the first tick though.)

AliceLR (approved):
OK, I have finished looking through the tests. ~~The only thing I noticed that I didn't note here or in another issue was that `portamento_after_keyoff.it` retriggers a note where it should not. I think this is related to the remaining issue in #145 and might be best dealt with there.~~ Nevermind, I checked Impulse Tracker and it is correct and OpenMPT is wrong. Everything looks good to me.

PR #267, opened 2020-12-10
Makefile: add distclean target, move lib/* removal from clean to distclean

sezero:


AliceLR (approved):
This looks good to me.

PR #271, opened 2020-12-13
player: read row events only once per row

### LINKED ISSUE #264: Possible IT pattern delay volume reset bug.
I noticed this while reviewing #262—in the test `openmpt/it/FineVolColSlide.it`, the left channel's volume on row 0 appears to reset at the start of the first repeat of row 0, causing the left channel volume to always be a step higher than the right channel for the rest of the test:
```
20 0 0 0 332507 88 0 960 -128 0 255
20 0 0 1 332507 88 0 960 127 0 255
40 0 1 0 332507 88 0 960 -128 79 255
40 0 1 1 332507 88 0 960 127 79 255
60 0 2 0 332507 88 0 960 -128 159 255
60 0 2 1 332507 88 0 960 127 159 255
80 0 3 0 332507 88 0 960 -128 239 255
80 0 3 1 332507 88 0 960 127 239 255
100 0 4 0 332507 88 0 960 -128 319 255
100 0 4 1 332507 88 0 960 127 319 255
120 0 5 0 332507 88 0 960 -128 399 255
120 0 5 1 332507 88 0 960 127 399 255
140 0 0 0 332507 88 0 1024 -128 0 255
140 0 0 1 332507 88 0 960 127 0 255
160 0 1 0 332507 88 0 1024 -128 79 255
160 0 1 1 332507 88 0 960 127 79 255
180 0 2 0 332507 88 0 1024 -128 159 255
180 0 2 1 332507 88 0 960 127 159 255
200 0 3 0 332507 88 0 1024 -128 239 255
200 0 3 1 332507 88 0 960 127 239 255
220 0 4 0 332507 88 0 1024 -128 319 255
220 0 4 1 332507 88 0 960 127 319 255
240 0 5 0 332507 88 0 1024 -128 399 255
240 0 5 1 332507 88 0 960 127 399 255
```
This may or may not be intended behavior but the test description suggests it is not:
```
 Unlike fine volume slides in the effect column, fine volume slides in the
 volume column are only ever executed on the first tick — not on multiples of
 the first tick if there is a pattern delay. Thus, the left and right channel
 of this example should always have the same volume.
```

cmatsuoka:
IT fine volume slides should be applied only once per row even if the
frame counter resets to 0 due to a pattern delay (IT effect SEx).
We're using bit 1 of the rowdelay_set flag (which is a bit confusing,
perhaps we should refactor this or change the names) to mark the
real first frame of this row, so also verify it when reading new
row events.

Fixes #264


src/player.c
```diff
@@ -530,7 +530,12 @@ static inline void read_row(struct context_data *ctx, int pat, int row)
...
 
 		if (check_delay(ctx, &ev, chn) == 0) {
-			if (!f->rowdelay_set || f->rowdelay > 0) {
+			/* rowdelay_set bit 1 is set only in the first tick of the row
+			 * event if the delay causes the tick count resets to 0. We test
+			 * it to read row events only in the start of the row. (see the
+			 * OpenMPT test case FineVolColSlide.it)
+			 */
+			if (!f->rowdelay_set || ((f->rowdelay_set & 2) && f->rowdelay > 0)) {
 				libxmp_read_event(ctx, &ev, chn);
```
AliceLR, line 539: Since `f->rowdelay_set` and `f->rowdelay` are 0 at this point on the first tick of a row it might be possible to omit the ` || ((f->rowdelay_set & 2) && f->rowdelay > 0)` part altogether. I haven't verified this and the new version works as-intended so it is probably not worth changing.

AliceLR (approved):
This looks good to me. I think the handling of `f->rowdelay_set` does need to be revised but the problems with it do not affect this usage of it. Particularly, `f->rowdelay_set` bit 1 should not be cleared in `update_volume` but instead at the end of the frame. I also agree the magic bit values (vs. something like `SET_NOTE()`, `RESET_NOTE()`, et al.) are a little cheesy.

src/player.c
cmatsuoka: There's plenty of room for refactoring and clean-ups there. The role of bit 1 in rowdelay_set perhaps could be better represented by a true tick counter (which doesn't wrap at the current speed).

PR #272, opened 2020-12-13
Fix IT fine volslide/row delay bug when multiple channels use volslide.

AliceLR:
This fixes an IT row delay bug I found while reviewing #271. The row delay first frame flag was cleared in `update_volume`, meaning if multiple channels tried to fine volslide on a row with a row delay, only the first channel would actually have the slide applied. I also added macros for the IT row delay flags.


sezero:
@cmatsuoka: PING?

sezero (approved):
Looks good to me.

PR #273, opened 2020-12-16
configure.ac: update to eliminate autoconf-2.70 warnings

sezero:
also some whitespace fixes.
(run `git diff -w` to see changes without whitespace changes.)

The only remaining warning from autoconf-2.70 is this:
`configure.ac:191: warning: AC_C_BIGENDIAN should be used with AC_CONFIG_HEADERS`
That is because AC_C_BIGENDIAN actually outputs this into config.h:
```
#if defined AC_APPLE_UNIVERSAL_BUILD
# if defined __BIG_ENDIAN__
#  define WORDS_BIGENDIAN 1
# endif
#else
# ifndef WORDS_BIGENDIAN
#  undef WORDS_BIGENDIAN
# endif
#endif
```
i.e. it takes apple universal builds into account. However, if we start
using config.h, it will mean adding HAVE_CONFIG_H ifdefs into every *.c
file, therefore I didn't do AC_CONFIG_HEADERS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,4 +1,4 @@
-dnl AC_CONFIG_AUX_DIR(./scripts)
+dnl AC_CONFIG_AUX_DIR(scripts)
 AC_INIT
 AC_ARG_ENABLE(depackers, [  --disable-depackers     Don't build depackers])
 AC_ARG_ENABLE(prowizard, [  --disable-prowizard     Don't build ProWizard])
@@ -24,13 +24,13 @@ AC_DEFUN([XMP_TRY_COMPILE],[
 AC_DEFUN([AC_CHECK_DEFINED],[
   AS_VAR_PUSHDEF([ac_var],[ac_cv_defined_$1])dnl
   AC_CACHE_CHECK([for $1 defined], ac_var,
-  AC_TRY_COMPILE(,[
+  AC_COMPILE_IFELSE([AC_LANG_PROGRAM([], [[
     #ifdef $1
     int ok;
     #else
     choke me
     #endif
-  ],AS_VAR_SET(ac_var, yes),AS_VAR_SET(ac_var, no)))
+  ]])],[AS_VAR_SET(ac_var, yes)],[AS_VAR_SET(ac_var, no)]))
   AS_IF([test AS_VAR_GET(ac_var) != "no"], [$2], [$3])dnl
   AS_VAR_POPDEF([ac_var])dnl
 ])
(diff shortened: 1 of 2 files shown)


sezero:
PING?

NoSuck:
What sort of system could provide adequate review of these changes?  I am not familiar with `AC_APPLE_UNIVERSAL_BUILD`.

sezero:
> What sort of system could provide adequate review of these changes? I am not familiar with `AC_APPLE_UNIVERSAL_BUILD`.

IIRC, Apple's compiler can handle multiple `-arch xxx` options on the
same command line generating two or more objects combined into a fat
`*.o` and `AC_APPLE_UNIVERSAL_BUILD` is defined in that case.

AFAIK, there is only one place that relies on `WORDS_BIGENDIAN` macro,
and we can happily change it from compile-time endian detection to
runtime detection and removing the need for `.AC_C_BIGENDIAN`.

configure.ac
```diff
@@ -59,17 +59,17 @@ AC_CHECK_DEFINED(__clang__)
...
-       *) if test "$ac_cv_defined___clang__" = "no" || test "${LDFLAGS#*fsanitize}" = "$LDFLAGS"
-          then
-            save_LDFLAGS="$LDFLAGS"
-            LDFLAGS="$LDFLAGS -Wl,--no-undefined"
-            AC_TRY_LINK([],[],[],[LDFLAGS="$save_LDFLAGS"])
-          fi
-          ;;
+ dnl For whatever reason, the Clang sanitizers and --no-undefined for
+ dnl shared libraries are incompatible.
+ *) if test "$ac_cv_defined___clang__" = "no" -o "${LDFLAGS#*fsanitize}" = "$LDFLAGS"
```
AliceLR, line 66: FWIW I used two `test`s with `||` instead of `-o` for this because, when I was working with a set of shell scripts for something else, [ShellCheck](https://github.com/koalaman/shellcheck) emitted warnings that `-o` can potentially be problematic. I have never personally encountered issues with `-o` so if this fixes an autoconf warning I'm OK with this. The other changes look good from a glance but I can verify them soon.

AliceLR (approved):
Everything looks good here.

I also checked into the usage of `WORDS_BIGENDIAN`. It is used in four places: the IT loader uses it to set a sample flag; both the sample loader and the OXM depacker use it to determine when to byte flip samples; the MD5 code uses it to decide between performing a memcpy and a copy with byte flips. Runtime checks would probably not significantly affect the performance of any of these and would probably not add much bloat. (There is also a short macro I found that can turn this into a compile time check for some compilers, but it requires C99.)

Another option (if a universal Mac build supporting PPC isn't required... which it probably isn't, and if it is, lipo) is to use custom actions for `AC_C_BIGENDIAN`, which might turn off that warning.

sezero:
Restored the two-tests version instead of `-o`. Will merge as soon as travis completes, thanks.

AliceLR:
To be clear I don't feel particularly strongly about that, I was just explaining my reasoning for it (because the way I did that does look unusual).

sezero:
>  (There is also a short macro I found that can turn this into a compile time check for some compilers, but it requires C99.)

Which macro is it?

AliceLR:
> Which macro is it?

From [this Stack Overflow thread](https://stackoverflow.com/questions/4239993/determining-endianness-at-compile-time), here are two C99 macros for this:

```C
#define IS_BIG_ENDIAN (*(uint16_t *)"\0\xff" < 0x100)
```
```C
#define I_AM_LITTLE (((union { unsigned x; unsigned char c; }){1}).c)
```

(I haven't taken the time to thoroughly verify these...)

NoSuck:
After thinking this over, it seems as if we are fighting autoconf instead of getting it to work for us.

>However, if we start using config.h, it will mean adding HAVE_CONFIG_H ifdefs into every *.c file...

Why every `*.c` file and not simply when necessary?  See [src/loaders/abk_load.c](https://github.com/libxmp/libxmp/blob/master/src/loaders/abk_load.c#L25) for example.

Perhaps we should use `AC_C_BIGENDIAN` after all.

sezero:
> After thinking this over, it seems as if we are fighting autoconf instead of getting it to work for us.

Looks like it :)

> > However, if we start using config.h, it will mean adding HAVE_CONFIG_H ifdefs into every *.c file...
> 
> Why every `*.c` file and not simply when necessary? See [src/loaders/abk_load.c](https://github.com/libxmp/libxmp/blob/master/src/loaders/abk_load.c#L25) for example.

Because if we tell autoconf to generate a config.h, all configure-generated macros
like `HAVE_XXX` will go into config.h, and IIRC many sources need them.

I am not against using a config.h, though:  I just need confirmation to do so, and I
can do that.

PR #274, opened 2021-01-03
API Addition: xmp_add_channels

NoSuck:
This enhances libxmp's API as discussed long ago in #100.

The test unit remains untested, because tests no longer successfully build on my system.  (I may open an issue for that.)

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -648,6 +648,43 @@ void xmp_end_player(xmp_context c)
 
     PageBreak
 
+Module editing
+~~~~~~~~~~~~~~
+
+.. _xmp_add_channels():
+
+int xmp_add_channels(xmp_context c, int start, int num)
+```````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.6]* Adds channels to the module. This function
+  currently only supports MOD, S3M, and IT formats, resulting in
+  undefined behavior when called on contexts of others.
+
+  **Parameters:**
+    :c: the player context handle.
+
+    :start: the channel index from which to displace existing channels.
+      Negative indices count in reverse such that -1 refers to the first
+      non-existing index (i.e. the number of channels).
+
+    :num: the number of channels to add. Negative numbers remove
+      channels and are clamped if they breach module dimensions.
+
+  **Returns:**
+    0 if channels were added successfully, ``-XMP_ERROR_INVALID``
+    if arguments were invalid, ``-XMP_ERROR_SYSTEM`` if allocation
+    failed, or ``-XMP_ERROR_INTERNAL`` for any other error.
+
+  **Examples:** ::
+
+    xmp_add_channels(c, -1, 2); /* Append two channels */
+    xmp_add_channels(c, 0, -1); /* Remove the first channel */
+    xmp_add_channels(c, 1, 3); /* Insert three channels after the first */
+
+.. raw:: pdf
+
+    PageBreak
+
 Player control
 ~~~~~~~~~~~~~~
 
--- a/libxmp.map
+++ b/libxmp.map
@@ -70,3 +70,8 @@ XMP_4.5 {
     xmp_test_module_from_file;
     xmp_test_module_from_memory;
 } XMP_4.4;
+
+XMP_4.6 {
+  global:
+    xmp_add_channels;
+} XMP_4.5;
--- a/src/player.c
+++ b/src/player.c
@@ -1802,8 +1802,8 @@ void xmp_end_player(xmp_context opaque)
 	int i;
 #endif
 
-	//if (ctx->state < XMP_STATE_PLAYING)
-	//	return;
+	if (ctx->state < XMP_STATE_PLAYING)
+		return;
 
 	ctx->state = XMP_STATE_LOADED;
 
--- a/src/player.c
+++ b/src/player.c
@@ -1926,3 +1926,121 @@ void xmp_get_frame_info(xmp_context opaque, struct xmp_frame_info *info)
 		}
 	}
 }
+
+int xmp_add_channels(xmp_context opaque, int start, int num)
+{
+	struct context_data *ctx = (struct context_data *)opaque;
+	struct module_data *m = &ctx->m;
+	struct player_data *p = &ctx->p;
+	struct xmp_module *mod = &m->mod;
+	int pat, c, t;
+
+	if (num == 0)
+		return 0;
+	if (num > 0) {
+		if (mod->chn + num > XMP_MAX_CHANNELS)
+			return -XMP_ERROR_INVALID;
+	}
+	else {
+		if (mod->chn + num < 0)
+			return -XMP_ERROR_INVALID;
+	}
+
+	if (start < 0)
+		start = mod->chn + 1 + start;
+	if (start < 0 || start > mod->chn)
+		return -XMP_ERROR_INVALID;
+
+	if (num < 0) {
+		if (start - num > mod->chn)
+			return -XMP_ERROR_INVALID;
+		/* Shift tracks and channels back; resize patterns' track index arrays */
+		for (pat = 0; pat < mod->pat; pat++) {
+			mod->xxp[pat] = realloc(mod->xxp[pat], sizeof (struct xmp_pattern) + sizeof (int) * (mod->chn + num - 1));
+			if (mod->xxp[pat] == NULL)
+				return -XMP_ERROR_SYSTEM;
+			for (c = 0; c < mod->chn + num; c++) {
+				t = pat * (mod->chn + num) + c;
+				mod->xxp[pat]->index[c] = t;
+				mod->xxt[t] = mod->xxt[pat * mod->chn + c - (c >= start ? num : 0)];
+			}
+		}
+		for (c = start; c < mod->chn + num; c++) {
+			mod->xxc[c].pan = mod->xxc[c - num].pan;
+			mod->xxc[c].vol = mod->xxc[c - num].vol;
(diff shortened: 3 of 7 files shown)


AliceLR:
> The test unit remains untested, because tests no longer successfully build on my system. (I may open an issue for that.)

Please do open an issue for this if you can't get it to work. In my experience sometimes I have to `make clean`/`autoconf`/`./configure` in the test-dev folder again because something inexplicably breaks (usually due to switching sanitizers).

Travis was able to run the test fortunately and detected some leaks that may just be caused by a missing `xmp_release_module`.

I will look at this more closely when I get a chance, thanks.

NoSuck:
The tip for `test-dev` worked.  Thank you.

I fixed the test unit's missing `xmp_release_module`.  Here's hoping that does the trick.

NoSuck:
That did *not* do the trick, so I will be using this branch to test with Travis CI.  Please pardon the mess...

docs/libxmp.rst
```diff
@@ -648,6 +648,43 @@ void xmp_end_player(xmp_context c)
...
 
+Module editing
+~~~~~~~~~~~~~~
+
+.. _xmp_add_channels():
+
+int xmp_add_channels(xmp_context c, int start, int num)
+```````````````````````````````````````````````````````
+
+  *[Added in libxmp 4.6]* Adds channels to the module. This function
```
sezero, line 659: Why 4.6?  4.5 isn't released yet.

libxmp.map
```diff
@@ -70,3 +70,8 @@ XMP_4.5 {
     xmp_test_module_from_file;
     xmp_test_module_from_memory;
 } XMP_4.4;
+
+XMP_4.6 {
```
sezero, line 74: Why 4.6? 4.5 isn't released yet.

NoSuck: Because discussion elsewhere seemed to indicate a desire to freeze features for an impending 4.5 release.  Personally, I would be happy to include `xmp_add_channels` (and other, forthcoming additions) in version 4.5 but would require input from others before doing so.

NoSuck:
Memory is leaked because `libxmp_virt_off` does not free anything and `libxmp_virt_on` allocates unconditionally.  I will spend some time thinking about this.

NoSuck:
Travis CI seems to be malfunctioning now.  Look at all of the [false errors](https://travis-ci.org/github/libxmp/libxmp/builds/752969423) in 0fb483c, for example.  I cannot make progress investigating the supposed memory leak until this issue is resolved.

If anyone would provide insight, I would appreciate it.  If I don't hear anything, I might open a separate issue for Travis later.

AliceLR:
It is not malfunctioning, but rather something is being double-freed due to the changes in src/load.c and src/player.c (reproduced locally with ASan).

NoSuck:
Thank you for the extra eyes.  I apologize for the false alarm.

AliceLR:
> Thank you for the extra eyes. I apologize for the false alarm.

No problem. Once I'm done cleaning up reports from #256 I'll look for and fix potential double frees, since clearly I missed some in b2e6499aa. While it surfaced because of removing some state checks there's really no excuse for libxmp not to immediately NULL any freed pointers (`libxmp_virt_off` lines 178 and 179).

NoSuck:
I completely agree about setting freed pointers to NULL.

Is there a way for us to deal with Travis errors like [this one](https://travis-ci.org/github/libxmp/libxmp/jobs/753117913) (from the latest commit here)?

AliceLR:
> Is there a way for us to deal with Travis errors like [this one](https://travis-ci.org/github/libxmp/libxmp/jobs/753117913) (from the latest commit here)?

I've seen errors like that before (here's [something similar that was occurring on some Mac builds](https://travis-ci.org/github/libxmp/libxmp/jobs/743831475)...) and I think they happen completely at random.

libxmp.map
AliceLR: There was discussion for a 4.5 release but for whatever it was never actually tagged. I think shortly after that the fuzzing crash reports started coming in so there hasn't been much of an opportunity to freeze things since. I agree that this should be considered a 4.5 API function.

src/player.c
```diff
@@ -1926,3 +1926,120 @@ void xmp_get_frame_info(xmp_context opaque, struct xmp_frame_info *info)
...
+	}
+
+	if (start < 0)
+		start = mod->chn + 1 + start;
+	if (start < 0 || start > mod->chn)
+		return -XMP_ERROR_INVALID;
+
+	if (num < 0) {
+		if (start - num > mod->chn)
+			return -XMP_ERROR_INVALID;
```
AliceLR, line 1956: Should `start + num < 0` be checked here? That would ensure that a number of channels larger than the position they are being removed from can not be removed. I don't know that would actually hurt anything though.

I'm not sure a check for `start - num > mod->chn` is necessary since the following loops are being bounded against `mod->chn + num`. But, for example, if you're removing 8 tracks starting from position 10 of a 12 track module, this check would result in 18 > 12 and fail (when that operation should be completely valid).

src/player.c
```diff
@@ -1926,3 +1926,120 @@ void xmp_get_frame_info(xmp_context opaque, struct xmp_frame_info *info)
...
+		/* See xmp_start_player */
+		//p->xc_data = realloc(p->xc_data, p->virt.virt_channels * sizeof (struct channel_data));
+		//if (p->xc_data == NULL)
+		//	return -XMP_ERROR_SYSTEM;
+	}
+
+	libxmp_virt_off(ctx);
+	/* See xmp_play_frame */
+	//libxmp_virt_reset(ctx);
+	//reset_channels(ctx);
```
AliceLR, line 2043: I *think* an `xmp_end_player` call here might be sufficient to reset virtual channels/`p->xc_data`/`f->loop` but it would require that the user manually calls `xmp_start_player` again (since the rate and format arguments do not seem to be stored). Alternatively, if this is being allowed in the playing state without stopping the playing state, I think `libxmp_virt_reset` and `reset_channels` might be correct (honestly, I'm not sure and I'm still figuring out how this part of libxmp works).

src/player.c
```diff
@@ -1926,3 +1926,120 @@ void xmp_get_frame_info(xmp_context opaque, struct xmp_frame_info *info)
...
+		}
+
+		/* This is needed to get actual speaker output when in instrument mode! */
+		libxmp_virt_off(ctx);
+		if (libxmp_virt_on(ctx, mod->trk) != 0)
+			return -XMP_ERROR_INTERNAL;
+		/* See xmp_start_player */
+		//p->xc_data = realloc(p->xc_data, p->virt.virt_channels * sizeof (struct channel_data));
+		//if (p->xc_data == NULL)
+		//	return -XMP_ERROR_SYSTEM;
```
AliceLR, line 2037: This should definitely be done if `xmp_end_player` isn't being used unconditionally here and if this function is allowed after `xmp_start_player`. Additionally, `f->loop` needs to be reallocated.

test-dev/test_api_add_channels.c
```diff
@@ -0,0 +1,40 @@
...
+
+	/*
+	ret = xmp_add_channels(opaque, 0, -1);
+	fail_unless(ret == 0, "unable to remove a channel");
+	*/
+
+	/*
+	ret = xmp_add_channels(opaque, 0, XMP_MAX_CHANNELS + 1);
+	fail_unless(ret == -XMP_ERROR_INVALID, "erroneously succeeded for invalid number of channels");
+	*/
```
AliceLR, line 35: These tests should definitely be uncommented (presumably they were not working due to the sanity checks I commented above?).

AliceLR (changes requested):
Here's my initial feedback from code inspection. Note: I haven't actually tested these recommendations locally and they may be useless :(

src/player.c
AliceLR: Also since I forgot to comment on this, the return value from realloc probably shouldn't be stored directly to `p->xc_data` since if it does return NULL the original allocation will leak.

src/player.c
```diff
@@ -1926,3 +1926,120 @@ void xmp_get_frame_info(xmp_context opaque, struct xmp_frame_info *info)
...
+	}
+	else {
+		if (mod->chn + num < 0)
+			return -XMP_ERROR_INVALID;
+	}
+
+	if (start < 0)
+		start = mod->chn + 1 + start;
+	if (start < 0 || start > mod->chn)
+		return -XMP_ERROR_INVALID;
```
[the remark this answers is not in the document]
AliceLR: Nevermind, I completely misunderstood this. :(

src/player.c
NoSuck: Your suspicions are correct.  Testing was performed in real time while a module played.  If I remember correctly, `reset_channels` was required to prevent playback from being interrupted.

More importantly, however, I apologize for not making it clearer [here](https://github.com/libxmp/libxmp/pull/274#issuecomment-753867642) that 9e74e75 is the commit to evaluate, as subsequent commits basically comprise experimentation involving Travis, as I am currently unable to reproduce the original leak locally.  This includes the commented-out lines here (and elsewhere, too, of course).

NoSuck:
I have decided to close this but would like to reopen a PR later, after setting up Travis CI locally.  (Hopefully that is not difficult; I have yet to look into it.)  Either way, it is wasteful and distracting to use a public PR for such testing.  I apologize for the mess thus far.

To anyone watching:  Libxmp checks off many important boxes needed for authoring a tracker, not only thanks to its powerful yet simple API but also because note data basically sits there in an array of `unsigned char` practically begging you to play with it 😉 I made this PR in hopes of allowing others to do so more easily in the future.

PR #282, opened 2021-01-06
xmp_set_tempo_factor fix

### LINKED ISSUE #276: xmp_set_tempo_factor
Newly added `xmp_set_tempo_factor()` (commit b1fb1ee84a1c28) has a
sanity check for `XMP_MAX_FRAMESIZE`, but it uses `m->time_factor`
and without assigning the new `val` to it.

Do we not need something like the following for it?
```
diff --git a/src/control.c b/src/control.c
index ca1422a..fba2ae4 100644
--- a/src/control.c
+++ b/src/control.c
@@ -579,7 +579,7 @@ int xmp_set_tempo_factor(xmp_context opaque, double val)
 	}
 
 	val *= 10;
-	ticksize = s->freq * m->time_factor * m->rrate / p->bpm / 1000 * sizeof(int);
+	ticksize = s->freq * val * m->rrate / p->bpm / 1000 * sizeof(int);
 	if (ticksize > XMP_MAX_FRAMESIZE) {
 		return -1;
 	}
```

The new function is also missing documentation in docs/libxmp.rst.

sezero:
- fixes XMP_MAX_FRAMESIZE sanity check.
- adds documentation for it

Closes #276.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -711,6 +711,21 @@ int xmp_set_row(xmp_context c, int row)
     The new row, ``-XMP_ERROR_INVALID`` if the new row is invalid or
     ``-XMP_ERROR_STATE`` if the player is not in playing state.
 
+.. _xmp_set_tempo_factor():
+
+int xmp_set_tempo_factor(xmp_context c, double val)
+```````````````````````````````````````````````````
+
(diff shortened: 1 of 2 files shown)


AliceLR (approved):
These changes look good to me, nice job catching this.

Issue #280, opened 2021-01-06
Memory errors revealed by AFL++, batch 7


debrouxl:
With HEAD at dbcb4d5ee9278256658f663c286025e55a10c3e0 + a manual application of the patch from #279 as it stands at the time of this writing, about half of the unique issues which were still showing up in my previous build have disappeared, and the size of the tarball of remaining crash-inducing samples has been cut by more than two thirds :)
However, I can still see 8 different stack traces, 2-3 of which appeared in the past 3 days. 3 of these seem to occur outside Prowizard loader files, which probably warrants a new issue.

```
Loading output/1/crashes/id:000135,sig:06,src:004753,time:6028808,op:havoc,rep:8 (1 of 1)
=================================================================
==257070==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000000410 at pc 0x7ffb337e7f45 bp 0x7ffd3672e590 sp 0x7ffd3672e588
READ of size 4 at 0x602000000410 thread T0
    #0 0x7ffb337e7f44 in depack_p10c .../xmp/libxmp/src/loaders/prowizard/pm10c.c:180:5
    #1 0x7ffb337c9d63 in pw_wizardry .../xmp/libxmp/src/loaders/prowizard/prowiz.c:112:6
    #2 0x7ffb337b5f69 in pw_load .../xmp/libxmp/src/loaders/pw_load.c:81:6
    #3 0x7ffb33703540 in load_module .../xmp/libxmp/src/load.c:461:18
    #4 0x7ffb337028ac in xmp_load_module .../xmp/libxmp/src/load.c:609:8
    #5 0x4cee43 in main .../xmp/xmp-cli/src/main.c:399:9
    #6 0x7ffb3321bd09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #7 0x420679 in _start (.../bin/xmp+0x420679)

Address 0x602000000410 is a wild pointer.
SUMMARY: AddressSanitizer: heap-buffer-overflow .../xmp/libxmp/src/loaders/prowizard/pm10c.c:180:5 in depack_p10c
```

```
Loading output/1/crashes/id:000512,sig:06,src:017168+012614,time:442942140,op:splice,rep:4 (1 of 1)
=================================================================
==260936==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffc796d9b60 at pc 0x7febe96533b7 bp 0x7ffc796d9410 sp 0x7ffc796d9408
[...]

PR #283, opened 2021-01-07
Fix out-of-bounds reads/writes in ST and MASI loaders, SQSH depacker.

AliceLR:
Fixes the following bugs found in #280:
* The ST .MOD loader test function allows invalid instrument numbers in some cases, but then could use the invalid number to index the instruments array when counting the total file size.
* The MASI PSM loader could get stuck in an infinite loop trying to parse a truncated SONG chunk.
* The MASI PSM loader was missing a sanity check on the maximum number of orders and could overflow the buffer used to store pattern names for the order list.
* The SQSH depacker was completely missing sanity checks while parsing an input bitstream, which could allow it to read beyond the end of the input buffer.


sezero (approved):

sezero:
These look good.

However, you need one missing sanity check in unsqsh.c: patch and
a related crasher (once provided by Lionel Debroux) are attached.
[fuzz.zip](https://github.com/libxmp/libxmp/files/5780970/fuzz.zip)
[patch.txt](https://github.com/libxmp/libxmp/files/5780973/patch.txt)

sezero:
Will be applying unsqsh parts of this PR to libmikmod, soon.

AliceLR:
Nice catch; I hadn't got to testing most of the other crashers since I assumed they were mostly ProWizard related. I will apply this patch/test and force push soon.

sezero:
>  this is from a different fuzzing set?

Yes, it was from a set Lionel Debroux sent to me about a year and a half ago for libmikmod.
I was lazy to work on the xpk parts, I now did :)

Issue #294, opened 2021-01-11
Slow loading files revealed by AFL++, batch 1


debrouxl:
For once, I had a look at "hangs" produced by the fuzzer instances. I seldom do, even if I remember doing it years ago for module files already; nevertheless, there are sometimes interesting tidbits there :)

Attached are a set of (partially redundant, as usual) files whose loading consumes a really excessive amount of time, namely 15s before SIGKILL struck (`timeout -k 5s 10s ...`). Some of the files, e.g. `output/1/hangs/id:000022,src:002068,time:2728441,op:havoc,rep:32` and its siblings, yield _excessive_ terminal output. By _excessive_, I mean that the simple script containing a loop which iterates on the ~500 hanging files produced by the fuzzers generated about **10 GB of terminal output**. You have been warned ;)

FTR, there's worse elsewhere. I recently stumbled across a DoS caused by excessive terminal output, about two orders of magnitude worse. The other program, which is now fixed, produced a total of 176 GB (!) of output, ~180 MB after compression with xz. I wasn't even trying to fuzz it, it's just that I happened to run it on hardware which caused the program to receive _weird_ input, against which it wasn't (fully) prepared...

[xmp_remaining_hangs_20210111_1658.tar.gz](https://github.com/libxmp/libxmp/files/5797160/xmp_remaining_hangs_20210111_1658.tar.gz)

EDIT: thousands of `unknown effect ff ff` (a message which can, under that form, be produced only by the gal4 and gal5 loaders, AFAICS) could be caused by bugs similar to recently fixed bugs: reading an unbounded, or poorly bounded, amount of data past EOF, and interpreting EOF as valid data.

AliceLR:
I will look into these soon, thanks. Considering that MASI bug and some other unusually slow files I've encountered, I'm not surprised there are more of these.

AliceLR:
Progress here: https://github.com/libxmp/libxmp/compare/master...AliceLR:fix-fuzzing-hangs

All of the following tests hanged due to a lack of error checks in the ZIP depacker's store handling:
```
	1/id:000004,src:000311,time:483868,op:havoc,rep:16
	1/id_000005,src_000311,time_485230,op_havoc,rep_4
	1/id_000006,src_000311,time_486255,op_havoc,rep_2
	1/id_000007,src_000311,time_492699,op_havoc,rep_8
	1/id_000008,src_000311+005697,time_497946,op_splice,rep_64
[...]

PR #295, opened 2021-01-13
Fix hangs and slow loading file bugs from AFL++ fuzzing.

AliceLR:
Fixes most (but not all) of the problems found in #294.

Fixed bugs affecting multiple formats:
* `scan_module`: tempo changes and other effects would prevent breaking the scan when >512 rows are played in a single pattern, causing an infinite loop.
* `scan_module`: modules with no valid orders could never break out of the scan loop.
* `scan_module`: infinite pattern loop detection (see `storlek_11.it`) only worked due to the scan count for the current row incidentally being the correct value when the >512 rows check breaks the scan. Other modules could still hang in situations like this, so now there's an explicit check to see if the current row's scan count has overflowed.
* `libxmp_load_sample`: when attempting to load a sample from file, if the file is currently positioned at the EOF, loading is skipped. This isn't a foolproof fix but it reduces the slowness and RAM consumption of a few inputs by quite a bit. (Samples that would run over the EOF but aren't at it could also be trimmed to help with this.)

Fixed bugs in loaders:
* `amf_load.c`: missing EOF checks in various places.
* `dbm_load.c`: missing EOF checks in `get_patt`.
* `fnk_load.c`: added a sanity check to make sure reported sample lengths aren't larger than the reported module length.
* `gal4_load.c`: missing EOF check in `get_patt`, usage of `printf` instead of `D_`.
* `gal5_load.c`: missing EOF check in `get_patt`, usage of `printf` instead of `D_`.
* `gdm_load.c`: missing EOF checks in the pattern scan and pattern load loops.
* `it_load.c`: missing EOF checks in pattern scan and pattern load loops.
* `masi_load.c`: scan of subchunks in `get_song_2` could get caught in an infinite seeking loop due to very large chunk sizes being interpreted as negative sizes and not being filtered out.
* `mod_load.c`: missing `hio_read` return value checks could result in uninitialized reads.
* `stx_load.c`: missing EOF check in the pattern load loop.

Fixed bugs in depackers:
* `gunzip.c`: missing EOF/error checks in the name and comment skipping loops could cause hangs.
* `inflate.c`: multiple missing EOF/error checks in `decompress` could cause hangs.
* `unzip.c`: missing EOF/error checks in the store handling could cause hangs.


sezero (approved):

sezero:
With this I'm getting a warning from scan.c which hasn't been there before:
```
src/scan.c: In function ‘scan_module’:
src/scan.c:57: warning: ‘row’ may be used uninitialized in this function
```
This is with my old gcc4.4.7.

Other than that, looks good.

AliceLR:
> With this I'm getting a warning from scan.c which hasn't been there before:
> 
> ```
> src/scan.c: In function ‘scan_module’:
> src/scan.c:57: warning: ‘row’ may be used uninitialized in this function
> ```
> 
> This is with my old gcc4.4.7.
> 
> Other than that, looks good.

I was able to reproduce this with GCC 4.9.5 using -O1 and it's because I should have used `break` instead of `goto end_module` in the invalid orders loop breaking check. This should be fixed now.

src/scan.c
```diff
@@ -116,11 +117,19 @@ static int scan_module(struct context_data *ctx, int ep, int chain)
...
-    gvol_memory = break_row = row_count = frame_count = 0;
+    gvol_memory = break_row = row_count = row_count_total = frame_count = 0;
+    orders_since_last_valid = any_valid = 0;
     start_time = time = 0.0;
     inside_loop = 0;
 
     while (42) {
+	/* Sanity check to prevent getting stuck due to broken patterns. */
+	if (orders_since_last_valid > 512) {
+	    D_(D_CRIT "orders_since_last_valid = %d @ ord %d; ending scan", orders_since_last_valid, ord);
```
sezero, line 128: Should this imply a failure?  I.e. should we set `any_valid` to 0 here?

AliceLR: I don't think so—some valid tracks (particularly ones containing a lot of loops) are capable of reaching this. This OctaMED module should loop for something like an hour(!) and encounters this check, but it is a valid module.

https://modland.com/pub/modules/OctaMED%20MMD1/Apollon/pilot%20i%20tyskland.mmd1

AliceLR: Sorry, I deleted a comment I made thinking this was the other >512 check. I think it theoretically might be possible to cause this check to occur and still have a partial valid module. I'm not entirely sure.

sezero: Will you investigate more, or should we merge?

AliceLR: I tried yesterday to create a test case that could cause that state to occur but failed to, so I think I'll probably have to step through the code. You made a valid point and what I suspect could occur is that if this check is flagged but allows the scan to succeed (`any_valid`=1), the player scan could later get stuck in an infinite loop (a problem I encountered before adding `any_valid` at all). So the solution may be to remove `any_valid` and just unconditionally return -1 here.

AliceLR: Okay, after some experimentation, I think that the current implementation should be fine—as long as the sequence has a valid entry point it won't get stuck in an infinite loop in `next_order` and `any_valid` makes sure that the entry point (eventually) reaches something valid.

sezero: OK then.

PR #296, opened 2021-01-13
Add style.css into html docs generation. 

LossyDragon:
The SourceForge version of the docs had a style.css linked within it. This recovers that stylesheet and embeds it into the generated html doc.


AliceLR (approved):
Wow, this looks a lot better than what was being output before. Thanks for recovering this.

PR #304, opened 2021-01-22
Fix dist-package check fails due to missing python-pil package.

AliceLR:
The GitHub dist-package check has been failing on my fork branches due to a 404 error when attempting to download the package `python-pil`. The "fix" is to just run `sudo apt update` before attempting to install any packages. I also added this to the emscripten check even though it wasn't having this issue.

edit: [example failed check](https://github.com/AliceLR/libxmp/runs/1745452638).


sezero (approved):
OK

PR #305, opened 2021-01-22
Fix hangs and leaks found by AFL++ fuzzing (hang batch 2).

### LINKED ISSUE #301: DoS (and smaller leaks) revealed by AFL++, batch 2
The fourth afl-fuzz instance I started recently for #299 is also doing much better with hangs or excessive memory consumption, thanks to #295 :)
However, it has started producing new samples whose behaviour is an outlier for both loading time - which is the proximate cause of afl-fuzz saving them - and memory consumption - which is closer to the ultimate cause of probably missing more sanity checks. The OOM killer correctly takes care of ASAN-instrumented libxmp+xmp within a few seconds of these samples markedly degrading system performance, at a point they already consume more than 4 GB of RAM... so beware, even though the consumption of a non-instrumented build should be smaller ;)
Years ago, I remember openmpt consuming more than 5 GB of RAM and seconds of CPU time loading a specially crafted sample, before getting killed by the OOM killer.

Also, the ASAN-instrumented build reports that 3 of these new samples, so far, cause 2 different memory leaks; I went looking in the logs, and found 2 older samples which cause the same memory leak when loading a specially crafted Scream Tracker 1.10 module:
```
Loading output/2/hangs/id:000172,src:019399,time:3986356923,op:havoc,rep:4 (1 of 1)
Module name  : Noname
Module type  : Scream Tracker 1.10
Module length: 0 patterns
Patterns     : 257
Instruments  : 32
Samples      : 32
Channels     : 4 [ 4 b b 4 ]
Duration     : 0min00s

==3843688==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1264640 byte(s) in 63232 object(s) allocated from:
    #1 0x7fb08b89eea0 in libxmp_alloc_pattern .../xmp/libxmp/src/loaders/common.c:110:18
    #2 0x7fb08b89eea0 in libxmp_alloc_pattern_tracks .../xmp/libxmp/src/loaders/common.c:158:6

SUMMARY: AddressSanitizer: 1264640 byte(s) leaked in 63232 allocation(s).
```

The other leak is:
```
Loading output/4/hangs/id:000023,src:001184,time:126933623,op:havoc,rep:4 (1 of 1)
.../bin/xmp: output/4/hangs/id:000023,src:001184,time:126933623,op:havoc,rep:4: Error loading module

==3899276==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2401408 byte(s) in 5 object(s) allocated from:
    #1 0x7f4e560044b4 in libxmp_alloc_subinstrument .../xmp/libxmp/src/loaders/common.c:84:20

SUMMARY: AddressSanitizer: 2401408 byte(s) leaked in 5 allocation(s).
```

[xmp_remaining_hangs_20210118_1100.tar.gz](https://github.com/libxmp/libxmp/files/5829546/xmp_remaining_hangs_20210118_1100.tar.gz)

[xmp_memory_leaks_20210118_1108.tar.gz](https://github.com/libxmp/libxmp/files/5829548/xmp_memory_leaks_20210118_1108.tar.gz)

EDIT: note that the tests have been performed with the patch from #300 manually applied.

AliceLR:
Fixes most hangs and leaks associated with #301. The patch to the sample loader might require some revision if the old handling for samples that extend past EOF is desired in some cases (maybe allow samples with length <65536?) but this general case is required to prevent out-of-memory situations from being possible.

* Fixes missing pattern count bounds check in the STM loader that would cause libxmp to leak pattern data due to eventually CLAMPing the pattern count.
* Fixes a leak in the Galaxy Music System 5.0 loader caused by not checking for duplicate instruments.
* Fixes hangs in stb_vorbis caused by a lot of missing EOF checks in `start_decoder`. I checked upstream stb_vorbis and these checks are also missing there, so this particular patch should probably be upstreamed.
* Fixes hangs and high memory consumption caused by samples with a valid starting offset but large invalid sizes extending beyond EOF.


debrouxl:
With that patch applied, there's a clear improvement, as usual:
* none of the 500+ files kept as hangs by the afl-fuzz instances takes more than ~3s to load on the same computer, and the whole set is handled in ~22s;
* no more leaks occur when dealing with that set of files;
* only 14 files trigger a maximum memory resident set higher than 100 MB, and among those, 3 exceed 1 GB. The rest are between ~8 and ~27 MB.

Here are the 14 memory consumption outliers, which are also CPU time outliers, plus a 15th file, which is a CPU time outlier (second slowest load time, depacking ends up failing certainly because the data is very corrupted) but whose RAM consumption is uneventful, a few hundreds of KBs higher than the lowest ones.
[xmp_higher_ram_consumption_20210123_1122.tar.gz](https://github.com/libxmp/libxmp/files/5860029/xmp_higher_ram_consumption_20210123_1122.tar.gz)

AliceLR:
Thanks for the new tarball—I unlinked this issue from #301 so it will stay open until I get a chance to check those out.

sezero:
> The patch to the sample loader might require some revision if the old handling for samples that extend past EOF is desired

Forgive my ignorance but is it possible to have samples that extend past EOF? Why would such handling be required?

AliceLR:
> Forgive my ignorance but is it possible to have samples that extend past EOF? Why would such handling be required?

Some old MODs and potentially other formats are only available with the sample data truncated, so rather than reject them the partial samples are loaded. The change is in how the part after the truncation is handled: before it would allocate the sample with its full length and zero the missing portion (allowing bad inputs to allocate massive amounts of RAM). The changed version shortens the sample length before attempting to load to prevent that.

sezero (approved):
OK

PR #306, opened 2021-01-29
Fix undetected song loops of one-pattern songs.

NoSuck:
Another fringe case.

### DIFF
--- a/src/control.c
+++ b/src/control.c
@@ -211,6 +211,10 @@ int xmp_set_row(xmp_context opaque, int row)
 	p->frame = -1;
 	f->num_rows = mod->xxp[mod->xxo[p->ord]]->rows;
 
+	/* Because we skipped the decrement in check_end_of_module. */
+	if (f->end_point && row)
+		f->end_point--;
+
 	return row;
 }
 


src/control.c
```diff
@@ -211,6 +211,10 @@ int xmp_set_row(xmp_context opaque, int row)
 	p->frame = -1;
 	f->num_rows = mod->xxp[mod->xxo[p->ord]]->rows;
 
+	/* Because we skipped the decrement in check_end_of_module. */
+	if (f->end_point && row)
+		f->end_point--;
+
```
AliceLR, line 217: I might be misunderstanding the purpose of `f->end_point`, but it seems to represent the number of times the sequence end order+row needs to be seen before the sequence is considered to have "looped". `check_end_of_module` only ever checks/decreases this value when the current order and row are equal to the sequence end order and row, respectively, so decrementing it regardless of those seems slightly suspicious (I checked the `set_position` behavior and it also seems strange—it assumes that the sequence looping order is always the lowest order in the sequence, which I'm not sure is necessarily true...).

That said, I don't know that there's really a "right" answer to loop detection when the player is jumping around to arbitrary orders/rows.

NoSuck: Your understanding of `f->end_point` is correct, as far as I can tell.  However, I have since realized additional issues:  `xmp_set_row` can be called multiple times in succession, decrementing `end_point` each time, for example--which would be totally wrong.  I will look for a better solution when I have time.  Should I close this PR?

For clarification, this shortcoming of `xmp_set_row` only surfaces when called on the final ordinal (pattern reference) in a song order.  For any other ordinal, `end_point` is set automatically when playback reaches row 0 of the next ordinal.

AliceLR: Since sezero tagged @cmatsuoka for feedback I think this should probably be left open, but I think you're correct that the current approach may need to be revised. Manually changing the order and row breaks the assumptions the scan made when determining the end point, though, so maybe it's not that big of a problem.

cmatsuoka: Agree with @NoSuck and @AliceLR, I think we should fix it in a different way that works in all cases.

That said, I also agree that implementation currently in libxmp is super confusing and should at some point be refactored to something that's easier to follow (eliminating the scan code if possible). Establishing a few rules of what we actually want to happen in different scenarios can help us to redesign this code and test it.

sezero: > Agree with @NoSuck and @AliceLR, I think we should fix it in a different way that works in all cases.

So, this should wait until after 4.5.0 release?

NoSuck: > ... (eliminating the scan code if possible).

Now *that* would be something.  I won't hold my breath, though. 😅

I will close this PR and instead open an issue that references it, until we figure out a strategy.

PR #307, opened 2021-01-31
Convert the docs/Changelog into UTF-8

Wohlstand:
Using the ISO-8859-15 causes problem in several text editors and code viewers because of clunky nature of ANSI charsets system which are depending on computer locales


sezero (approved):
This is OK.

PR #308, opened 2021-01-31
Ignore files and directories generated by the lite build

Wohlstand:
When do build the lite version, there is a lot of junk getting appears in the git status. Maybe do ignore that?


sezero (approved):
This looks OK.  Will merge as soon as Travis completes.

PR #309, opened 2021-02-01
rewrite umx loader to use the embedded offset, size and object type

sezero:
... instead of linear search.

Similar versions are in libmikmod and libmodplug. For the
upkg parser code, see:  https://github.com/sezero/umr.git

### DIFF
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -23,6 +23,11 @@
 #include <stddef.h> /* offsetof() */
 #include "loader.h"
 
+#ifdef _WIN32
+#undef strcasecmp
+#define strcasecmp _stricmp
+#endif
+
 extern const struct format_loader libxmp_loader_xm;
 extern const struct format_loader libxmp_loader_it;
 extern const struct format_loader libxmp_loader_s3m;
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -291,7 +291,7 @@ static int32 probe_header (void *header)
 	/* byte swap the header - all members are 32 bit LE values */
 	p = (unsigned char *) header;
 	swp = (uint32 *) header;
-	for (i = 0; i < (int)sizeof(struct upkg_hdr)/4; i++, p += 4) {
+	for (i = 0; i < UPKG_HDR_SIZE/4; i++, p += 4) {
 		swp[i] = p[0] | (p[1] << 8) | (p[2] << 16) | (p[3] << 24);
 	}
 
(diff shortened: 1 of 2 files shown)


src/loaders/umx_load.c
```diff
@@ -43,97 +42,374 @@ const struct format_loader libxmp_loader_umx = {
...
+	"IT", "S3M", "XM", "MOD",
+	"WAV", "MP2", NULL
+};
+
+/* decode an FCompactIndex.
+ * original documentation by Tim Sweeney was at
+ * http://unreal.epicgames.com/Packages.htm
+ * also see Unreal Wiki:
+ * http://wiki.beyondunreal.com/Legacy:Package_File_Format/Data_Details
+ */
```
AliceLR, line 100: Not a real review yet (sorry, it is late, will do tomorrow) but I checked docs/formats/ and currently there is no UMX doc. Both of these links seem dead, but would it be worth pulling one of these from Wayback (or somewhere else)?

sezero: Wayback links: 
https://web.archive.org/web/20080122034944/http://unreal.epicgames.com/Packages.htm
https://web.archive.org/web/20201023045227/http://wiki.beyondunreal.com/Legacy:Package_File_Format
https://web.archive.org/web/20201023062426/http://wiki.beyondunreal.com/Legacy:Package_File_Format/Data_Details

Don't know whether it's worth adding something under docs/

sezero: OK, I copied the attached txt from beyondunreal wiki. 
Would it be appropriate to add under docs/?
[upkg.txt](https://github.com/libxmp/libxmp/files/5911580/upkg.txt)

sezero: Well, I went ahead and added the documentation under docs/formats/ with the last commit.

src/loaders/umx_load.c
```diff
@@ -43,97 +42,374 @@ const struct format_loader libxmp_loader_umx = {
...
-	uint8 buf[TEST_SIZE], *b = buf;
-	uint32 id;
 
-	LOAD_INIT();
+	/* byte swap the header - all members are 32 bit LE values */
+	p = (unsigned char *) header;
+	swp = (uint32 *) header;
+	for (i = 0; i < UPKG_HDR_SIZE/4; i++, p += 4) {
+		swp[i] = p[0] | (p[1] << 8) | (p[2] << 16) | (p[3] << 24);
+	}
```
AliceLR, line 301: This seems questionable—`header` here is a stack allocated char array passed from `process_upkg` and I suspect it is not guaranteed to be DWORD aligned. Would it be less confusing (and safer) to just read these values directly into a `upkg_hdr` struct with `hio_read32l`?

AliceLR:
I set up Unreal Gold and Unreal Tournament to test their music with this and predictably everything still works. Aside from the questionable byteswapping in `probe_header` I think this patch looks OK.

src/loaders/umx_load.c
sezero: Done with the new commit: please review.

AliceLR (approved):
This looks good to me.

sezero:
> This looks good to me.

Thanks. Will merge as soon as Travis completes.

PR #310, opened 2021-02-02
Makefile.vc (CFLAGS): define _CRT_SECURE_NO_WARNINGS to silence compiler

sezero:


AliceLR (approved):
This is OK as far as I am concerned. I don't have MSVC set up to test this but these warnings sound like they'd be annoying for libxmp. Most potentially unsafe strncpy, etc. usages have already been patched with the help of GCC's more context-aware warnings to work around the inherent security issues of those functions.

PR #311, opened 2021-02-02
Fix misc. crashes and hangs found by AFL++ fuzzing.

### LINKED ISSUE #301: DoS (and smaller leaks) revealed by AFL++, batch 2

### LINKED ISSUE #303: Memory errors revealed by AFL++, batch 9
The fuzzer has produced a new stack-based OOB write, in yet another ProWizard loader:
[xmp_remaining_crashes_20210120_1048.tar.gz](https://github.com/libxmp/libxmp/files/5841605/xmp_remaining_crashes_20210120_1048.tar.gz)
```
Loading output/4/crashes/id:000023,sig:06,src:012683+009969,time:307286193,op:splice,rep:8 (1 of 1)
==2167106==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffd7bd4de80 at pc 0x7fc919fa8cf7 bp 0x7ffd7bd2d8f0 sp 0x7ffd7bd2d8e8
WRITE of size 1 at 0x7ffd7bd4de80 thread T0
    #0 0x7fc919fa8cf6 in depack_p4x .../xmp/libxmp/src/loaders/prowizard/p40.c:247:17
    #1 0x7fc919f78b23 in pw_wizardry .../xmp/libxmp/src/loaders/prowizard/prowiz.c:112:6
    #2 0x7fc919f64d89 in pw_load .../xmp/libxmp/src/loaders/pw_load.c:81:6
    #3 0x7fc919eb1890 in load_module .../xmp/libxmp/src/load.c:461:18
    #4 0x7fc919eb0bfc in xmp_load_module .../xmp/libxmp/src/load.c:613:8
    #5 0x4cee43 in main .../xmp/xmp-cli/src/main.c:399:9
    #6 0x7fc9199c6d09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #7 0x420679 in _start (.../bin/xmp+0x420679)

Address 0x7ffd7bd4de80 is located in stack of thread T0 at offset 132256 in frame
    #0 0x7fc919fa6eaf in depack_p4x .../xmp/libxmp/src/loaders/prowizard/p40.c:28

  This frame has 5 object(s):
    [32, 1056) 'tmp' (line 30)
    [1184, 132256) 'tr' (line 33) <== Memory access at offset 132256 overflows this variable
    [132512, 133536) 'track_addr' (line 34)
    [133664, 133788) 'SampleAddress' (line 37)
    [133824, 133948) 'SampleSize' (line 38)
HINT: this may be a false positive if your program uses some custom stack unwind mechanism, swapcontext or vfork
      (longjmp and C++ exceptions *are* supported)
SUMMARY: AddressSanitizer: stack-buffer-overflow .../xmp/libxmp/src/loaders/prowizard/p40.c:247:17 in depack_p4x
```

AliceLR:
Fixes #301. Fixes #303. Includes fixes for the following bugs (despite the branch name, most are not ProWizard):

* Off-by-ones in The Player 5.x/6.0 and 6.1a test functions would allow out of bounds reads by indexing a non-existent sample (https://github.com/libxmp/libxmp/issues/303#issuecomment-764513079).
* Fixed missing row number sanity checks (leading to stack corruption) and a related large stack allocation in The Player 4.x loader (https://github.com/libxmp/libxmp/issues/303#issue-789844937).
* Various DigiBooster Pro chunk handlers could be run before the `INFO` chunk had been loaded, allowing invalid frees in `xmp_release_module` (https://github.com/libxmp/libxmp/issues/303#issuecomment-770028071).
  * I also screwed up an earlier sanity check I added here by mistaking a row count for a track count. This has been corrected and I added another valid test file for this format.
* Duplicate `PATT` chunks in the Digital Tracker loader could corrupt the pattern/track counts, causing invalid frees in `xmp_release_module` (https://github.com/libxmp/libxmp/issues/303#issuecomment-768080146).

The following fixes are for the second set of slow loading files from https://github.com/libxmp/libxmp/issues/301#issuecomment-768163976:

* The Digital Tracker loader was also missing a track count sanity check, resulting in hangs and high RAM usage.
* The MDL sample loader was missing sanity checks on sample sizes before allocating sample buffers, resulting in hangs and high RAM usage.
* The Impulse Tracker compressed sample loader was missing sanity checks on sample sizes before allocating sample buffers, resulting in hangs and high RAM usage. I also fixed a couple of other minor bugs here:
  * A debug comment was referencing the wrong name field, causing crashes with `-DDEBUG` builds.
  * `hio_error` state carried over from truncated samples in the sample loader could cause loading to fail after seeking to pattern data. This affected something like a couple dozen files from the ModLand collection (I didn't bother picking a test file).
  * Maximum sample size sanity check should only be applied when the sample flag is set. This affected one IT module that I found.

The following fix is tentative and should be looked at by someone with more experience with the depackers than me (@sezero?):

* The LHA depacker could encounter characters with invalid code lengths of 0 bits, leading to potentially very slow loads and large temporary file output. I'm not 100% sure the fix here is correct since LHA documentation is sparse. This might really be a bug in the tree table generation code.


sezero:
>     * The Impulse Tracker [...] I also fixed a couple of other minor bugs here:
>       * `hio_error` state carried over from truncated samples in the sample loader could cause loading to fail after seeking to pattern data.

I tested this part by removing the hio_error() calls resetting the error flag in my umx patch (#309), and umx-contained IT files did load, however S3M and XM ones did not: they possibly have the same issue.

>  following fix is tentative and should be looked at by someone with more experience with the depackers than me
>     * The LHA depacker could encounter characters with invalid code lengths of 0 bits [...]

Don't know much about it myself. Maybe looking at https://github.com/jca02266/lha.git
may help (huf.c and others there have CVE notes)

sezero (approved):
Looks good at first glance.  Do you want more eyes?

AliceLR:
> Don't know much about it myself. Maybe looking at https://github.com/jca02266/lha.git
> may help (huf.c and others there have CVE notes)

Thanks for the link, I will check it out.

> Looks good at first glance. Do you want more eyes?

That might be a good idea.

sezero:
> > Looks good at first glance. Do you want more eyes?
> 
> That might be a good idea.

Requested review from @cmatsuoka

cmatsuoka (approved):
Thanks Alice for fixing the fuzzing errors, looks good to me.

Some random comments about the formats and plans I had for future versions of libxmp:

- Many of the obscure formats like DT are really experimental and the code quallity is quite low. I wonder if we should group them and build only if configured with `--enable-experimental-formats` or something like that is specified -- I'm pretty sure they're not very accurate and they increase the fuzz attack surface a lot.

- MDL replays reasonably well but has a number of problems too, some of them requiring changes deep in the player core to be handled correctly. If some day better MDL support is implemented, we could use less guesswork and check the sources (Jörn was able to recover the tracker sources originally thought to be lost in a disk crash).

- One thing I'd like to do with Asle's Prowizard is to spin it off as a separate lilbrary that scans and converts entirely in memory so it could be used by other projects as well, with a better architecture and reconciling it with the original sources, adding more formats, etc. This would work well with the idea to drop disk i/o from libxmp, which would make it work better in platforms without stdio.

AliceLR:
Sorry for taking so long getting around to replying to this:

> * Many of the obscure formats like DT are really experimental and the code quallity is quite low. I wonder if we should group them and build only if configured with `--enable-experimental-formats` or something like that is specified -- I'm pretty sure they're not very accurate and they increase the fuzz attack surface a lot.

Being able to opt out of some of the more obscure formats seems like a good idea. That might also provide a staging area for the Graoumf Tracker and X-Tracker loaders to be reintroduced and fixed. Do you have a full list of formats that should be considered for that group?

> * MDL replays reasonably well but has a number of problems too, some of them requiring changes deep in the player core to be handled correctly. If some day better MDL support is implemented, we could use less guesswork and check the sources (Jörn was able to recover the tracker sources originally thought to be lost in a disk crash).

What exactly are the problems preventing MDL modules from playing accurately right now? It looks like it has multiple effects channels like DBM/GDM/ULT, but I'm not really familiar with this format.

> * One thing I'd like to do with Asle's Prowizard is to spin it off as a separate lilbrary that scans and converts entirely in memory so it could be used by other projects as well, with a better architecture and reconciling it with the original sources, adding more formats, etc. This would work well with the idea to drop disk i/o from libxmp, which would make it work better in platforms without stdio.

This seems reasonable—ProWizard's approach of identifying+converting formats seems at odds with how libxmp currently tests formats (e.g. how libxmp has to run the ProWizard format tests twice, once in the test function and again when loading).

PR #315, opened 2021-02-18
bug #312 fixes

sezero:
Small set of patches fixing _some_ of the memory leaks / integer overflows
reported at https://github.com/libxmp/libxmp/issues/312

### DIFF
--- a/src/depackers/arcfs.c
+++ b/src/depackers/arcfs.c
@@ -27,10 +27,10 @@ struct archived_file_header_tag {
 	unsigned char method;
 	unsigned char bits;
 	char name[13];
-	unsigned long compressed_size;
+	unsigned int compressed_size;
 	unsigned int date, time, crc;
-	unsigned long orig_size;
-	unsigned long offset;
+	unsigned int orig_size;
+	unsigned int offset;
 };
 
 
@@ -110,7 +110,7 @@ static unsigned char *read_file_data(FILE *in,
 	unsigned char *data;
 	int siz = hdrp->compressed_size;
 
-	if (!siz || (data = malloc(siz)) == NULL) {
+	if (siz <= 0 || (data = malloc(siz)) == NULL) {
 		goto err;
 	}
 	if (fseek(in, hdrp->offset, SEEK_SET) < 0) {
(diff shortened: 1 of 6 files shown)


src/depackers/arcfs.c
```diff
@@ -110,7 +110,7 @@ static unsigned char *read_file_data(FILE *in,
 	unsigned char *data;
 	int siz = hdrp->compressed_size;
 
-	if (!siz || (data = malloc(siz)) == NULL) {
+	if (siz <= 0 || (data = malloc(siz)) == NULL) {
```
AliceLR, line 113: Considering this field is unsigned in the header tag struct could this just be made into a `size_t`?

src/depackers/arcfs.c
```diff
@@ -27,10 +27,10 @@ struct archived_file_header_tag {
 	unsigned char method;
 	unsigned char bits;
 	char name[13];
-	unsigned long compressed_size;
+	unsigned int compressed_size;
 	unsigned int date, time, crc;
-	unsigned long orig_size;
-	unsigned long offset;
+	unsigned int orig_size;
+	unsigned int offset;
```
AliceLR, line 33: Considering some of these fields were `long` the `int`s are suspect too—are those supposed to be 16-bit DOS ints? I think using `fread` hacks to fill in struct data like this should be avoided for exactly this sort of issue. Like you mentioned, stdint.h or compatible types could also be used instead... probably better to just get rid of the offending `fread` usage though. Other bad things here: there's an implicit padding byte after `name` that is hidden by this hack.

AliceLR (changes requested):
Most patches look good to me. The arcfs.c patches still look questionable (that isn't your fault though).

AliceLR (approved):
Since one of the things I requested changes doesn't really matter and the other was a mistake on my part I'm going to go ahead and just approve this.

src/depackers/arcfs.c
sezero: Nope, unnecessary. If we have to, we can change it to uint32 and check the `value > INT_MAX`
or something.

src/depackers/arcfs.c
sezero: The `int`s are most probably amiga ints and 32 bits.

sezero:
OK, merging.

PR #320, opened 2021-03-06
Fix uninitialized reads and various other bugs found by AFL++.

AliceLR:
Fixes a handful of crash/leak bugs and a lot of uninitialized reads found in #319 (but not quite all of them).

* Duplicate SA chunks in the MDL loader could cause leaks. The duplicate/misordered chunk checks also needed to be cleaned up.
* The OKT loader could cause crashes when a CMOD chunk was encountered out of order.
* The Funktracker loader could attempt to allocate and load patterns with 0 channels, leading to heap corruption.
* The Digital Tracker loader could leak subinstruments.
* Various loaders could leak samples by loading over `MAX_SAMPLES` samples. This number is arbitrary so I limited checking this to one place in the shared sample allocation code.
* The Novotrade unpacker would allow garbage in the order list to get passed through to the common ProWizard loader, causing problems.
* Something like 20 other uninitialized memory read fixes that I'm not going to list here. Most of them were missing `hio_read` return value checks. See these comments for a summary:
  * Initial triage by Lionel Debroux: https://github.com/libxmp/libxmp/issues/319#issuecomment-787035390
  * All of the other UMRs detected from MSan by this batch: https://github.com/libxmp/libxmp/issues/319#issuecomment-791894403

This is missing fixes for a few more #319 UMRs found by valgrind but not by MSan, but this branch is already big enough as it is.


sezero (approved):
This looks OK. Will merge as soon as travis completes.

sezero:
About bzip2, though: bzip2 is an actively maintained software. How far is libxmp behind mainstream? Is mainstream version not compatible with what we have?

AliceLR:
> About bzip2, though: bzip2 is an actively maintained software. How far is libxmp behind mainstream? Is mainstream version not compatible with what we have?

I'm not sure. It seems like generally the depackers are all contributed code so libxmp doesn't need to pull in any external dependencies (which isn't necessarily a bad thing and e.g. makes it easier to MSan). I have a feeling some of them have more up-to-date equivalents with more permissive licenses. (I think in that case readlzw.c and readrle.c should be higher priority for replacement though since the Digital Symphony loader pulls them in even when depackers are turned off.)

PR #322, opened 2021-03-07
Fix uninitialized reads and high RAM usage found by AFL++ fuzzing, part 2.

AliceLR:
This cleans up the remaining junk leftover from the first few fuzzing batches in #319 and also fixes a high memory consumption bug.

* The MMD2/3 loader had the same iinfo instrument name uninitialized read bug that the MMD0/1 loader had.
* The MMDInfo loader for MMD2/3 modules could still generate very large allocations, so this is now bounded to `hio_size`.
* The Player 5.x/6.0a and The Player 6.1a modules could reference uninitialized sample addresses. It doesn't seem like this should be allowed but for now I just made these zero initialize the sample address array.


sezero (approved):
OK

PR #329, opened 2021-03-23
mkstemp: don't use pid_t for windows builds.

sezero:
for whatever reason, mingw-w64 defines pid_t as __int64 for WIN64,
but getpid() always returns int.

### DIFF
--- a/src/mkstemp.c
+++ b/src/mkstemp.c
@@ -46,7 +46,6 @@
 #include <process.h>
 #define getpid _getpid
 #define open _open
-typedef int pid_t;
 #endif
 
 #ifndef O_BINARY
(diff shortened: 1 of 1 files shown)


AliceLR (approved):
This looks good to me. (Wonder why the MinGW64 headers use __int64... since it's redefined in several places including pthread.h, maybe it's a holdover from the Linux headers?)

sezero:
>  (Wonder why the MinGW64 headers use __int64...

Finally took time and found the old discussion at mingw-w64-public mailing list:
https://sourceforge.net/p/mingw-w64/mailman/mingw-w64-public/thread/1456671365-21759-1-git-send-email-sw%40weilnetz.de/#msg34890417

Issue #222, opened 2020-11-12
Build regression tests with Visual Studio in Windows


cmatsuoka:
We should provide means to run regression tests for our WIndows build in Travis.

AliceLR:
I added a Makefile.vc.in file in #331 and got the tests running with nmake so that's at least some progress here. There are a couple of issues: first, all_tests.c still requires GNU Make and the resulting Makefile.vc is a little verbose, so they still require some support from outside of MSVC (shouldn't be an issue for Travis-CI). Second, some of the tests were failing/crashing for me, and (assuming these can be reproduced elsewhere and aren't due to user error) should be resolved before adding Travis-CI tests:
```
test 185: test_api_test_module: at .\test_api_test_module.c:16: errno test module fail: **fail**
test 186: test_api_load_module: at .\test_api_load_module.c:20: errno code: **fail**
test 188: test_api_load_module_from_file:
test 189: test_api_test_module_from_file:
test 206: test_api_smix_load_sample: at .\test_api_smix_load_sample.c:32: errno: **fail**
```

sezero:
> ```
> test 185: test_api_test_module: at .\test_api_test_module.c:16: errno test module fail: **fail**
> ```

This one is because errno not matching EISDIR: Because fopen()ing a
directory would actually succeed, xmp_load_module and xmp_test_module
manually set errno to EISDIR but not for MSVC, hence the test failure.
I suggest something like the following:
```diff
diff --git a/src/load.c b/src/load.c
index 9052849..ccc592a 100644
--- a/src/load.c
+++ b/src/load.c
@@ -336,6 +336,10 @@ static int test_module(struct xmp_test_info *info, HIO_HANDLE *h)
        return -XMP_ERROR_FORMAT;
 }
 
+#if defined(_MSC_VER)
+#define S_ISDIR(x) (((x)&_S_IFDIR) != 0)
+#endif
+
 int xmp_test_module(const char *path, struct xmp_test_info *info)
 {
        HIO_HANDLE *h;
@@ -348,12 +352,10 @@ int xmp_test_module(const char *path, struct xmp_test_info *info)
        if (stat(path, &st) < 0)
                return -XMP_ERROR_SYSTEM;
 
-#ifndef _MSC_VER
        if (S_ISDIR(st.st_mode)) {
                errno = EISDIR;
                return -XMP_ERROR_SYSTEM;
        }
-#endif
 
        if ((h = hio_open(path, "rb")) == NULL)
[...]

PR #331, opened 2021-03-24
Use _spawnl() for unit tests on Windows, add unit test Makefile.vc.in.

AliceLR:
This reduces the time it takes to run libxmp-tests.exe in my local MSYS2 setup from ~15 seconds to about 7 seconds. Considering how much I have been running the tests locally, this is a huge improvement. Presumably this will also be useful for adapting the tests to Visual Studio (#222). Thanks to Lionel Debroux for bringing the _spawn* family of functions to my attention. :)

(Suggestions for improving this are definitely welcome. The _spawn* functions apparently have existed since ~1993 and should be fairly portable.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,10 @@ test-dev/libxmp-covertest
 test-dev/libxmp-test
 
 # VC++ files
+/.vs
+/.vscode
+*.exp
+*.lib
 *.suo
 *.ncb
 
--- a/test-dev/Makefile.in
+++ b/test-dev/Makefile.in
@@ -596,6 +596,11 @@ dist-test:
 clean:
 	rm -f *.o core *~ $(T_OBJS)
 
+vc-prepare:
+	@echo Generate Makefile.vc
+	@sed -e 's!@SRCS@!$(subst /,\\,$(T_OBJS:.o=.c))!' \
+	     Makefile.vc.in > Makefile.vc
+
 #
 # Utilities
 #
--- a/test-dev/Makefile.vc.in
+++ b/test-dev/Makefile.vc.in
@@ -1,0 +1,26 @@
+CC	= cl
+CFLAGS	= /O2 /W3 /I..\include /I..\src /I..\src\win32 /DWIN32 \
+	  /Dinline=__inline /DPATH_MAX=1024 /D_USE_MATH_DEFINES /D_CRT_SECURE_NO_WARNINGS
+LDFLAGS = /RELEASE /OUT:$(EXE)
+EXE	= libxmp-tests.exe
+
+SOURCES	= @SRCS@
+ALL_SOURCES	= $(SOURCES)
+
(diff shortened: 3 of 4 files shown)


AliceLR:
Please hold off on merging this, there are some other issues in main.c that seem to need to be fixed to get this to build with MSVC.

AliceLR:
Since I had to get it working to test these changes with MSVC I ended up just writing a Makefile.vc.in for near future use with Travis-CI. Due to the massive list of source files inserted into Makefile.vc and not really having a good way to generate `all_tests.c` right now the generated Makefile.vc should probably just not be committed (I believe the Travis-CI environment either has `make` or can easily install it, so that won't be an issue there).

sezero:
Is this ready for applying now?

AliceLR:
I think so. I checked over everything again and found a mistake in the nmake clean rule, but everything else looks fine.

sezero:
> ```
> test 185: test_api_test_module: at .\test_api_test_module.c:16: errno test module fail: **fail**
> ```

Looked briefly at vs2008 crt source, nothing maps to EISDIR in _dosmaperr() as far as I can see.

sezero:
> ```
> test 186: test_api_load_module: at .\test_api_load_module.c:20: errno code: **fail**
> test 206: test_api_smix_load_sample: at .\test_api_smix_load_sample.c:32: errno: **fail**
> ```

Huh.. Why is errno not ENOENT..

AliceLR:
I added debug lines and in both of those situations errno is 0. Re: the two without messages, usually I only see that happen with crashes (since the test prints "fail" in Windows right now...). The load test without a message also hangs for quite a bit before continuing on to the next test.

sezero:
I wonder, have you run test-dev built with mingw and ever encountered
these failures, or are these really msvc-specific and not always-win32?

AliceLR:
These tests all work as expected with mingw. I've only seen these new errors specifically with the new MSVC Makefile in this pull request.

sezero:
> These tests all work as expected with mingw.

And _not_ cygwin?

AliceLR:
> And _not_ cygwin?

Correct—to clarify, all of my MSYS2 development is done in the MINGW64 environment, which specifically uses MinGW and generates native executables linked to MSVCRT.DLL (for instance I can and have run my `libxmp-tests.exe` from Windows without any involvement from MSYS2 and it still works). MinGW might add a compatibility layer for fopen/errno or whatever this issue is might not affect the MSVCRT.DLL runtime.

sezero:
Then I'm out of ideas.  Maybe The system msvcrt.dll is OK, but msvcrXX runtime dlls are broken?? Whatever.

AliceLR:
FWIW since the MSVC tests aren't being automated yet, further diagnosing of these issues might be better suited to #222.

sezero (approved):

PR #333, opened 2021-03-25
Update .travis.yml to add unit tests to MSVC checks.

### LINKED ISSUE #222: Build regression tests with Visual Studio in Windows

AliceLR:
Updates .travis.yml to add MSVC regression tests. This required minor tweaks to how the *nix regression tests Makefile operates, but nothing too severe. The files `Makefile.vc` and `all_tests.txt` need to be generated by the *nix Makefile for nmake to be able to run the tests. `Makefile.vc` is updated with `make vc-prepare` and only needs to be changed if a new non-test object or libxmp internal object needs to be added. `all_tests.txt` is required for the MSVC Makefile to generate its sources list and `all_tests.c`, and is automatically updated when the regression tests are run with the *nix Makefile.

Fixes #222.


sezero:
`The command "cmd.exe //C 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Auxiliary\Build\vcvarsall.bat' amd64 '&&' cd test-dev '&&' copy Makefile.in Makefile '&&' mingw32-make vc-prepare all_tests.c '&&' nmake -f Makefile.vc" exited with 2.`

I _think_ the only way to do this might be pregenerating test-dev/Makefile.vc

AliceLR:
Yeah, you might be right. Not entirely sure what mingw32-make doesn't like about vc-prepare... oh well.

AliceLR:
Trying again with mingw32-make only used for all_tests.c. If that doesn't work, all_tests.c might need to be pregenerated or the current way the tests are listed in the Makefile might need to be reconsidered...

sezero:
May it be that you need to prepend `./` to `all_tests.c`?

AliceLR:
After some experimentation with a copy of mingw32-make I have locally it seems that the expansions of those lines might just be too large for sh.

sezero:
Pregenerating it on unix, then.

AliceLR:
I think that would work but my concern is every time a test is added, on top of the Makefile.in there are now two generated files that need to be committed. There's also the looming problem of some terminals having relatively low ARG_MAX values (MSYS2 is 32000, the test list is currently around 20k), but that can probably be dealt with later.

My thought right now is to only pregenerate a single list of test names for MSVC plus have an MSVC Makefile that creates its own sources list and all_tests.c. I have Makefile.vc generating all_tests.c currently with the following hacks:

Unix Makefile all_tests.txt pregeneration code:
```Makefile
all_tests.txt: Makefile
        $(file >$@) $(foreach i,$(TEST_NAMES),$(file >>$@,$i))
        @true
```

Makefile.vc
```Makefile
all_tests.c: all_tests.txt
        type nul > $@
        for /F "tokens=*" %%I in ( $** ) do echo declare_test(%%I); >> $@
```

The sources list could probably be a similar job of generating a Makefile to include.

AliceLR:
It looks like this new method of handling the tests lists works even if it's kind of ridiculous. Tests ran and completed for both MSVC checks with one fail each (`test_api_load_module` crashing again, that should have been fixed by the last PR...).

sezero:
I think this is needed?
```diff
diff --git a/test-dev/test_api_load_module.c b/test-dev/test_api_load_module.c
index 4daddb0..2eb68c2 100644
--- a/test-dev/test_api_load_module.c
+++ b/test-dev/test_api_load_module.c
@@ -54,7 +54,7 @@ TEST(test_api_load_module)
        fail_unless(state == XMP_STATE_UNLOADED, "state error");
 
        /* invalid format */
-       ret = xmp_load_module(ctx, "Makefile");
+       ret = xmp_load_module(ctx, "Makefile.in");
        fail_unless(ret == -XMP_ERROR_FORMAT, "invalid format");
 
        state = xmp_get_player(ctx, XMP_PLAYER_STATE);
```

sezero:
OK, I was too late :)

sezero:
On mysetup,  I cannot generate all_tests.txt. Make version is 3.81.

sezero:
> On mysetup, I cannot generate all_tests.txt. Make version is 3.81.

The following works with make-3.81, though:
```
$(TEST_PATH)/all_tests.txt: Makefile
	@$(RM) $@; \
	for i in $(TEST_NAMES); do \
		echo "$$i" >> $@; \
	done
```

AliceLR:
Alright, since $(file) seems to not be very portable I replaced it with something more like that.

sezero (approved):

Issue #325, opened 2021-03-16
xmp_load_module_from_memory disables bounds checks when size <= 0.


AliceLR:
> the size of the module, or 0 if the size is unknown or not specified. If size is set to 0 certain module formats won't be recognized, the MD5 digest will not be set, and module-specific quirks won't be applied.

https://github.com/libxmp/libxmp/blob/master/docs/libxmp.rst#int-xmp-load-module-from-memory-xmp-context-c-const-void-mem-long-size

```C
	/* Use size < 0 for unknown/undetermined size */
	if (size == 0)
		size--;
```
https://github.com/libxmp/libxmp/blob/master/src/load.c#L632

What exactly is the use case for this behavior? When will the user *not* know the size of the memory buffer that is provided? Should this really be encouraged? This seems comically unsafe and is the cause of a crash found in #302.

sezero:
I truly don't like it, it really is seeking for trouble.

@cmatsuoka?

AliceLR:
(My opinion for whatever it's worth: this should be removed ASAP, the field should be made an `unsigned long` for 4.5.0 and should be extended to `size_t` for 5.0.)

sezero:
> (My opinion for whatever it's worth: this should be removed ASAP,

Seconded, although it is a behavior change, alhough+ I don't know whether anyone relies on that..

> the field should _maybe_ be made an `unsigned long` for 4.5.0, and should be extended to `size_t` for 5.0.)

Which field, the `long` size one? I think it should stay signed so we can easily check for bad args :)
For 5.0, `ptrdiff_t` or `ssize_t` maybe?

AliceLR:
> Which field, the `long` size one? I think it should stay signed so we can easily check for bad args :)
> For 5.0, `ptrdiff_t` or `ssize_t` maybe?

Yes, I think you're correct, which is why I added "maybe": if anyone is actually using a negative value it'd be good to catch it, so leaving it for now and a `ptrdiff_t` for 5.0 would be better.

cmatsuoka:
Remove it. I agree that if we load from a memory buffer we should know what data we're dealing with. It was probably added a long time ago without considering any robustness implication.

sezero:
> Remove it. I agree that if we load from a memory buffer we should know what data we're dealing with. It was probably added a long time ago without considering any robustness implication.

I'll be working on a patch shortly.

PR #337, opened 2021-03-29
remove support for size <= 0:

sezero:
- return -XMP_ERROR_INVALID from xmp_load_module_from_memory() and
  xmp_test_module_from_memory() if size <= 0, remove all mention of
  size <= 0 from their documentation.
- remove non-positive sizes support from memio.c. whitespace tidy-ups.
- remove hio_size() <= 0 check from load.c::set_md5sum()
- remove file_len >= 0 checks from sample.c::libxmp_load_sample()
- document that the size parameter to xmp_load_module_from_file() is
  actually ignored. (don't know whether it had ever mattered..)
- change test_read_mem_hio_nosize so its tests always fail

Fixes: https://github.com/libxmp/libxmp/issues/325

Notes: hio.c::hio_open_file() doesn't fail if get_size() returns
negative: needs further check / handling.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -312,9 +312,7 @@ int xmp_test_module_from_memory(const void \*mem, long size, struct xmp_test_inf
     :mem: a pointer to the module file image in memory. Multi-file modules
       or compressed modules can't be tested in memory.
 
-    :size: the size of the module, or 0 if the size is unknown or not
-      specified. If size is set to 0 certain module formats won't be
-      recognized.
+    :size: the size of the module.
 
     :test_info: NULL, or a pointer to a structure used to retrieve the
       module title and format if the memory buffer is a valid module.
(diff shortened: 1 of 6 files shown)


AliceLR (approved):
Removing the `size >= 0` checks from places not related specifically to memory files might cause some breakage when `xmp_load_module_from_file` is provided with e.g. a pipe but I can't imagine that ever worked reliably to begin with, so IMO no big loss. This looks good to me.

sezero:
> Removing the `size >= 0` checks from places not related specifically to memory files might cause some breakage when `xmp_load_module_from_file` is provided with e.g. a pipe but I can't imagine that ever worked reliably to begin with, so IMO no big loss. This looks good to me.
> 
> edit: `test_read_mem_hio_nosize` should probably be altered to reflect the new behavior though.

Thanks, applied to master.

PR #340, opened 2021-03-31
Add codecov configuration

cmatsuoka:
Add codecov status configuration. We may need to adjust this to
obtain the results we want.


AliceLR (approved):
I'm not personally too familiar with Codecov but this configuration looks sensible. (Going off of the reference [here](https://docs.codecov.io/docs/codecovyml-reference).)

Issue #335, opened 2021-03-27
add libxmp_syserrno()?


sezero:
As noticed in https://github.com/libxmp/libxmp/issues/222, statically linked dlls have issues sharing errno.
Should we add something like libxmp_syserrno(), returning simply errno?

sezero:
Here is a draft: https://github.com/libxmp/libxmp/compare/master...sezero:errno

AliceLR:
I think having something like this and encouraging its use over plain `errno` seems like a good idea. Some thoughts though:

1) In environments where libxmp is dynamically linked this can return an `errno` value from something unrelated. Should this instead be an `xmp_context` value specific to (and only set by) libxmp?
2) Should there be a way to clear it built-in?

sezero:
> 1. In environments where libxmp is dynamically linked this can return an
>   `errno` value from something unrelated. Should this instead be an
>   `xmp_context` value specific to (and only set by) libxmp?

xmp_syserrno() should only be called upon receiving -XMP_ERROR_SYSTEM.
With that in mind, current tree can always return an unrelated errno, too. As for
tying it to xmp_context: xmp_context creation can also fail with -XMP_ERROR_SYSTEM,
so I'm not sure how we can handle this.

(Do you have any proof-of-concept code?)

>     2. Should there be a way to clear it built-in?

I'm most certainly against that.

sezero:
@cmatsuoka: What do you think?

sezero:
I'm making this into a PR, let's improve upon that.

PR #342, opened 2021-03-31
added xmp_syserrno() to libxmp api.

sezero:
To be used to retrieve errno upon receiving -XMP_ERROR_SYSTEM from
a libxmp function call. Useful when either libxmp or its client is
statically linked to libc.

Closes: https://github.com/libxmp/libxmp/issues/335

@AliceLR reported concerns, quoting from #335:
> In environments where libxmp is dynamically linked this can return an
> `errno` value from something unrelated. Should this instead be an
> `xmp_context` value specific to (and only set by) libxmp?

My anwser was: 
xmp_syserrno() should only be called upon receiving `-XMP_ERROR_SYSTEM`.
With that in mind, current tree can always return an unrelated `errno`,
too. As for tying it to xmp_context: xmp_context creation can also fail
with `-XMP_ERROR_SYSTEM`, so I'm not sure how we can handle this.

> Should there be a way to clear it built-in?

Me: I'm most certainly against that.


sezero:
@cmatsuoka, and @AliceLR: please review / improve

AliceLR (approved):
This looks good to me.

> My anwser was:

To be clear, I think both of those replies are fair. Those were intended more as questions than concerns. I don't really have any recommendations for improving this: it does something useful, its usage makes sense, and it doesn't break anything...

sezero:
> This looks good to me.
> 
> > My anwser was:
> 
> To be clear, I think both of those replies are fair. Those were intended more as questions than concerns. I don't really have any recommendations for improving this: it does something useful, its usage makes sense, and it doesn't break anything...

OK then, mergin. Thanks!

PR #344, opened 2021-04-10
Use fork/exec and ptpopen.c instead of popen for external depackers.

### LINKED ISSUE #327: Shell command injection when calling external unmo3 or unrar
I noticed this while sifting through the output of valgrind as part of triaging the remaining issues from #323 .
At https://github.com/libxmp/libxmp/blob/master/src/load.c#L167 , libxmp delegates depacking for >= 100-byte files which start with "MO3" or "Rar" magic numbers to external `unmo3` and `unrar` tools, respectively. AFAICT, there's no format string injection in that path, as the filename is correctly pasted through `%s`... but there's a much bigger issue: `popen()` is being used, https://github.com/libxmp/libxmp/blob/master/src/load.c#L95 and below. Therefore, I looked for a command injection vulnerability based on the file name; indeed, there is command injection.

PoC:
```
$ echo -n 'MO34567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890' > '$(echo command injection > foo; LANG=C date >> foo; cat foo)'

$ LD_LIBRARY_PATH=~/xmp_prefix_nosan/lib ~/xmp_prefix_nosan/bin/xmp --load-only '$(echo command injection > foo; LANG=C date >> foo; cat foo)'; cat foo
Extended Module Player 4.1.0
Copyright (C) 1996-2016 Claudio Matsuoka and Hipolito Carraro Jr
Using null output
Mixer set to 44100 Hz, 16bit, cubic spline interpolated stereo
Press 'h' for help

Loading $(echo command injection > foo; LANG=C date >> foo; cat foo) (1 of 1)
Can not open STDOUT
.../xmp_prefix_nosan/bin/xmp: $(echo command injection > foo; LANG=C date >> foo; cat foo): Unrecognized file format
command injection
Sun Mar 21 23:26:59 CET 2021

$ echo -n 'Rar34567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890' > '$(echo command injection > foo; LANG=C date >> foo; cat foo)'

$ LD_LIBRARY_PATH=~/xmp_prefix_nosan/lib ~/xmp_prefix_nosan/bin/xmp --load-only '$(echo command injection > foo; LANG=C date >> foo; cat foo)'; cat foo
Extended Module Player 4.1.0
Copyright (C) 1996-2016 Claudio Matsuoka and Hipolito Carraro Jr
Using null output
Mixer set to 44100 Hz, 16bit, cubic spline interpolated stereo
Press 'h' for help

Loading $(echo command injection > foo; LANG=C date >> foo; cat foo) (1 of 1)
Can not open STDOUT
.../xmp_prefix_nosan/bin/xmp: $(echo command injection > foo; LANG=C date >> foo; cat foo): Unrecognized file format
command injection
Sun Mar 21 23:27:04 CET 2021
```

The `xmp` front-end is usually launched from a shell or a file explorer anyway (i.e. the user can normally already launch commands), but still - opening oddly named files containing the MO3 or RAR magic numbers (which might themselves be embedded into other archive formats, right ?) opens the door to arbitrary command execution with the privileges of the current user. I can't see a privilege escalation here on most modern computers which don't have setuid/setgid binaries left and right, and usually have proper ownership and mode on system binaries.
The call to `popen()` was already there in `src/misc/load.c` at the beginning of the Git history in June 2001, though at the time, and for a number of years after that, more external programs than just `unmo3` and `unrar` were executed.

The only correct way to launch external programs I'm aware of, is:
* not to depend on any form of external shell;
* build the argv array for the target program;
* set up file descriptors as needed to gather stdout (and stderr, if needed) from the target program;
* use functions from the `fork`+`exec` (*nix) or `spawn` (Windows) families.
Needless to say, this is inherently non-portable...

AliceLR:
Fixes #327. The short version is that, for unix systems, `popen` is vulnerable to shell injection, so those systems now use `fork`/`exec` to invoke external depackers. Windows builds now unconditionally use ptpopen.c as CreateProcess is known to not have this issue (vs. MinGW and Cygwin, which ¯\\\_(ツ)_/¯).

Known issues with this branch:
- [x] Some platforms may incorrectly detect `fork`/`exec` existing, causing the external depackers to fail (DJGPP) (them working prior to this also seems to have been questionable).
- [ ] The list of platforms that load.c currently allows to use `popen` is arbitrary and I don't know if any of them are actually safe.


sezero:
I'm not overly happy with this situation.

- For djgpp stuff:  I still think the patch from https://github.com/libxmp/libxmp/issues/327#issuecomment-808722682
  should be applied: popen() from djgpp works, it's unrar for dos which seems to screw-up

- The whole code is actually a mess: we're making it possibly messier.
  Do we truly need to support unmo'ing or unrar'ing archives? What if we
  drop the support for it?

@cmatsuoka, what do yo think?

AliceLR:
IMO if support for external depackers can be dropped altogether that would be the best solution, but for now I'm assuming that isn't reasonable. In my (limited, subjective) experience, outside of making the test files for this branch I haven't created or used a RAR archive since the mid-'00s for any purpose besides rearchiving it to ZIP or 7z. I'd never even heard of MO3 prior to working on this patch, but mixing non-free sample encodings into module formats sounds like a terrible idea that shouldn't be supported. Only 11 of these seem to exist on the Mod Archive and I didn't even see a directory for them on the ModLand ftp.

Re: DJGPP, I think that's fair, and I can modify this patch soon (instead of all of the extra DJGPP hacks in your patch it might be better to just keep detecting `popen` and choose `popen` before `fork` for the whitelisted platforms).

Re: this being a mess, yes, it is. As long as external depackers are supported I don't think it's a particularly avoidable mess. However, it can mostly be hidden in new compilation unit(s) instead of load.c, and I can go ahead and do that. There are a couple of solutions that might be worth considering: 1) `execute_command` can just be moved out, or 2) it might be worth throwing out the "internal" vs. "external" distinction altogether, adding RAR/MO3/OXM to the regular depackers list, and making depacking functions that wrap `execute_command` for RAR/MO3. The `execute_command` ugliness can then be entirely restricted to the depackers folder.

sagamusix:
> I'd never even heard of MO3 prior to working on this patch, but mixing non-free sample encodings into module formats sounds like a terrible idea that shouldn't be supported.

OpenMPT's MO3 decoder is (like the rest of OpenMPT) BSD-licensed and due to the expired patent status of MP3 a complete MO3 implementation can be done with libre software. It's just that noone besides me bothered with writing an MO3 decoder since there is UNMO3.

> Only a few of these seem to exist on the Mod Archive

I prefer having the original files on Mod Archive where they are available, that's one of the reasons. But the format is more popular than you might think, it was used for many demos and musicdisks in the early 2000s, and also the few games that still use module music.

sagamusix:
I think I also might have discussed this before with Claudio, xmp also really shouldn't be supporting unrar passthrough. It makes sense to support module-specific packers, but support for general-purpose archive formats should probably be best left to a layer above that.

AliceLR:
> I think I also might have discussed this before with Claudio, xmp also really shouldn't be supporting unrar passthrough. It makes sense to support module-specific packers, but support for general-purpose archive formats should probably be best left to a layer above that.

There has been talk before of separating both the depackers and ProWizard to separate libraries and I agree this is generally a good idea (aside from cases like MUSE, OXM, or Digital Symphony's LZW sample packing, where the packing isn't really separable from the format). Depacking RARs might be better left to needing a soft dependency e.g. libunarr, and it'd be easier to justify pulling in OpenMPT's MO3 code into a depacker library than libxmp directly.

sagamusix:
> and it'd be easier to justify pulling in OpenMPT's MO3 code into a depacker library than libxmp directly.

I should probably mention here that OpenMPT doesn't treat MO3 as a container - it directly decodes the MO3 data to OpenMPT internals, it doesn't convert the MO3 file to MOD/XM/S3M/IT/MTM as an intermediate step like unmo3 does. So it cannot be easily factored out of OpenMPT without taking a whole lot of other stuff with it (saving routines for those other formats, for instance), too.

> Digital Symphony's LZW sample packing

Funny that you mention that, I'm working on a Digital Symphony loader for OpenMPT. :) The LZW packing in the format indeed cannot be separated out, because it's interleaved with the rest of the file structure, and also because there are some peculiar differences compared to other LZW variants, so e.g. an LZW implementation from a GIF decoder would not work there...

AliceLR:
> So it cannot be easily factored out of OpenMPT without taking a whole lot of other stuff with it (saving routines for those other formats, for instance), too. But the format is completely documented at https://github.com/lclevy/unmo3/tree/master/spec so writing a new implementation isn't that hard (but a lot of work).

That is good to know, thanks. I took a brief look at it before while looking into alternative solutions to #327 and it looked like at least the decoding portions were self-contained, but I didn't check out the format-specific bits.

> Funny that you mention that, I'm working on a Digital Symphony loader for OpenMPT. :) The LZW packing in the format indeed cannot be separated out, because it's interleaved with the rest of the file structure, and also because there are some peculiar differences compared to other LZW variants, so e.g. an LZW implementation from a GIF decoder would not work there...

Just a heads up, if you're using libxmp's support as a base for that, I'd be careful. Something like half or more of the Digital Symphony modules I tried loading while testing `LIBXMP_NO_DEPACKERS` had samples that sounded quite corrupt. If you know what libxmp is doing wrong (maybe the aforementioned LZW differences?), feel free to open an issue though... :)

sagamusix:
> Just a heads up, if you're using libxmp's support as a base for that, I'd be careful. Something like half or more of the Digital Symphony modules I tried loading while testing LIBXMP_NO_DEPACKERS had samples that sounded quite corrupt. If you know what libxmp is doing wrong (maybe the aforementioned LZW differences?), feel free to open an issue though... :)

No, my loader is based on a format description I got from a friend (an old-time Acorn user, so he was interested in getting his old modules made on that platform playing in a modern tracker ;)). I did have to peek at xmp's LZW decompression to fix some of the special cases but I didn't copy it (it's based on GPLed code so that would be a bad idea for OpenMPT). xmp also doesn't support sigma-delta compression, which is something I want to support (have to sift through some ARM assembly code for that...). This might be the reason for the broken samples you were hearing.

sezero:
What are we waiting for this one?

AliceLR:
There was some concern that this fix is too dirty to actually merge, and we were waiting on feedback from @cmatsuoka re: whether the external depackers should be kept going forward (and if so, how).

sezero (approved):
This is OK with me. I still suggest additionally applying the djgpp patch
from https://github.com/libxmp/libxmp/issues/327#issuecomment-808722682

sezero:
>  we were waiting on feedback from @cmatsuoka re: whether the external depackers should be kept going forward (and if so, how).

I agree that at least many of the depackers are bloat, I am not sure that we handle it in 4.5.0.

AliceLR:
> I agree that at least many of the depackers are bloat, I am not sure that we handle it in 4.5.0.

OK. Would you say the following is fair to make this patch ready for merge, then?

1) I'll apply the DJGPP patch (sorry for not geting around to it yet...).
2) All `execute_command` cruft can be moved to a separate compilation unit in `src/depackers/`.
3) This might be a good opportunity to do the same thing to xfnmatch.{c,h} and `libxmp_exclude_match`, since these are also only used in the context of the depackers.

sezero:
That's OK by me.

sezero:
BTW, should we apply now or will you work on the improvements you suggested in this PR?

AliceLR:
I will work on the improvements suggested in this PR, sorry for the delay. I've been working on a quick and dirty implementation of #351 (API change) and want to give #360 a look too (IMO anything hanging or any bizarre hacks to prevent hanging should be considered a blocker until investigated further). These shouldn't take long, then I'll immediately get to work on this.

AliceLR:
Is there a good/foolproof way to add `unrar` detection to `test-dev/configure.ac` and install `unrar` to the CI environments? Currently `src/depackers/depacker.c` is at ~41% coverage and most likely that's entirely due to no coverage of the `execute_command` functions (can't tell for sure from codecov.io, GitHub is rate limiting me...).

sezero:
> Is there a good/foolproof way to add `unrar` detection to `test-dev/configure.ac` and install `unrar` to the CI environments?

No idea at all.

> edit: alternatively since if `unrar` invocation fails, the load functions return `-XMP_ERROR_DEPACK`, it's probably OK to just check for that return value and bypass the rest of the test (so no autoconf detection would be needed).

Sounds reasonable.

AliceLR:
I've added the RAR test. Time to see if the changes to .travis.yml were actually good enough to get this to work...

sezero:
What else is left for this?

AliceLR:
Nothing, I think. I'm still not sure about the platform whitelist for `popen`  but it's probably not going to hurt much to leave that alone.

sezero:
> I'm still not sure about the platform whitelist for `popen` but it's probably not going to hurt much to leave that alone.

We can handle that in other patchsets.

Issue #341, opened 2021-03-31
Files with particularly low coverage.


AliceLR:
From a quick check of codecov, here are some files that stick out as having particularly low coverage:
* `src/loaders/asif.c`: 0%
* `src/loaders/med2_load.c`: 3.16%
* `src/loaders/umx_load.c`: 11.48%
* `src/mix_paula.c`: 12.24%
* `src/loaders/prowizard/gmc.c`: 18.35%
* `src/loaders/sym_load.c`: 50% (I found a ton of these but a lot of them don't work in libxmp right now.)
* `src/loaders/amf_load.c`: 50.7%
* `src/tempfile.c`: 60.71%
* `src/loaders/masi_load.c`: 61.73%
* `src/loaders/gdm_load.c`: 64.29%

Most ProWizard loaders have fairly good coverage now, I think mainly thanks to the recent fuzzing efforts.

Previous patches I submitted without tests (these might help the GDM and OctaMED loaders): #98 #129 #236 #165 #187

I can put together some tests addressing a few of the format loaders soon. Any thoughts re: improving coverage for Paula emulation and tempfile.c?

AliceLR:
Merging #500 greatly improved coverage for both `src/period.c` (now 80%) and `src/loaders/far_load.c` (now 91.2%), so I'm removing both. (The new compilation unit `src/far_extras.c` is at 92.8%.)

PR #355, opened 2021-06-02
Fix numerous DSMI AMF bugs reported in issue #350.

### LINKED ISSUE #350: AMF findings
I recently found that [DSMI](https://github.com/ochrons/dsmi/) was open-sourced by its author and dug through its code to solve some of the mysteries surrounding missing format revisions, and claims made in many older AMF loaders (such as that of libxmp).

Here are some things that should be addressed in libxmp:
- The panning command is 7-bit panning + surround, like in S3M. In fact, I asked Otto Chrons if DMP/DSMI is the source of this command, because it's the earliest source I have found, and he recalls that he added those commands to MOD/S3M at a time when the trackers themselves didn't.
- In AMF version 1 (currently not supported by libxmp, and I have no such files), the length of each track should be extended by 1 (i.e. 3 extra bytes compared to later revisions), unless the track length is 0.
- The track length is in fact not a 24-bit integer, but a 16-bit integer followed by a track type byte. Most of the time this type byte is 0, but in `AVOID.AMF` (can be found in the DSMI repository linked above) it isn't always 0, so the tracks start to appear to be huge halfway through.
- In file versions < 10, the loop start and end are 16-bit variables instead of 32-bit variables. There are v8 and v9 files in the DSMI repository for testing.
- File versions 1 and 8 are hardcoded to 4 channels (the byte containing the number of channels is not present in the header)
- Panning table for 16 channels is introduced in format version 9.
- There is some weird fingerprinting for files that supposedly only contain a 16-bit loop start and no loop end. I think this is a fairy tale coming from Miodrag Vallat's AMF.TXT. There are even two filenames cited in the source, but not even the original DSMI source code handles this special case (while otherwise keeping excellent backwards compatibility), so I would suspect that those files were simply broken. Without seeing those files myself, I refuse to believe that this information is correct. :)

Note: DSMI itself rejects file versions 0 and 2-7, so I suppose they don't exist.

AliceLR:
Fixes #350.

Fixed bugs that were reported by @sagamusix:
* Support for DSMI modules from versions 0.1, 0.8, and 0.9 has been implemented.
  * Sample length, loop start, and loop end sizes for versions <1.0 were incorrect. These are all 16-bit and present in valid files, and a loop end value of `0xffff` indicates no loop.
  * Version 0.1 modules should have +1 added to the track size (unverified).
  * Version 0.1 and 0.8 modules are fixed to 4 channels and don't have a channel count byte.
  * Version 0.1 and 0.8 modules do not have the channel remap table, but 0.9 does.
  * Fingerprinting for unusual DSMI 1.0 modules with no sample loop end values has been restricted specifically to version 1.0 (as it was breaking 0.8 and 0.9 modules).
* Loading of track sizes incorrectly treated the the size as a 24-bit integer, when it should be 16-bit (the extra byte is unused).
* Event rows past the current track length caused modules to be rejected instead of being treated as the track end.
* The panning effect was incorrectly implemented, causing panning values of `0x40` (full right) to overflow to 0 and ignoring `0x64` for surround (same as S3M `0xA4`).

Fixed additional bugs I found:
* Buggy track index mapping caused references to the blank track 0 to be replaced with references to the first track.
* Volume values were not correctly incremented by 1 when converted to event volume.
* Volume values of `0xff` indicate that the event does not set volume.
* Incorrect note `0x7f` handling has been removed.

This patch also adds (probably too many) DSMI test modules, which should help with #341 a little bit.


src/loaders/amf_load.c
```diff
@@ -314,10 +344,21 @@ static int amf_load(struct module_data *m, HIO_HANDLE *f, const int start)
...
+		 * just a word. The purpose of the third byte is unknown, and
+		 * DSMI just ignores it.
+		 */
+		size = hio_read16l(f);
+		hio_read8(f);
+
 		if (hio_error(f))
 			return -1;
 
+		/* Version 0.1 AMFs apparently have an extra event when the
```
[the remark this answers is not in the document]
AliceLR: Yeah, I agree that would definitely be a better way to explain it, and it makes sense. I'll patch this comment after I've found a copy of M2AMF that produces 0.1 AMFs on this BBS CD archive (or more likely, after I give up, because it feels like I've been finding every version except for the ones relevant to AMF 0.1 and 1.0).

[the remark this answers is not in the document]
AliceLR: I updated this comment.

src/loaders/amf_load.c
```diff
@@ -169,41 +184,45 @@ static int amf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		for (i = 0; i < mod->ins; i++) {
 			b = hio_read8(f);
 			if (b != 0 && b != 1) {
-				ver = 0x09;
+				no_loopend = 1;
 				break;
 			}
 			hio_seek(f, 32 + 13, SEEK_CUR);
 			if (hio_read32l(f) > 0x100000) { /* check index */
```
sagamusix, line 191: index should never be greater than the number of instruments (as per how it's loaded in the AMF loader of DSMI), so this check can be tightened.

AliceLR: That's what I thought, I just hadn't looked into it. I've updated this check.

src/loaders/amf_load.c
```diff
@@ -264,12 +295,19 @@ static int amf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 
 	/* Tracks */
 
-	trkmap = calloc(sizeof(int), mod->trk);
+	/* Index 0 is a blank track that isn't stored in the file. To keep
```
sagamusix, line 298: I don't think it's really necessary to document the bugs of older XMP versions here (and further below).

sezero (approved):

PR #358, opened 2021-06-04
Remove test-dev/md5.c, use md5.c from src/ for regression tests.

AliceLR:
This removes the redundant md5.c/h from the test-dev directory and makes the regression tests build with the normal libxmp MD5 implementation instead. The duplicate md5.c only differs by a single line currently and made it possible for the implementations to diverge. The regression tests already pull in other objects from src/, so there was no reason for this one to be duplicated.

This was probably also negatively affecting the code coverage report of src/md5.c (see #341).

edit: since everything looks OK now, I'm going ahead and marking this ready for review. I'll leave test/md5.c as an exercise for someone else... : )


AliceLR:
Found a minor issue (building `gen_module_data` broke) that should be resolved now. I'm going to let this run the tests again and mark it ready when it's done.

sezero:
We may make `test` directory to use `src/md5.*` too, what do you think?

BTW, do you think that we should add the `extern "C"` markers to `src/md5.h`?

AliceLR:
> We may make `test` directory to use `src/md5.*` too, what do you think?

I took a quick look at this and decided not to touch it because `test/Makefile` (also, the corresponding targets in `Makefile.vc`/`watcom.mif`) lacks the existing infrastructure for pulling in `src/` objects that `test-dev/Makefile.in` does. This is a lazy excuse though, and I can definitely add that. I can test the main and MSVC Makefiles, but I'll need help testing the Watcom Makefiles.

> BTW, do you think that we should add the `extern "C"` markers to `src/md5.h`?
> (The `test-dev` version has them.)

I don't think there is any harm in adding these to `src/md5.h`, so I'll go ahead and do that. libxmp's core code is pure C right now but who knows if any C++ will ever need to creep in. I just had to rewrite MZX's mixers in templates because the macros that were used before were borderline incomprehensible. I notice libxmp's `mix_all.c` is similarly cursed... :)

AliceLR:
Interesting bit of history on the `extern "C"` usage that I checked into since the `md5.h` files had obviously diverged: Claudio removed the `cdefs.h` include entirely in this commit https://github.com/libxmp/libxmp/commit/8328cd0cbba3be7c5a0f1bd00f4cfc14c5d577c9, but the `extern "C"` usage derives from a fix (as opposed to removal) you committed in 2019 https://github.com/libxmp/libxmp/commit/1d161113554e6f5a6b493b476883dd44d506dc23.

IMO generalizing them (preferably with something like `#define LIBXMP_BEGIN_DECLS` etc.) is better than total removal, but it's not like there's any C++ in libxmp for it to matter. I'll go ahead and copy your version of these into `md5.h` for now.

AliceLR:
Working on a fix for the libxmp-lite issue. It's not 100% clear to me how the lite dist test folder is constructed and libxmp-lite explicitly does not include md5.o, so I just moved the md5.{c,h} files from test/ to lite/test/. (I also ran into the issue #356 fixed and had to rebase...)

sezero (approved):

PR #359, opened 2021-06-04
Fix DSMI AMF speed effect conversion.

AliceLR:
The AMF speed effect supports the full range of values from 0-255 like S3M, and should be converted to the S3M speed effect. Like GDM, however, it was using the MOD speed effect instead, which treats values >0x20 as BPMs unless a separate quirk is enabled.

This fixes a bug where `Metromania Partie 1.amf` did *not* have over 2 minutes of silence at the end (see https://github.com/sezero/mikmod/pull/42).


sezero (approved):
OK.

PR #362, opened 2021-06-05
Implement xmp_*_module_from_callbacks API functions.

### LINKED ISSUE #351: [Feature request] Add a xmp_load_module function for custom streams.
libopenmpt, MikMod, and libvorbis are all capable of opening files from a custom stream specified by a user-provided stream pointer and set of read/seek/tell callbacks. On the other hand, libxmp currently can *not* do this, and either a filename, `FILE *`, or buffer needs to be provided. This means that if the user is using their own IO abstraction either the underlying `FILE *`/buffer needs to be used (if available) or the stream content needs to be copied to a temporary buffer or file.

Considering 4.5.0 is still blocked on #344, it seems like adding an `HIO_HANDLE` type that operates on user-provided read/seek/tell callbacks (and corresponding `xmp_load_module` and `xmp_test_module` variants) could be added without further delaying a release. (This was briefly discussed in #134 due to the `fdopen` leak bug.)

Any thoughts?

AliceLR:
Here's an initial messy implementation of load/test from callbacks API functions. These allow a user-defined stream to be provided to libxmp instead of requiring a memory buffer or real file.

These changes almost definitely need revisions. Particularly, I hate using `unsigned long` instead of `size_t` for the read function, but I wasn't sure if adding an `#include <stddef.h>` to `xmp.h` is OK or not. Other aspects of this patch are likely questionable too.

Fixes #351.

edit: I should note that I don't strictly need these functions, I just thought they might be convenient. If these functions are more trouble than they're worth to keep around it's OK to just reject this patch. Specifically, I think the performance of repeatedly calling into these callbacks to load modules might be pretty bad, and there's currently no form of buffering implemented.


include/xmp.h
```diff
@@ -320,6 +320,11 @@ struct xmp_frame_info {			/* Current frame information */
 	struct xmp_channel_info channel_info[XMP_MAX_CHANNELS];		/* Current channel information */
 };
 
+struct xmp_callbacks {
+	unsigned long (*read_func)(void *dest, unsigned long len, unsigned long nmemb, void *priv);
+	int           (*seek_func)(void *priv, long offset, int whence);
+	long          (*tell_func)(void *priv);
+};
```
sezero, line 327: Users may want a `close_func` here

AliceLR: I think that's a reasonable addition. Do you think it's OK to add an stddef.h include to give `read_func` some `size_t`s instead of `unsigned long`? (It'd only be a "problem" on x86-64 Windows with extremely large files...)

sezero: If someone crazy truly wants to read more than 4g, then it's a problem - we
don't do that within libxmp. Can you think of any other sources of issue?

As for stddef.h inclusion in xmp.h: I'm not against it, nor am I objecting it - 
only if really necessary. Comments? @cmatsuoka, @sagamusix?

sagamusix: I can only speak from libopenmpt perspective where we used stddef.h/cstddef everywhere: short/long/long long etc are all terrible names and as far as I'm concerned they cannot disappear quickly enough.

sezero: Well, thinking of it, (unsigned) long vs size_t _might_ be problem with modules
in large containers where the tell function would report wrong values.

sagamusix: File APIs in general should not depend on pointer size. It is entirely possible to operate on files > 4 GB in a 32-bit process if done correctly. Any offsets really should be (u)int64_ts.

sezero: That makes us guarantee things.

Then again, the callback api doesn't know anything about the file: the client itself does.
The (unsigned) long use would be safe provided that the module files themselves are
not really 'large' files even if they are contained in large containers.

So, speaking for myself, I'm good with (unsigned) long use, but I don't object to size_t.
(u)int64 on the other hand, would dictate stdint.h.

sagamusix: It's 2021. Is someone really going to compile the very latest version of libxmp on a platform that doesn't have stdint.h?

sezero: > It's 2021. Is someone really going to compile the very latest version of libxmp on a platform that doesn't have stdint.h?

The only toolchain that doesn't provide stdint.h possibly would be MSVC < 2013 (?),
which we can handle by some ugliness in our header.

sagamusix: I think at this point, if someone wants to keep using such an old MSVC version, they can be expected to provide their own stdint.h. It was already a common practice some years ago when those were the latest versions of MSVC.

sezero: @cmatsuoka, @AliceLR, what do you think?

If I have to vote, I vote for keeping (unsigned) long as they are.

AliceLR: Another thing—should `xmp_test_module_from_callbacks` ignore `close_func` if it's added? Presumably you would want to provide the same open stream and callbacks to both, but if the callbacks included a `close_func`, you'd either need two sets of callbacks or to reopen the stream...

sezero: Ugh... Well, if `close_func` is not NULL, I'd think that the user implies that we should close the thing?

AliceLR:
(Rebased to `master` to resolve the `src/Makefile` merge conflict.)

include/xmp.h
AliceLR: Okay, now that I've caught up on the backlog (I had this page left open when I commented and GitHub didn't load the conversation from 3 hours ago in...).

My take on `stdint.h` is roughly the same for my own projects but in this case libxmp has done a good job of steering clear of C99-isms, which would cause issues with older MSVCs (which hopefully everyone has updated away from). Adding an include for a C99 header, particularly in a public API header like `xmp.h` seems questionable.

In the case of `seek_func` and `tell_func` I used `long` mainly because the corresponding `stdio.h` functions also use those. Using 64-bit types for offsets would be preferable but if the only good way to do it is `stdint.h` in a public header... ew.

Re: `size_t` I definitely prefer it to `unsigned long` for the case of `read_func`. I think for that particular function that (unlike seek/tell) it's incredibly unlikely that 64-bit values will be required there ever.

sezero: Even `size_t` won't be necessary if we assume that there are no modules
which are larger than 4GB (see my points at https://github.com/libxmp/libxmp/pull/362#discussion_r646041441)

sezero: > Even `size_t` won't be necessary if we assume that there are no modules
> which are larger than 4GB (see my points at [#362 (comment)](https://github.com/libxmp/libxmp/pull/362#discussion_r646041441))

Besides, it's no different from `xmp_load_module_from_memory` having a `long size` parameter.
So I vote that we keep the `(unsigned) long`s as they are in this.

AliceLR: I'll go ahead and just leave the `long`/`unsigned long` usage then (it can be revisited for 5.0 if it really ends up being a problem, which I doubt...).

sezero: OK then, the only thing left is the `close_func` being implemented?

AliceLR: I think so. (It could probably use some other light cleanup and better documentation too.)

AliceLR: I've implemented this function. There was another thing I want to check into, which is that for whatever reason the Prowizard module in the `xmp_test_module_from_callbacks` test wasn't working correctly, and I ended up removing it. Prowizard modules should work fine with this function so I don't know what the deal is.

sezero:
>  for whatever reason the Prowizard module in the `xmp_test_module_from_callbacks` test wasn't working correctly, and I ended up removing it. Prowizard modules should work fine with this function so I don't know what the deal is.

Never looked at prowizard stuff seriously to make a guess here.
Was there an error message or something?

> edit: It might be that Prowizard's check function relies on `fread` past EOF not actually setting the error indicator unless the stream starts at EOF?

Eh?

AliceLR:
That was completely wrong by the way. I should have checked it first, but the test module that the FILE version of the test uses is compressed...

sezero (approved):
This is good.

PR #366, opened 2021-06-06
Add MED2 format test, fix MED2 BPM handling.

AliceLR:
Here's a MED2 format loader test since that loader's coverage is so low due to not having one (3.16%). Turns out the BPM handling was total junk, so I fixed that while I was at it! This format only supports song files, so you have to play it from a directory where `med2test_PanFlute` exists or use `xmp_set_instrument_path` to get the sample to load.

The MED3 and MED4 BPM handling is very suspicious too, but I didn't touch it here (uses the `125 * bpm / 33` formula which overflows libxmp's effect parameter instead of `MED_TIME_FACTOR`, at a glance the handling of BPMs 1-10 is wrong too). Note these are the *old* MED formats, not the OctaMED formats where this stuff was already fixed. : )


AliceLR:
(Rerunning the coverage test even though it passed since Codecov didn't seem to receive any info from the last one...)

sezero (approved):

PR #370, opened 2021-06-07
Fix GCC warning in iff.c due to (correct) strncpy usage.

AliceLR:
Another minor tidying patch. Most of the -Wstringop-truncation warnings in libxmp were fixed by #179, but the `strncpy` in iff.c had to be reverted in #184 because, for a change, it's actually being used the way it was intended. GCC has been complaining about it since, though, and the simplest non-breaking fix seems to be to just expand the buffer size so it can be terminated. Not that it matters much, but this is a slight size increase for `iff_info` for 32-bit machines and was existing padding for 64-bit.

With this patch, neither GCC 10 nor GCC 11 emits any warnings on my end during the libxmp build process. :)

### DIFF
--- a/src/loaders/iff.c
+++ b/src/loaders/iff.c
@@ -162,12 +162,18 @@ int libxmp_iff_register(iff_handle opaque, const char *id,
 {
 	struct iff_data *data = (struct iff_data *)opaque;
 	struct iff_info *f;
+	int i = 0;
 
 	f = malloc(sizeof(struct iff_info));
 	if (f == NULL)
 		return -1;
 
-	strncpy(f->id, id, 4);
+	/* Note: previously was an strncpy */
+	for (; i < 4 && id && id[i]; i++)
+		f->id[i] = id[i];
+	for (; i < 4; i++)
+		f->id[i] = '\0';
+
 	f->loader = loader;
 
 	list_add_tail(&f->list, &data->iff_list);


sezero:
Another solution would be something like:
```diff
diff --git a/src/loaders/iff.c b/src/loaders/iff.c
index e6edf04..7b22fd4 100644
--- a/src/loaders/iff.c
+++ b/src/loaders/iff.c
@@ -162,12 +162,16 @@ int libxmp_iff_register(iff_handle opaque, const char *id,
 {
 	struct iff_data *data = (struct iff_data *)opaque;
 	struct iff_info *f;
+	int i = 0;
 
 	f = malloc(sizeof(struct iff_info));
 	if (f == NULL)
 		return -1;
 
-	strncpy(f->id, id, 4);
+	for ( ; i < 4 && id[i]; ++i)
+		f->id[i] = id[i];
+	for ( ; i < 4; ++i)
+		f->id[i] = 0;
 	f->loader = loader;
 
 	list_add_tail(&f->list, &data->iff_list);
```

AliceLR:
I did try out a few different fixes and something like that was one of them. It did work just fine and doesn't touch the struct, so if you'd think it's worth the extra verbosity, I'm OK with directly replacing `strncpy`.

sezero:
IMO, Let's keep the struct as is.

AliceLR:
OK, I've updated it to use an `strncpy` replacement instead.

sezero (approved):

AliceLR:
> Do we really need the `id` check here? (The first one, not the `id[i]` one.)

I added that on the off chance this function somehow gets misused internally (ideally there'd be a compile-time way to enforce passing a non-null string literal there, oh well). I suspect reasonable compilers will unswitch that check due to unrolling those loops, but worst case I don't think it will seriously hurt anything.

sezero:
OK, let's keep it.

PR #371, opened 2021-06-07
Update date in copyright notice for 4.5.0.

AliceLR:
Updates the dates in the copyright notices for all files that had "-2018". I was originally limiting this to files modified since the last update, but it turned out there were so few that weren't modified that I just replaced them all.

### DIFF
--- a/lite/README
+++ b/lite/README
@@ -20,7 +20,7 @@ Libxmp API.
 LICENSE
 
 Extended Module Player Lite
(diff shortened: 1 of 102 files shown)


sezero:
lite/src/format.c, lite/src/loaders/mod_load.c and lite/README need
updating, too.

AliceLR:
Good catch, I forgot to check the lite/ dir.

sezero (approved):

PR #372, opened 2021-06-07
misc cleanups / tidy-ups:

sezero:
This patchset greatly reduces the number of errors from c++ compilers.
The normal C builds are not affected - e.g. the watcom-built os/2 dlls
before and after this patchset are identical.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -321,6 +321,33 @@ struct module_data {
 };
 
 
+struct pattern_loop {
+	int start;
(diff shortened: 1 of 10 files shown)


AliceLR (approved):
I haven't seen warnings from these particular things but the only C++ compiler I've used to build libxmp is MSVC (which only has a large number of signed/unsigned compare and narrowing warnings). This looks good to me, though.

PR #380, opened 2021-06-18
Fix crash when xmp_set_row is used on an IT end marker.

AliceLR:
`xmp_set_position` allows setting the current order to an IT end marker, but `xmp_set_row` was missing safety checks for this situation.

### DIFF
--- a/src/control.c
+++ b/src/control.c
@@ -185,16 +185,17 @@ int xmp_set_row(xmp_context opaque, int row)
 	struct xmp_module *mod = &m->mod;
 	struct flow_control *f = &p->flow;
 	int pos = p->pos;
-	int pattern = mod->xxo[pos];
+	int pattern;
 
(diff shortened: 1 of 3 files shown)


AliceLR:
(Looks like Travis-CI broke because it moved to travis-ci.com... ¯\\\_(ツ)_/¯)

sezero (approved):

sezero:
I'll merge this manually

> (Looks like Travis-CI broke because it moved to travis-ci.com... ¯\_(ツ)_/¯)

How do we fix it for us?

AliceLR:
> How do we fix it for us?

I don't know—I was under the impression the migration was supposed to be automated but if I try to view libxmp's page on travis-ci.com I get permission denied. I'm not familiar with how Travis integration works. @cmatsuoka might know how to fix this?

PR #382, opened 2021-06-19
Allow up to 255 sequences to be scanned.

AliceLR:
The `sequence_control` array theoretically allows up to 255 sequences to exist, but because `player_data::scan` and `module_data::seq_data` are allocated as part of their respective structs, the sequence count is limited to 16. This patch raises that limit and moves `player_data::scan`, the larger of the two scan arrays, to a separate heap allocation.

Large numbers of sequences may be used in games relying on dynamic order changes, which is somewhat common in MegaZeux games (the module I was given, which I'm not sure I'm allowed to share, contains 18). Since the scan doubles as a validity check for module flow, increasing the number of sequences is the easiest way to safely support this.

This should not affect the API at all: `xmp_module_info` already uses a pointer to an array of an arbitrary number of `xmp_sequence` structs, and since this array is still allocated as part of the `xmp_context`, this pointer will never change during the lifetime of any given `xmp_context`. `player_data::scan` was never exposed to the API, so (re)allocating it should break nothing.


AliceLR:
Manual AddressSanitizer (Fedora 34, clang 12) and Visual Studio (x86_64, MSVC 2019) tests passed. (It seems that at some point Fedora and/or clang's MemorySanitizer support for `stat()` without instrumentation broke, so I don't have a useful result there. The tests get spammed with failures due to `struct stat st` being "uninitialized" in load.c.)

sezero (approved):
Looks fair to me. @cmatsuoka, @sagamusix: Is this OK?

sagamusix:
Concept-wise it's all fine by me, can't comment much on the implementation itself

sezero:
Mergin this. Let's see how it goes.

PR #383, opened 2021-06-21
Add libFuzzer frontend for libxmp.

### LINKED ISSUE #302: OSS-Fuzz and/or libFuzzer integration.

AliceLR:
Based on Lionel Debroux's initial libFuzzer frontend from #302.

@debrouxl any thoughts/recommendations?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,8 +15,11 @@
 *.cov
 a.out
 lib/
+test-dev/gen_mixer_data
+test-dev/gen_module_data
 test-dev/libxmp-covertest
-test-dev/libxmp-test
+test-dev/libxmp-tests
+test-dev/libxmp_fuzz_*
 
 # VC++ files
 /.vs
@@ -73,7 +76,11 @@ test/.test
 test/libxmp-covertest
 test/libxmp-tests
 test/test8.raw
+test-dev/.fuzzer
 test-dev/.read_test
+test-dev/ARTIFACTS
+test-dev/CORPUS
+test-dev/NEW_CORPUS
 test-dev/all_tests.c
 test-dev/write_test
 coverage/
--- a/.gitignore
+++ b/.gitignore
@@ -74,8 +74,7 @@ libxmp.pc
 Makefile
 test/.test
 test/libxmp-covertest
-test/libxmp-tests
-test/test8.raw
+test/libxmp-test
 test-dev/.fuzzer
 test-dev/.read_test
 test-dev/ARTIFACTS
--- a/test-dev/Makefile.in
+++ b/test-dev/Makefile.in
@@ -644,7 +644,7 @@ dist-test:
 	mkdir -p $(DIST)/$(TEST_PATH)
 	cp -RPp $(addprefix $(TEST_PATH)/,$(TEST_DFILES)) $(DIST)/$(TEST_PATH)
 
-clean:
+clean: fuzzerclean
 	rm -f *.o core *~ $(T_OBJS)
 
 distclean: clean
@@ -722,3 +722,65 @@ $(TEST_PATH)/all_tests.txt: $(TEST_PATH)/Makefile
 	done
 
 $(addprefix $(SRC_PATH)/,$(TEST_INTERNAL)): $(SRC_PATH)/player.h
+
+#
+# Fuzzers
+#
+
+FUZZLIB_PATH		= .fuzzer
+FUZZ_ASAN_FLAGS		= -O3 -fsanitize=fuzzer,address -fno-omit-frame-pointer -g
+FUZZ_MSAN_FLAGS		= -O3 -fsanitize=fuzzer,memory -fsanitize-memory-track-origins=2 -fno-omit-frame-pointer -g
+FUZZ_UBSAN_FLAGS	= -O3 -fsanitize=fuzzer,undefined -fno-sanitize-recover=all -fno-sanitize=shift-base -fno-omit-frame-pointer -g
+FUZZLIB_ASAN_FLAGS	= $(subst fuzzer,fuzzer-no-link,$(FUZZ_ASAN_FLAGS))
+FUZZLIB_MSAN_FLAGS	= $(subst fuzzer,fuzzer-no-link,$(FUZZ_MSAN_FLAGS))
+FUZZLIB_UBSAN_FLAGS	= $(subst fuzzer,fuzzer-no-link,$(FUZZ_UBSAN_FLAGS))
+
+.PHONY: fuzzers fuzzerclean $(FUZZLIB_PATH)
+
+fuzzers: libxmp_fuzz_asan libxmp_fuzz_msan libxmp_fuzz_ubsan
+
+fuzzerclean:
+	rm -rf $(FUZZLIB_PATH) libxmp_fuzz_asan libxmp_fuzz_msan libxmp_fuzz_ubsan
+
+$(FUZZLIB_PATH):
+	mkdir -p $(FUZZLIB_PATH)
+	cp -r ../configure.ac ../Makefile.in ../libxmp.pc.in ../docs ../include ../src ../test $(FUZZLIB_PATH)
+
+$(FUZZLIB_PATH)/libxmp-asan.a: $(FUZZLIB_PATH)
+	(cd $(FUZZLIB_PATH) && \
+		autoconf && \
+		CC="clang" CFLAGS="$(FUZZLIB_ASAN_FLAGS)" LDFLAGS="$(FUZZLIB_ASAN_FLAGS)" ./configure && \
+		$(MAKE) static && \
+		cp lib/libxmp.a libxmp-asan.a && \
+		$(MAKE) distclean)
+
+$(FUZZLIB_PATH)/libxmp-msan.a: $(FUZZLIB_PATH) | $(FUZZLIB_PATH)/libxmp-asan.a
+	(cd $(FUZZLIB_PATH) && \
+		autoconf && \
+		CC="clang" CFLAGS="$(FUZZLIB_MSAN_FLAGS)" LDFLAGS="$(FUZZLIB_MSAN_FLAGS)" ./configure && \
+		$(MAKE) static && \
+		cp lib/libxmp.a libxmp-msan.a && \
+		$(MAKE) distclean)
+
+$(FUZZLIB_PATH)/libxmp-ubsan.a: $(FUZZLIB_PATH) | $(FUZZLIB_PATH)/libxmp-msan.a
+	(cd $(FUZZLIB_PATH) && \
+		autoconf && \
+		CC="clang" CFLAGS="$(FUZZLIB_UBSAN_FLAGS)" LDFLAGS="$(FUZZLIB_UBSAN_FLAGS)" ./configure && \
+		$(MAKE) static && \
+		cp lib/libxmp.a libxmp-ubsan.a && \
+		$(MAKE) distclean)
+
+libxmp_fuzz_asan: libxmp_fuzz.c $(FUZZLIB_PATH)/libxmp-asan.a
+	@CMD='$(LD) $(FUZZ_ASAN_FLAGS) -o $@ $< -L$(FUZZLIB_PATH) -lxmp-asan'; \
+	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo LD $@ ; fi; \
+	eval $$CMD
+
+libxmp_fuzz_msan: libxmp_fuzz.c $(FUZZLIB_PATH)/libxmp-msan.a
+	@CMD='$(LD) $(FUZZ_MSAN_FLAGS) -o $@ $< -L$(FUZZLIB_PATH) -lxmp-msan'; \
+	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo LD $@ ; fi; \
+	eval $$CMD
+
+libxmp_fuzz_ubsan: libxmp_fuzz.c $(FUZZLIB_PATH)/libxmp-ubsan.a
+	@CMD='$(LD) $(FUZZ_UBSAN_FLAGS) -o $@ $< -L$(FUZZLIB_PATH) -lxmp-ubsan'; \
+	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo LD $@ ; fi; \
+	eval $$CMD
--- a/test-dev/libxmp_fuzz.c
+++ b/test-dev/libxmp_fuzz.c
@@ -1,0 +1,32 @@
+#include <stddef.h>
+#include <stdint.h>
+#include <stdio.h>
+#include "../include/xmp.h"
+
+#ifdef __cplusplus
+extern "C"
+#endif
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+	xmp_context opaque = xmp_create_context();
+	FILE *f;
+
+	/* Fuzz loaders. */
+	if (xmp_load_module_from_memory(opaque, data, size) == 0)
+	{
+		/* FIXME fuzz playback. */
+		xmp_release_module(opaque);
+	}
+
+	/* Fuzz depackers. */
+	f = fmemopen((void *)data, size, "rb");
+	if (f != NULL)
+	{
+		struct xmp_test_info info;
+		xmp_test_module_from_file(f, &info);
+		fclose(f);
+	}
+
+	xmp_free_context(opaque);
+	return 0;
+}
--- a/test-dev/libxmp_fuzz.sh
+++ b/test-dev/libxmp_fuzz.sh
@@ -1,0 +1,56 @@
+#!/bin/sh
+
+[ ! -f "libxmp_fuzz_asan" ] &&  { echo "Couldn't find libxmp_fuzz_asan. Run 'make fuzzers'!"; exit 1; }
+
+#export LD_LIBRARY_PATH=".fuzzer"
+mkdir -p "ARTIFACTS"
+
+DEFAULT_PARAMETERS="CORPUS data/ openmpt/ -artifact_prefix=ARTIFACTS/ -timeout=30"
+
+COMMAND="$1"
+shift
+
+case "$COMMAND" in
+	#
+	# Merge corpus.
+	#
+	merge)
+		NEW_CORPUS="NEW_CORPUS"
+		OLD_CORPUS="CORPUS"
+		[ -n "$1" ] && { NEW_CORPUS="$1"; }
+		[ -n "$2" ] && { OLD_CORPUS="$2"; }
+		mkdir -p "$NEW_CORPUS"
+		./libxmp_fuzz_asan -merge=1 "$NEW_CORPUS" "$OLD_CORPUS"
+		./libxmp_fuzz_msan -merge=1 "$NEW_CORPUS" "$OLD_CORPUS"
+		./libxmp_fuzz_ubsan -merge=1 "$NEW_CORPUS" "$OLD_CORPUS"
+		;;
+
+	#
+	# Fuzz with given parameters.
+	#
+	asanx)
+		./libxmp_fuzz_asan -artifact_prefix="ARTIFACTS/" "$@"
+		;;
+	msanx)
+		./libxmp_fuzz_msan -artifact_prefix="ARTIFACTS/" "$@"
+		;;
+	ubsanx)
+		./libxmp_fuzz_ubsan -artifact_prefix="ARTIFACTS/" "$@"
+		;;
+
+	#
+	# Fuzz with default parameters.
+	#
+	asan)
+		mkdir -p "CORPUS"
+		./libxmp_fuzz_asan $DEFAULT_PARAMETERS "$@"
+		;;
+	msan)
+		mkdir -p "CORPUS"
+		./libxmp_fuzz_msan $DEFAULT_PARAMETERS "$@"
+		;;
+	ubsan)
+		mkdir -p "CORPUS"
+		./libxmp_fuzz_ubsan $DEFAULT_PARAMETERS "$@"
+		;;
+esac


sagamusix:
From experience with fuzzing libopenmpt, I would suggest to also add the following calls to the test program after the module was loaded successfully:

- Set play position (both order/row and duration in seconds, depending on what libxmp supports)
- Render a few frames of audio
- Maybe retrieval of sample texts etc.

The first two items on that list should preferably be interleaved to catch possible state inconsistencies that could result in bugs (I've had a few of those in libopenmpt so it's possible that libxmp suffers from similar bugs).

AliceLR:
> From experience with fuzzing libopenmpt, I would suggest to also add the following calls to the test program after the module was loaded successfully:

I think fuzzing these would be a very good idea. libxmp supports both order+row and time point seeking in its API, and I did just run into an `xmp_set_row` crash that shouldn't have been hard to catch. I need to do some testing to see how much (if at all) that would affect the duration of runs, in case it needs to be separate to keep loader/depacker fuzzing faster.

debrouxl:
This is a good first step.
One minor suggestion and two general notes:
* given that at some point, there ought to be targets for building the front-end - and therefore the library as a prior task - with undefined and memory sanitizer instrumentation, we probably want to name the target in `test-dev/Makefile.in` e.g. `libxmp_fuzz_asan` from the get go, or even add both of these `libxmp_fuzz_ubsan` and `libxmp_fuzz_msan` targets at the same time ?
* past fixes for memory errors, usage of undefined behaviour and excessive CPU usage have made the lack of temp file cleanup code (much) less relevant than it was when I highlighted that limitation of my simplistic fuzzing front-end in #302 , although I don't think I checked for real: ISTR that the `while true; do find ... -delete; sleep; done` loop was still running in the background throughout my latest tests with the simplistic fuzzing front-end. This can be revisited later if the need arises.
* Saga Musix's suggestions are indeed pretty useful, all the more it looks like that at least one bug was caught up as a result :)

AliceLR:
> * given that at some point, there ought to be targets for building the front-end - and therefore the library as a prior task - with undefined and memory sanitizer instrumentation, we probably want to name the target in `test-dev/Makefile.in` e.g. `libxmp_fuzz_asan` from the get go, or even add both of these `libxmp_fuzz_ubsan` and `libxmp_fuzz_msan` targets at the same time ?

I think there's a couple ways to go about doing this.

A similar setup to what is used to static vs. shared vs. coverage builds might—essentially have a completely separate set of rules in the main Makefile to build fuzzing libraries. A separate set of rules for every set of sanitizer flags that are needed would probably get tedious so I'm not sure what the best way to address this is.

Alternatively, there could be test-dev meta targets to configure and build all of the separate fuzzing libraries in a duplicate source tree. This would probably be cleaner and not impact anything outside of test-dev at all, so I think I'll look at trying this first.

> * past fixes for memory errors, usage of undefined behaviour and excessive CPU usage have made the lack of temp file cleanup code (much) less relevant than it was when I highlighted that limitation of my simplistic fuzzing front-end in #302 , although I don't think I checked for real: ISTR that the `while true; do find ... -delete; sleep; done` loop was still running in the background throughout my latest tests with the simplistic fuzzing front-end. This can be revisited later if the need arises.

To make sure I'm understanding the problem correctly: libxmp unlinks tempfiles after loading, but in the event of a depacker or loader crash this doesn't happen, so they accumulate? I wonder if there's a good way to insert something for this into the fuzzing function directly so it doesn't rely on a separate background process.

debrouxl:
> To make sure I'm understanding the problem correctly: libxmp unlinks tempfiles after loading, but in the event of a depacker or loader crash this doesn't happen, so they accumulate?

Exactly. But given that depacker / loader crashes have become pretty infrequent, thanks to the fixes mainly performed by you after my reports, this is no longer a meaningful issue. I just wanted to post an update on this matter I raised in #302 as a review comment on this PR, so that you don't spend time on creating a workaround for it :)

AliceLR:
Updated patch pushed with initial Makefile targets to generate the fuzzers. It's kind of slow to build the full set (even with `make fuzzers -j8`), but since fixing these issues in batches seems to work best maybe they won't actually need to be rebuilt that often.

The new MSan fuzzer is *already* finding loader bugs somehow, so I'm glad you brought those up.

Next up will be improving the test program so it covers more than just the loaders/depackers/ProWizard/scan.

> Exactly. But given that depacker / loader crashes have become pretty infrequent, thanks to the fixes mainly performed by you after my reports, this is no longer a meaningful issue. I just wanted to post an update on this matter I raised in #302 as a review comment on this PR, so that you don't spend time on creating a workaround for it :)

Alright, I'll not worry about that for now then... :)

AliceLR:
I rebased this to master and updated the UBSan flags to 1) disable recovery so it's easier to pinpoint problem files and 2) disable `shift-base` since it's going to flag almost everything when playback/mixing are added to the test function.

AliceLR:
I've added a call to `xmp_test_module_from_file` (via `fmemopen`) to the fuzzing function to make sure the depackers get covered as well as the loaders. Still no playback/seeking fuzzing yet, since there are no doubt still loader/depacker issues to sift through.

debrouxl:
Oh. I saw your message, but missed your previous force push performed on the next day, and then I saw today's force push. The computer could have spent a bit of CPU time on fuzzing in the meantime if I had noticed.

AliceLR:
I don't remember exactly what I force pushed 6 days ago (probably #409), but this time was #413.

debrouxl:
The `memopen` + `xmp_test_module_from_file` code makes the ASAN libfuzzer-based front-end over two orders of magnitude slower here... so I've disabled it for now :)

I suppose that I'll attempt to gather profiling information later.

AliceLR:
> The `memopen` + `xmp_test_module_from_file` code makes the ASAN libfuzzer-based front-end over two orders of magnitude slower here... so I've disabled it for now :)

Sounds like there might be some more slow load files affecting the depackers in your corpus—does reducing the timeout make it turn up anything interesting? It's probably OK to turn them off, though. I've had a few things come up for them on my end but aside from stb-vorbis (minor UB I don't want to touch yet) the most recent thing I had come up is an allocation bounding issue in the ARC/ArcFS depackers that doesn't look easily fixable.

debrouxl:
The exec rate was fairly constant over the past ~1.5K files processed by libxmp_fuzz_asan, i.e. the smallest ones, before I killed the program: 5-6 execs per second. Without `memopen` + `xmp_test_module_from_file`, these files are processed at a rate of over 1000 execs/s.

AliceLR:
New force-push is to fix a leak bug in the debugger function because I forgot to add a corresponding `fclose` for the `fmemopen`...

That drop in execs sounds roughly like what I was encountering with libmodplug with MMCMP and MIDI enabled (average of 270 execs/s would drop all the way to 10 execs/s before I finally just turned them off). It'd find a lot of slow load issues in those places that were likely nontrivial fixes.

debrouxl:
I don't know.

I have another opportunity to try the new version, as `libxmp_fuzz_asan` stopped itself because of a module whose processing takes > 10s on a Core i7-8750H. Here it is: [xmp_slow_module_20210712_01.tar.gz](https://github.com/libxmp/libxmp/files/6799614/xmp_slow_module_20210712_01.tar.gz) .

debrouxl:
It has occurred to me only now that since libfuzzer uses SHA-1 (...) for sample file names, I can use sha1sums on my sample files, then use the output to find and filter every file which prevents a restart. This yields 
[xmp_oom_module_20210712_02.tar.gz](https://github.com/libxmp/libxmp/files/6807461/xmp_oom_module_20210712_02.tar.gz)
which is redundant with the previous attachment: the allocated memory sizes and stack traces are the same.

debrouxl:
The run stopped less than 1M execs after the restart, finding 3 slow files in the process. Tw
[xmp_slow_modules_20210713_01.tar.gz](https://github.com/libxmp/libxmp/files/6807914/xmp_slow_modules_20210713_01.tar.gz)

The one which aborted the run is
```
artifact_prefix='./'; Test unit written to ./timeout-84394b3f0a0e762649597f7af82fa69a6b76da97
==47672== ERROR: libFuzzer: timeout after 10 seconds
    #0 0x52b1c1 in __sanitizer_print_stack_trace (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x52b1c1)
    #1 0x474288 in fuzzer::PrintStackTrace() (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x474288)
    #2 0x45a239 in fuzzer::Fuzzer::AlarmCallback() (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x45a239)
    #3 0x7effb168513f  (/lib/x86_64-linux-gnu/libpthread.so.0+0x1413f)
    #4 0x471862 in __sanitizer_cov_trace_const_cmp8 (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x471862)
    #5 0x561ca7 in mread .../xmp/libxmp/test-dev/.fuzzer/src/memio.c:44:7
    #6 0x55f510 in hio_read .../xmp/libxmp/test-dev/.fuzzer/src/hio.c:264:9
    #7 0x6fdb64 in read_typname .../xmp/libxmp/test-dev/.fuzzer/src/loaders/umx_load.c:225:3
    #8 0x6fdb64 in probe_umx .../xmp/libxmp/test-dev/.fuzzer/src/loaders/umx_load.c:274:6
    #9 0x6fdb64 in process_upkg .../xmp/libxmp/test-dev/.fuzzer/src/loaders/umx_load.c:343:9
    #10 0x6fc45a in umx_test .../xmp/libxmp/test-dev/.fuzzer/src/loaders/umx_load.c:351:9
    #11 0x55b539 in load_module .../xmp/libxmp/test-dev/.fuzzer/src/load.c:259:17
    #12 0x55c05f in xmp_load_module_from_memory .../xmp/libxmp/test-dev/.fuzzer/src/load.c:445:8
    #13 0x5521f8 in LLVMFuzzerTestOneInput .../xmp/libxmp/test-dev/libxmp_fuzz.c:15:6
    #14 0x45b821 in fuzzer::Fuzzer::ExecuteCallback(unsigned char const*, unsigned long) (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x45b821)
    #15 0x45af65 in fuzzer::Fuzzer::RunOne(unsigned char const*, unsigned long, bool, fuzzer::InputInfo*, bool*) (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x45af65)
    #16 0x45c990 in fuzzer::Fuzzer::MutateAndTestOne() (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x45c990)
    #17 0x45d405 in fuzzer::Fuzzer::Loop(std::__Fuzzer::vector<fuzzer::SizedFile, fuzzer::fuzzer_allocator<fuzzer::SizedFile> >&) (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x45d405)
    #18 0x44ce15 in fuzzer::FuzzerDriver(int*, char***, int (*)(unsigned char const*, unsigned long)) (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x44ce15)
    #19 0x474a62 in main (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x474a62)
    #20 0x7effb1360d09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #21 0x421639 in _start (.../xmp/libxmp/test-dev/libxmp_fuzz_asan+0x421639)
```
i.e. another module considered as UMX, like yesterday's slow module.

sezero:
It enters umx_load.c::read_typename() with a crazy idx value of 431196131

sezero:
If I apply the following, it bails on short read at i==643. Comments?
```diff
diff --git a/src/loaders/umx_load.c b/src/loaders/umx_load.c
index 0a591b3..ca58c78 100644
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -221,8 +221,8 @@ static int read_typname(HIO_HANDLE *f, const struct upkg_hdr *hdr,
 	if (idx >= hdr->name_count) return -1;
 	buf[63] = '\0';
 	for (i = 0, l = 0; i <= idx; i++) {
-		hio_seek(f, hdr->name_offset + l, SEEK_SET);
-		hio_read(buf, 1, 63, f);
+		if (hio_seek(f, hdr->name_offset + l, SEEK_SET) < 0) return -1;
+		if (hio_read(buf, 1, 63, f) < 63) return -1;
 		if (hdr->file_version >= 64) {
 			s = *(signed char *)buf; /* numchars *including* terminator */
 			if (s <= 0 || s > 64) return -1;
```

debrouxl:
Ignoring the return value of `hio_seek` and `hio_read` calls has been the cause of lots of similar issues.

AliceLR:
I got sidetracked by something else, but in the meantime my ASan fuzzer instance I left running in the background finally crashed on something in stb_vorbis. (Hopefully the update will fix it.) Will look at the OoM cases soon, but they probably belong in a new issue.

> If I apply the following, it bails on short read at i==643. Comments?

This patch looks OK to me.

sezero:
It's been a long time since I read the upkg format, but I _think_ short reads
i.e. less than 63, _should_ be OK (the string can be a small one) provided it's
a non-zero read? So, I tend to check the result of hio_read() te be non-zero
instead of less than 63. With this, it bails at i==648 instead of 643. Comments?
@AliceLR?
```diff
diff --git a/src/loaders/umx_load.c b/src/loaders/umx_load.c
index 0a591b3..f91bc9e 100644
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -221,8 +221,8 @@ static int read_typname(HIO_HANDLE *f, const struct upkg_hdr *hdr,
 	if (idx >= hdr->name_count) return -1;
 	buf[63] = '\0';
 	for (i = 0, l = 0; i <= idx; i++) {
-		hio_seek(f, hdr->name_offset + l, SEEK_SET);
-		hio_read(buf, 1, 63, f);
+		if (hio_seek(f, hdr->name_offset + l, SEEK_SET) < 0) return -1;
+		if (!hio_read(buf, 1, 63, f)) return -1;
 		if (hdr->file_version >= 64) {
 			s = *(signed char *)buf; /* numchars *including* terminator */
 			if (s <= 0 || s > 64) return -1;
```

AliceLR:
> It's been a long time since I read the upkg format, but I _think_ short reads
> i.e. less than 63, _should_ be OK (the string can be a small one) provided it's
> a non-zero read? So, I tend to check the result of hio_read() te be non-zero
> instead of less than 63.

I checked upkg.txt since I'm not familiar with the UMX format, but this (and the updated patch) is correct. The relevant section:
```
Older package versions (<64, original Unreal engine) store the Name
type as a zero-terminated ASCII string; "UT2k3", for example would
be stored as: "U" "T" "2" "k" "3" 0x00

Newer packages (>=64, UT engine) prepend the length of the string
plus the trailing zero. Again, "UT2k3" would be now stored as:
0x06 "U" "T" "2" "k" "3" 0x00
```

sezero:
I think I can do better, will post a new patch

sezero:
OK, here it is:  It rejects the file if the offset tables are past file end
or if they point to somewhere inside the file header, and in the case
of the offender in this bug report the rejection come very early.

OK to apply?
```diff
diff --git a/src/loaders/umx_load.c b/src/loaders/umx_load.c
index 0a591b3..3b4a24b 100644
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -219,10 +219,10 @@ static int read_typname(HIO_HANDLE *f, const struct upkg_hdr *hdr,
 	char buf[64];
 
 	if (idx >= hdr->name_count) return -1;
-	buf[63] = '\0';
+	memset(buf, 0, 64);
 	for (i = 0, l = 0; i <= idx; i++) {
-		hio_seek(f, hdr->name_offset + l, SEEK_SET);
-		hio_read(buf, 1, 63, f);
+		if (hio_seek(f, hdr->name_offset + l, SEEK_SET) < 0) return -1;
+		if (!hio_read(buf, 1, 63, f)) return -1;
 		if (hdr->file_version >= 64) {
 			s = *(signed char *)buf; /* numchars *including* terminator */
 			if (s <= 0 || s > 64) return -1;
@@ -248,6 +248,13 @@ static int probe_umx   (HIO_HANDLE *f, const struct upkg_hdr *hdr,
 	idx = 0;
 	fsiz = hio_size(f);
 
+	if (hdr->name_offset	>= fsiz ||
+	    hdr->export_offset	>= fsiz ||
+	    hdr->import_offset	>= fsiz) {
+		D_(D_INFO "UMX: Illegal values in header.\n");
+		return -1;
+	}
+
 	/* Find the offset and size of the first IT, S3M or XM
 	 * by parsing the exports table. The umx files should
 	 * have only one export. Kran32.umx from Unreal has two,
@@ -303,12 +310,12 @@ static int32 probe_header (HIO_HANDLE *f, struct upkg_hdr *hdr)
 		return -1;
 	}
 	if (hdr->name_count	< 0	||
-	    hdr->name_offset	< 0	||
 	    hdr->export_count	< 0	||
-	    hdr->export_offset	< 0	||
 	    hdr->import_count	< 0	||
-	    hdr->import_offset	< 0	) {
-		D_(D_INFO "UMX: Negative values in header\n");
+	    hdr->name_offset	< 36	||
+	    hdr->export_offset	< 36	||
+	    hdr->import_offset	< 36) {
+		D_(D_INFO "UMX: Illegal values in header.\n");
 		return -1;
 	}
 
```

sezero:
In addition, a sanity check in there was off-by-one:
```diff
diff --git a/src/loaders/umx_load.c b/src/loaders/umx_load.c
index 0a591b3..a576f57 100644
--- a/src/loaders/umx_load.c
+++ b/src/loaders/umx_load.c
@@ -225,7 +225,7 @@ static int read_typname(HIO_HANDLE *f, const struct upkg_hdr *hdr,
 		hio_read(buf, 1, 63, f);
 		if (hdr->file_version >= 64) {
 			s = *(signed char *)buf; /* numchars *including* terminator */
-			if (s <= 0 || s > 64) return -1;
+			if (s <= 0 || s > 63) return -1;
 			l += s + 5;	/* 1 for buf[0], 4 for int32 name_flags */
 		} else {
 			l += (long)strlen(buf);
```

sezero:
Actually, the upper bounds check for `s` in there seems unnecessary..

AliceLR:
The revised patch seems OK to me.

sezero:
umx patch applied. (BTW, you may want to cherry-pick https://github.com/sezero/libmodplug/commit/218f6d0fa2b2960247511cb29fb3e64e20090a0d into your libmodplug pull request)

AliceLR:
Force pushed for the UMX patch and stb-vorbis 1.22.

AliceLR:
Force pushed for #421.

AliceLR:
Force pushed to get some more misc. patches, including a heap corruption bug in the Coconizer loader I encountered while fuzzing for the new LZW decoder.

AliceLR:
Force pushed again to update with fixes for the ARC issues reported here and for some unrelated MSan issues I found while merging my corpus. I also updated `libxmp_fuzz.sh` to merge with all three fuzzers when the `merge` command is used, since this seems to have picked up quite a few files the ASan merge missed.

AliceLR:
...and rebased this time to get the `stat()` warning workaround for MSan.

AliceLR:
Rebased to get the Digital Symphony LZW decoder.

debrouxl:
* this libfuzzer front-end derived and improved from my minimal one in #302 has already pointed a significant number of issues, most of which were fixed, though there are still some ubsan and msan errors;
* the integration into the build system doesn't look weird to me: other projects have special fuzzing targets which are enabled in dev mode;
*  I feel bad for you having to rebase this branch quite often.

=> if one of the other reviewers agrees, I think that it should be committed in its current state, as a pretty good first step, which can always be built upon later :)

Saga Musix's suggestion of calling several more API functions, which you already partially implemented on your side for setting the row, remains very useful. If the impact on the run time of seek / instrument-sample info / render calls proves to be too significant, some of it can be gated by e.g. `getenv`.
Amusingly, as part of my hearing tests for #404 , I've just stumbled across a reproducible OOB read retrieving the combined samples / instruments list (i command) on a module from ModArchive, which I downloaded a while ago because it has good rating, I presume: https://modarchive.org/index.php?request=view_by_moduleid&query=66343 . That's the real trigger for writing this entry.

```
==2054807==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x621000031e68 at pc 0x0000004cd2a9 bp 0x7ffc36f6a700 sp 0x7ffc36f6a6f8
READ of size 4 at 0x621000031e68 thread T0
    #0 0x4cd2a8 in info_ins_smp .../xmp/xmp-cli/src/info.c:258:10
    #1 0x4cfd5e in main .../xmp/xmp-cli/src/main.c:554:7
    #2 0x7fb2a00a0d09 in __libc_start_main csu/../csu/libc-start.c:308:16
    #3 0x420679 in _start (.../bin/xmp+0x420679)

0x621000031e68 is located 32 bytes to the right of 4424-byte region [0x621000030d00,0x621000031e48)
allocated by thread T0 here:
    #1 0x7fb2a054c0cb in libxmp_init_instrument .../xmp/libxmp/src/loaders/common.c:53:36

SUMMARY: AddressSanitizer: heap-buffer-overflow .../xmp/xmp-cli/src/info.c:258:10 in info_ins_smp
```

sezero:
That beek-subs-something.it, yes? Interestingly, valgrind doesn't give me any errors with or without 4044 applied.

AliceLR:
> though there are still some ubsan and msan errors;

Which ones? I'm aware of a few OXM/Vorbis-related minor issues I haven't touched yet, but any others should probably have an issue.

> * I feel bad for you having to rebase this branch quite often.
> 
> => if one of the other reviewers agrees, I think that it should be committed in its current state, as a pretty good first step, which can always be built upon later :)

It isn't bothering me to rebase this branch, but if the constant pings are an issue I'll get this ready for review (just need to fix the stale `.fuzzers/` issue, probably just by moving that part to a phony target).

> Saga Musix's suggestion of calling several more API functions, which you already partially implemented on your side for setting the row

I didn't, actually—the `xmp_set_row` bug was reported downstream by a MegaZeux user.

> Amusingly, as part of my hearing tests for #404 , I've just stumbled across a reproducible OOB read retrieving the combined samples / instruments list (i command) on a module from ModArchive

Will look into this, thanks.

AliceLR:
> (just need to fix the stale `.fuzzer/` issue, probably just by moving that part to a phony target).

Yep, that seems to have been sufficient. Marking this ready for review.

.gitignore
```diff
@@ -15,8 +15,11 @@
 *.cov
 a.out
 lib/
+test-dev/gen_mixer_data
+test-dev/gen_module_data
 test-dev/libxmp-covertest
-test-dev/libxmp-test
+test-dev/libxmp-tests
```
sezero, line 21: While you're at it, change `test/libxmp-tests`to `test/libxmp-test` an remove non-existent `test/test8.raw`

Go ahead and apply with that change.

sezero (approved):

.gitignore
AliceLR: Will do. I was under the impression another pull request did address that, but I can't find it.

AliceLR:
> Amusingly, as part of my hearing tests for #404 , I've just stumbled across a reproducible OOB read retrieving the combined samples / instruments list (i command) on a module from ModArchive

Looks like the issue here is this module has numerous instruments that reference nonexistent samples: each instrument seems to reference exactly one sample with the same number as the instrument, but the last sample is 0x4f and the instruments continue to 0x5d.

My take on this is that these invalid sample numbers accurately reflect the provided input data and libxmp has internal mechanisms specifically to handle this edge case, so this is xmp-cli's fault for not properly checking subinstrument sample IDs against the module's sample count. (This could also be interpreted as libxmp's fault for exposing `xmp_module` to the API in the first place.)

PR #389, opened 2021-06-24
Remove 256 bytes requirement for modules in some API functions.

AliceLR:
The functions `xmp_load_module`, `xmp_test_module`, and `xmp_test_module_from_file` all have a strange requirement that files passed to them need to be 256 bytes or larger. This size requirement has been commented on before due to it unusually applying only with depackers and/or non-lite libxmp.

This requirement is also entirely baseless: here's a GDM module I made containing both a valid sample and valid pattern that's only 242 bytes long (nevermind I had to massage the pattern in a hex editor...). It only loads when using memory/file/callbacks loading functions because of this check. [small.gdm.zip](https://github.com/libxmp/libxmp/files/6707555/small.gdm.zip)

Smaller is probably also possible with the correct format choice, or by not having a pattern or sample at all. I think the original Schism date test S3M and IT were both under the limit prior to being modified to bypass this check. I think this makes the point well enough, though. This file has been added as a test input to the `xmp_load_module` and `xmp_test_module*` API function tests.


sezero (approved):

sezero:
Looks OK to me,

@cmatsuoka: Is this OK?

sezero:
Alice: rebase and force-push this so that the buildbots run with it.

AliceLR:
Done. Any idea how to get rid of the Travis-CI check from the list?

sezero:
I can only guess that we have travis enabled in project settings and it needs to be removed.
@cmatsuoka, correct?

Issue #400, opened 2021-06-26
Usage of integer division.


AliceLR:
Several places in libxmp use integer division, which is slow and very unfriendly to platforms without a dedicated integer division instruction like ARMv6 (and are still slow even if there is one). Unlike libxmp's floating point usage, this seems to have creeped into several inner loops. Note this only applies to non-literal divisors (literal divisors get optimized to a MUL+SHR).

Roughly in decreasing order of urgency:

* `mixer.c` uses an integer division once per sample frame for declicking(!). This only affects small portions of samples but should be fixed. (#404)
* The following loaders use integer division for pattern events instead of two for loops for some reason:
  * `mod_load.c` (#457)
  * `st_load.c` (#609)
  * `xm_load.c` (#461)
  * ~~`abk_load.c`~~ effect conversion
  * `asylum_load.c` (#501)
  * `coco_load.c` (#443)
  * `far_load.c` (#500)
  * ~~`flt_load.c`~~ AM instrument calculation
  * `fnk_load.c` (#446)
  * `okt_load.c`
  * `sfx_load.c`
  * `stm_load.c` (#482)
* `player.c` uses these when computing envelopes (probably reworkable) and when computing the current tick (can be manually unswitched so it's guaranteed to only be performed once per tick).
* `mix_paula.c`
* Misc. depackers.

I found these by searching a disassembly of libxmp.dll for any places `idiv*` was emitted, which may not be foolproof.

AliceLR:
The most promising replacements for the declicking loop seem to be:

(Requires hardware floating point.)
```C++
static void floating_point(int32_t *buf, int count)
{
        int32_t left, right;
        get_smpl(left, right);

        float max_x2 = (float)count * (float)count;

        while(count--)
        {
                float coef = (float)count * (float)count / max_x2;
                *buf++ += (int32_t)(left  * coef);
                *buf++ += (int32_t)(right * coef);
        }
}
```
```C++
static void fixed_point(int32_t *buf, int count)
{
        int32_t left, right;
        get_smpl(left, right);

        static constexpr int FPSHIFT = 10;
        int tickval = (1 << FPSHIFT) / count;
        if(tickval < 1)
                tickval = 1;

        int tickmul = tickval * count;
#define DO_MULT(x) (tickmul * ((x) >> FPSHIFT))

        while((tickmul -= tickval))
        {
                *buf++ += DO_MULT(DO_MULT(left));
                *buf++ += DO_MULT(DO_MULT(right));
[...]

Issue #390, opened 2021-06-24
More errors flagged by MemorySanitizer & UndefinedBehaviourSanitizer...


debrouxl:
Merging my local corpus of many broken files with the msan-instrumented libfuzzer front-end from #383 finds several more stack traces which produce UMRs, and are not fixed by today's #388 . Here's the small tarball of redundant testcases: [libxmp_msan_crashes_20210624_01.tar.gz](https://github.com/libxmp/libxmp/files/6709639/libxmp_msan_crashes_20210624_01.tar.gz)

Sorry, I've erased the log which used to contain the stack traces produced by msan instrumentation during the lengthy merge process, by starting another job in the same path... namely, a ubsan-instrumented libfuzzer front-end. It has produced the following output:
```
src/loaders/iff.c:192:7: runtime error: member access within null pointer of type 'struct iff_info'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/iff.c:192:7 in 
src/loaders/iff.c:46:7: runtime error: member access within null pointer of type 'struct iff_info'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/iff.c:46:7 in 
src/loaders/prowizard/di.c:204:15: runtime error: signed integer overflow: 196864 - -2147417592 cannot be represented in type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/prowizard/di.c:204:15 in 
src/loaders/prowizard/p61a.c:221:13: runtime error: index 260 out of bounds for type 'uint8 [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/prowizard/p61a.c:221:13 in 
src/period.c:263:6: runtime error: -nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/period.c:263:6 in 
src/loaders/prowizard/p61a.c:338:21: runtime error: index 260 out of bounds for type 'uint8 [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/prowizard/p61a.c:338:21 in 
src/loaders/prowizard/p61a.c:182:13: runtime error: index 260 out of bounds for type 'uint8 [256]'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/loaders/prowizard/p61a.c:182:13 in 
[...]

PR #404, opened 2021-07-01
Improve anticlick performance and fix anticlick filter volume bug.

AliceLR:
This fixes two bugs in the anticlick routine found in #400:

* The existing anticlick routine was slow due to relying on repeated integer division and, for large tick sizes, could truncate very large portions of the source sample. It has been replaced by a (much) faster fixed point variant of the same algorithm that outputs a more accurate curve.
* @debrouxl pointed out that the module "The Hero Within" by @sagamusix has click issues and would be worth testing the updated algorithm against. I diagnosed the remaining clicks and found that they're caused by behavior in the filter mixers where the previous sample values would be stored after volume was applied, almost entirely negating the anticlick volume ramp. Volume is now applied after the filter calculation, entirely fixing the popping issue (as far as I can tell) in this module.

I also added some deadcode to the mixer tests to make it easier to regenerate their output.

(edit: probably worth noting that this is still currently being tested against existing modules in the aforementioned issue.)


AliceLR:
Oops, need to fix the regular test too. Will push in a bit.

AliceLR:
Here's a small set of modules (plus The Hero Within) I've found from my local ModLand copy with noticeable filter usage that may cause clicks in master but doesn't in this branch. It doesn't seem that there are actually that many modules using filters that have that playback issue, maybe because rapid notes with filters tend to use more percussive sounds that mask the click. The Hero Within is still the most obvious one with this problem I've heard so far.

[it_filter_anticlick.zip](https://github.com/libxmp/libxmp/files/6751686/it_filter_anticlick.zip)

sezero:
@cmatsuoka, what's the verdict on this?

AliceLR:
FWIW I don't think merging this needs to be rushed. I'm fairly confident at this point that both of these changes are correct but it's not like e.g. a critical build system bug or vulnerability. (It doesn't help that fixing the mixer tests blew up the size of the diff to >7k lines, despite the actual part of the diff worth reviewing only being a couple dozen at best.)

If you want to cut a release for 4.5.1 I think it would be OK to do that and delay merging this (and further patches for #390, #400, other playback bugs) to 4.5.2. libxmp uses semantic versioning as far as I can tell so there can be as many release versions as we need. :)

AliceLR:
(Nothing new here, just rebased.)

debrouxl:
On my side, the hearing test on ~65 modules by Saga Musix, as well as several other ones, is alright: no noticeable new clicks, or reduction in noticeable clicks outside of The Hero Within.
FWIW, of course: it suffices that the relevant modules don't use the sample / effect / etc. conditions impacted by this change for no impact to be heard...

AliceLR:
I forgot to post it here, but I did manage to reproduce this bug in a custom module, and confirmed this branch does remove the pops here as well: [it_filter_pop.it.zip](https://github.com/libxmp/libxmp/files/6772944/it_filter_pop.it.zip)

sezero (approved):

sezero:
How much more testing does this one need?

debrouxl:
I haven't listened to my whole usual set of modules, but at this rate, it's going to take a while... and anyway, this hearing test has no value if none of these modules happens to use the sample / effect / note conditions where a change in the anticlick algorithm produces an audible change in output.

AliceLR:
From my testing I was fairly satisfied that this adequately fixed the filter bug without breaking anything, but I don't have a library of ITs I regularly listen to to compare.

sezero:
What to do with this then? @cmatsuoka is unresponsive these days..

NoSuck:
I tried to test these changes, but `make` failed (even after `autoconf && ./configure --prefix=/usr && make clean`):

```
make: *** No rule to make target 'src/xfnmatch.h', needed by 'src/load_helpers.lo'.  Stop.
```

sezero:
You are running with an outdated libxmp source tree

AliceLR:
> I tried to test these changes, but `make` failed (even after `autoconf && ./configure --prefix=/usr && make clean`):
> 
> ```
> make: *** No rule to make target 'src/xfnmatch.h', needed by 'src/load_helpers.lo'.  Stop.
> ```

Try a `make distclean`—I don't think regular `make clean` removes `depend` currently. (xfnmatch was moved to src/depackers/ since it's only used by those.)

AliceLR:
Rebased to master since the `-Wunused-result` fixes conflicted with the mixer data generation deadcode this patch adds.

debrouxl:
FWIW, I've listened to several hundred more files, without hearing ill effects caused by this patch. The same caveats still apply. More modules to listen to...

sezero:
I'm tempted to apply this patch

debrouxl:
No obvious change on 52 modules from Awesome, either.

sezero:
Let's merge this ?

sezero:
PING?

debrouxl:
No objection from me, but I'm not a reviewer anyway.

AliceLR:
I'm fine with merging this. I don't think it's actually a big enough change for us to keep bugging Claudio about it, and all testing so far points to this patch working correctly.

sezero:
OK, merging.

PR #406, opened 2021-07-03
Fix UndefinedBehaviorSanitizer errors found in existing regression tests.

AliceLR:
This fixes errors found by UBSan in \*existing\* test files as discussed in #390 (`-fsanitize=undefined -fno-sanitize=shift-base`). This patch will allow a UBSan regression test to be added to the existing sanitizer regression tests.

This probably does NOT fix anywhere near most of the UBSan errors reported by @debrouxl in #390, so expect more of these patches.

* The hand-rolled `offsetof` implementation in list.h relied on undefined behavior and has been replaced with actual `offsetof`.
* The Player 6.1a loader was missing a lot of sanity checks on event repeats, some of which flagged UBSan.
* The BZip2 depacker relied on a hack to implement 1 indexing that UBSan (and I) did not like.
* `libxmp_c2spd_to_note` could attempt to convert `-nan` to an integer due to negative `c2spd` values not being filtered out.
* inflate.c's `reverse_bitstream` could sometimes attempt to shift an `unsigned int` by 32.
* Missing MDL instrument loop sanity checks could lead to `int` overflows.
* Missing The Player 4.0a loop start sanity checks could lead to `int` overflows.
* The LZW depacker could left shift by an arbitrary value up to 255.
* The LZW depacker could attempt to square 65535 in an `int` instead of using an `unsigned int`.


AliceLR:
This seems to have broken coverage for a check in `pw_load.c` even though I modified the existing test so it wouldn't... will investigate.

sezero (approved):
Looks good to me in general.  Do you have follow-up patches, or is this
considered finished?

Is the bzunzip change tested? (we have a bz2 test under test-dev, yes?)
Searched the bunzip2 version we use, it looks like the one in linux kernel,
I guess they don't concern themselves with crafted data because the bz2
code in there is for decompressing the kernel. There was one fix there we
_might_ be interested in, though:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/lib/decompress_bunzip2.c?id=b5c8afe5be51078a979d86ae5ae78c4ac948063d

AliceLR:
> Looks good to me in general. Do you have follow-up patches, or is this
> considered finished?

This particular patch is finished.

> Is the bzunzip change tested? (we have a bz2 test under test-dev, yes?)

There is a bz2 test as well as a fuzzer test for it (for a total coverage of 94.9% apparently). I haven't personally tried any other files. (My changes should also be exactly equivalent to what existed before...)

> Searched the bunzip2 version we use, it looks like the one in linux kernel,
> I guess they don't concern themselves with crafted data because the bz2
> code in there is for decompressing the kernel. There was one fix there we
> _might_ be interested in, though:
> https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/lib/decompress_bunzip2.c?id=b5c8afe5be51078a979d86ae5ae78c4ac948063d

I can't tell if this actually would affect things that much considering there's a later check that does use `>=`, but for a different variable.

sezero:
OK then, merging.

PR #411, opened 2021-07-06
period.c: always use libxmp_round()

sezero:
we already do that for certain primary targets such as msvc,
and our implementation is pretty much universal.

Also updated the configury, and also Android.mk which was missing
the `HAVE_POWF` define, thus was using `pow` instead.

Speaking of `pow` vs `powf`, I had a mild intention of using `pow` in
filter.c, but Claudio said it possibly needs computing in floats: it
certainly needs cleanup to remove double constants use and speed-check.


AliceLR (approved):
This looks OK to me.

PR #412, opened 2021-07-07
update configuration for libm dependency:

sezero:
- configure.ac: updated list from libtool to filter systems that don't
  have libm, or don't need it.
- libxmp.pc.in: add -lm to Libs.private only if present (or needed)


AliceLR (approved):
This looks good to me.

PR #413, opened 2021-07-07
Fix undefined behavior found by libFuzzer/UBSan, part 2.

AliceLR:
Fixes more undefined behavior from #390, as well as other issues I found. This still isn't everything reported in that issue (I skipped a bunch of ProWizard stuff for now).

Loaders:
* The MED4 loader could cause signed overflows in a couple of places where offsets didn't really need to be added in the first place. These have been reworked to relative seeks.
* The OctaMED loaders both contained some really sketchy code that seeks to the sample data, reads two variables, jumps around to two other arbitrary places in the file, and then seeks back to finish reading the sample data. As part of the offset calculations for this, signed overflows could occur. I suspect this is *not* how the `InstrExt` and `InstrInfo` arrays are intended to be loaded, so I refactored these to load both arrays in one go and use relative seeks. (edit: also refactored the loading of the `blockarr` and `smplarr` values to behave the same.)
* The XM loader was missing a check on its instrument header size fields to prevent signed underflows. The instrument/sample header size defines also had the wrong values (sample header size is part of the sample header), so I corrected them.

Depackers:
* The BZ2 depacker contained a bit hack that used `int` instead of `unsigned int`, which made UBSan unhappy.
* Fix UB in the OXM pattern loader caused by very large pattern header lengths.
* In some very invalid cases (that I wasn't able to pin down), the LZW depacker was capable of generating codes where the previous code was set but the first code ID was `LZW_UNUSED`. The place this issue was occurring has been updated to check both before trying to use the (invalid) first code ID to index the chars array.
* The S404 depacker's efficiency value was used to derive shift exponents with no bounds checks. I found another implementation of a depacker for this format elsewhere that had stricter checks than I added, but I'm not familiar enough with it to make a call there.
* The LHA depacker could read too many `c_len` values in some invalid cases.

~~NOTE: not ready for merge yet, the OctaMED changes need tests. I'm suspicious that some of the existing OctaMED fuzzer modules aren't actually covering the checks they were originally supposed to, and I'll also most likely need to construct inputs that cause the UB prior to the new patch by hand. Since codecov's branch reports are reliable (unlike the PR summaries for some reason), I want to let it do the work figuring that out for me. : )~~ More or less resolved.

Also, for a less garbage diff to review the OctaMED loader changes, I found `git diff --histogram` does a good job with this patch locally.


AliceLR:
OK, the OctaMED loaders have several more tests now and I'm convinced I didn't break anything in them. More review-friendly diff of the OctaMED loader changes here: https://gist.github.com/AliceLR/233fedf92c9a9fbfe127dc3d26943478

sezero:
I'll be looking at this today.

src/loaders/med4_load.c
```diff
@@ -546,9 +546,13 @@ static int med4_load(struct module_data *m, HIO_HANDLE *f, const int start)
 			} else if (_type == -1) {
 				hio_seek(f, 20, SEEK_CUR);
 				num_smp += hio_read16b(f);
+				_len -= 22;
 			}
 
-			hio_seek(f, _pos + _len, SEEK_SET);
```
sezero, line 551: What are we doing with the `_pos` variable we read above now?

sezero:
> OK, the OctaMED loaders have several more tests now and I'm convinced I didn't break anything in them.

Valid files are still read properly too, yes?

sezero:
>  More review-friendly diff of the OctaMED loader changes here: https://gist.github.com/AliceLR/233fedf92c9a9fbfe127dc3d26943478

I'm having a hard time following it, but looked OK at first glance.
If you don't want more eyes, I can mark this as OK.

src/loaders/med4_load.c
AliceLR: Good catch, that variable isn't needed anymore. I'll push a patch to remove it soon.

AliceLR:
> Valid files are still read properly too, yes?

The individual files I've tested so far load and play the same as before, yes. I can do a `xmp --load-only` test on my ModLand MED folders to make sure nothing new fails, but the changes in the OctaMED loaders should be exactly equivalent to what existed before, just with fewer seeks.

> > More review-friendly diff of the OctaMED loader changes here: https://gist.github.com/AliceLR/233fedf92c9a9fbfe127dc3d26943478
> 
> I'm having a hard time following it, but looked OK at first glance.
> If you don't want more eyes, I can mark this as OK.

Will reply back re: this.

AliceLR:
I confirmed that for my MED folders (MMD0 through MMD3), the output of `xmp --load-only` for master and for this branch is identical, meaning this doesn't break any existing modules to the point they would load differently (error, different pattern/instrument/sample counts, etc). I'll give the OctaMED patch one last look myself, but I think this might be good enough to not require extra eyes.

sezero:
OK, tell me when you're ready and we apply this.

AliceLR:
OK, finished giving it a final check. I found some debug fflush usage I didn't catch before but the intentional changes look correct to me.

sezero (approved):

PR #416, opened 2021-07-11
Allow loading MED2 samples from the module directory.

### LINKED ISSUE #414: Weird behavior of the xmp_loader_med2 unit test
While I working on the CMake build, I found that one of the unit tests has a weird behavior:

When I try to run the existing unit test and the Automake-built library, it reports the next:
```
med2_load [src/loaders/med2_load.c:221] [ 0] med2test_PanFlute                26ac 0000 0000   V40
```

But, when I try to run the player (`examples/player-sdl.c`),
```bash
LD_LIBRARY_PATH=../lib ./player-sdl '/home/vitaly/_git_repos/libxmp/test-dev/data/m/med2test.med'
```

it gives me the next result:
```
med2_load [src/loaders/med2_load.c:221] [ 0] med2test_PanFlute                0000 0000 0000   V40
```

And, the absolutely same case happens at unit tests when I build them by my CMake build, and this file gets failed because of zero sub-instruments.

Can you explain what happens with that file? Where is the error?

At the file, I do see the big block of zeroes. I didn't familiar with MED2 files, and therefore I can't tell, is this file valid or it's a bad file?
изображение

AliceLR:
This is a bandaid fix for the MED2 test failing for some setups: when the instrument path doesn't work (usually due to being relative), the MED2 loader now allows loading samples from the current module directory. I don't think this is how MED2 is supposed to work but IMO it's a reasonable way to distribute MED2 modules and it's how other song files are handled by libxmp.

Should fix #414, but we'll see.

### DIFF
--- a/src/loaders/med2_load.c
+++ b/src/loaders/med2_load.c
@@ -81,7 +81,7 @@ int med2_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		if (hio_read(buf, 1, 40, f) != 40)
 			return -1;
 
-		libxmp_instrument_name(mod, i, buf, 32);
+		libxmp_instrument_name(mod, i, buf, 40);
 		if (libxmp_alloc_subinstrument(mod, i, 1) < 0)
 			return -1;
 	}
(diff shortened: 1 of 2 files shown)


Wohlstand:
This thing fixes the example player, however, the library built by me still fails, now I investigating the reason at my side :thinking:

Wohlstand:
Okay, the reason for CMake failure was the missing of the `HAVE_DIRENT` check that I had to add, and that resolved the true reason - the `libxmp_check_filename_case()` call had always failed, which caused the inability to find an instrument. So, gonna to add other missing checks, I took them from off the Makefile, gonna fill my CMake build, but current stuff one I pushed

Wohlstand:
So, I did the necessary fix at CMake build, and now med2 unit tests get passed, so, your patch fixes the player behavior :fox_face:  but unit test fix is my fault

Wohlstand (approved):
I tested this, and this works, example player now works fine! :fox_face: :+1:

sezero (approved):

PR #419, opened 2021-07-14
Make libxmp-lite valid C++ (also minor fixes)

Clownacy:
Hello! I maintain a shallow fork of libxmp-lite as a built-in library of a project of mine.

In order to compile properly in my project, my fork has had several tweaks and fixes applied to it. Most notably, it is able to compile as valid C++. I figured it would benefit both of us if these modifications were upstreamed: that way, the codebase is made more robust and portable, and I don't have to reapply so many modifications whenever I update my project's copy of the library.

Unfortunately, these modifications _only_ extend to libxmp-lite: my fork isn't based on the regular libxmp, so these changes won't make the entire libxmp library valid C++ nor will it address all of libxmp's uses of absolute include paths.

Hopefully someone more familiar with libxmp than me can apply my modifications to the rest of the codebase. If not, then at least libxmp-lite can benefit from this, even if libxmp doesn't.

I recommend reviewing each commit individually instead of using the 'Files changed' tab: viewing all of the changes at once is pretty messy.

### DIFF
--- a/src/control.c
+++ b/src/control.c
@@ -310,6 +310,9 @@ int xmp_channel_vol(xmp_context opaque, int chn, int vol)
 }
 
 #ifdef USE_VERSIONED_SYMBOLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 LIBXMP_EXPORT_VERSIONED extern int xmp_set_player_v40__(xmp_context, int, int) LIBXMP_ATTRIB_SYMVER("xmp_set_player@XMP_4.0");
 LIBXMP_EXPORT_VERSIONED extern int xmp_set_player_v41__(xmp_context, int, int)
 			__attribute__((alias("xmp_set_player_v40__"))) LIBXMP_ATTRIB_SYMVER("xmp_set_player@XMP_4.1");
@@ -323,6 +326,10 @@ asm(".symver xmp_set_player_v40__, xmp_set_player@XMP_4.0");
 asm(".symver xmp_set_player_v41__, xmp_set_player@XMP_4.1");
 asm(".symver xmp_set_player_v43__, xmp_set_player@XMP_4.3");
 asm(".symver xmp_set_player_v44__, xmp_set_player@@XMP_4.4");
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #define xmp_set_player__ xmp_set_player_v40__
@@ -430,6 +437,9 @@ int xmp_set_player__(xmp_context opaque, int parm, int val)
 }
 
 #ifdef USE_VERSIONED_SYMBOLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 LIBXMP_EXPORT_VERSIONED extern int xmp_get_player_v40__(xmp_context, int) LIBXMP_ATTRIB_SYMVER("xmp_get_player@XMP_4.0");
 LIBXMP_EXPORT_VERSIONED extern int xmp_get_player_v41__(xmp_context, int)
 		__attribute__((alias("xmp_get_player_v40__"))) LIBXMP_ATTRIB_SYMVER("xmp_get_player@XMP_4.1");
@@ -446,6 +456,10 @@ asm(".symver xmp_get_player_v41__, xmp_get_player@XMP_4.1");
 asm(".symver xmp_get_player_v42__, xmp_get_player@XMP_4.2");
 asm(".symver xmp_get_player_v43__, xmp_get_player@XMP_4.3");
 asm(".symver xmp_get_player_v44__, xmp_get_player@@XMP_4.4");
+#endif
+
+#ifdef __cplusplus
+}
 #endif
 
 #define xmp_get_player__ xmp_get_player_v40__
--- a/src/control.c
+++ b/src/control.c
@@ -31,7 +31,7 @@ xmp_context xmp_create_context(void)
 {
 	struct context_data *ctx;
 
-	ctx = calloc(1, sizeof(struct context_data));
+	ctx = (struct context_data*)calloc(1, sizeof(struct context_data));
 	if (ctx == NULL) {
 		return NULL;
 	}
--- a/src/control.c
+++ b/src/control.c
@@ -31,7 +31,7 @@ xmp_context xmp_create_context(void)
 {
 	struct context_data *ctx;
 
-	ctx = (struct context_data*)calloc(1, sizeof(struct context_data));
+	ctx = (struct context_data *)calloc(1, sizeof(struct context_data));
 	if (ctx == NULL) {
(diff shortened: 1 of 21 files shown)


sezero:
I'm pretty OK with this.

The only suspicion I had was with the `D_` variadic macros:
I _think_ the original intention _probably_ was supporting
ancient gcc versions where `__VA_ARGS__` wasn't there yet.
I usually have definitions around for that in my projects:
```
#if defined(__GNUC__) && !(defined(__STDC_VERSION__) && __STDC_VERSION__ >= 199901L)
#define Sys_Printf(fmt, args...)	CON_Printf(_PRINT_TERMONLY, fmt, ##args)
#else
#define Sys_Printf(...)			CON_Printf(_PRINT_TERMONLY, __VA_ARGS__)
#endif
```
However, I found an old gcc295 builds around and tried your
patch, it surprisingly worked. So I guess it really is OK and
workarounds like above or libxmp has probably are for even
older gcc.

With that out of the way, I guess you can remove the watcom-
specific part from there because it just follows the standart
the same way the new version just below it.

One thing failing is this - on linux with versioned symbols
support:
```
src/control.c:314: error: ‘int xmp_set_player_v41__(char*, int, int)’ aliased to undefined symbol ‘xmp_set_player_v40__’
src/control.c:316: error: ‘int xmp_set_player_v43__(char*, int, int)’ aliased to undefined symbol ‘xmp_set_player_v40__’
src/control.c:318: error: ‘int xmp_set_player_v44__(char*, int, int)’ aliased to undefined symbol ‘xmp_set_player_v40__’
src/control.c:434: error: ‘int xmp_get_player_v41__(char*, int)’ aliased to undefined symbol ‘xmp_get_player_v40__’
src/control.c:436: error: ‘int xmp_get_player_v42__(char*, int)’ aliased to undefined symbol ‘xmp_get_player_v40__’
src/control.c:438: error: ‘int xmp_get_player_v43__(char*, int)’ aliased to undefined symbol ‘xmp_get_player_v40__’
src/control.c:440: error: ‘int xmp_get_player_v44__(char*, int)’ aliased to undefined symbol ‘xmp_get_player_v40__’
make: *** [src/control.lo] Error 1
```

Now, how to get anound that...

Off-topic:  I also noticed that the lite version is not including
src/win32/debug.c, not in makefiles nor in release-packaging, ugh..

sezero:
The versioned syms failure is cured by this:
```diff
diff --git a/src/control.c b/src/control.c
--- a/src/control.c
+++ b/src/control.c
@@ -310,6 +310,9 @@
 }
 
 #ifdef USE_VERSIONED_SYMBOLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 LIBXMP_EXPORT_VERSIONED extern int xmp_set_player_v40__(xmp_context, int, int) LIBXMP_ATTRIB_SYMVER("xmp_set_player@XMP_4.0");
 LIBXMP_EXPORT_VERSIONED extern int xmp_set_player_v41__(xmp_context, int, int)
 			__attribute__((alias("xmp_set_player_v40__"))) LIBXMP_ATTRIB_SYMVER("xmp_set_player@XMP_4.1");
@@ -325,6 +328,10 @@
 asm(".symver xmp_set_player_v44__, xmp_set_player@@XMP_4.4");
 #endif
 
+#ifdef __cplusplus
+}
+#endif
+
 #define xmp_set_player__ xmp_set_player_v40__
 #else
 #define xmp_set_player__ xmp_set_player
@@ -430,6 +437,9 @@
 }
 
 #ifdef USE_VERSIONED_SYMBOLS
+#ifdef __cplusplus
+extern "C" {
+#endif
 LIBXMP_EXPORT_VERSIONED extern int xmp_get_player_v40__(xmp_context, int) LIBXMP_ATTRIB_SYMVER("xmp_get_player@XMP_4.0");
 LIBXMP_EXPORT_VERSIONED extern int xmp_get_player_v41__(xmp_context, int)
 		__attribute__((alias("xmp_get_player_v40__"))) LIBXMP_ATTRIB_SYMVER("xmp_get_player@XMP_4.1");
@@ -448,6 +458,10 @@
 asm(".symver xmp_get_player_v44__, xmp_get_player@@XMP_4.4");
 #endif
 
+#ifdef __cplusplus
+}
+#endif
+
 #define xmp_get_player__ xmp_get_player_v40__
 #else
 #define xmp_get_player__ xmp_get_player
```

sezero:
@AliceLR, @cmatsuoka: What do you say?

src/control.c
```diff
@@ -31,7 +31,7 @@ xmp_context xmp_create_context(void)
 {
 	struct context_data *ctx;
 
-	ctx = calloc(1, sizeof(struct context_data));
+	ctx = (struct context_data*)calloc(1, sizeof(struct context_data));
```
AliceLR, line 34: Minor style complaint, but libxmp pretty consistently spaces the left side of type `*`s from what I've seen (e.g. `struct context_data *`).

AliceLR:
So far the parts of this patch I've reviewed—fixed variadic macros, header guards, local includes, fixed global declarations, malloc casts—look good to me (aside from the cast style nitpick). I'll leave versioned symbols to someone more familiar with them than I am...

sezero:
>  I'll leave versioned symbols to someone more familiar with them than I am...

Oh, that part is easy: c++ name mangling was in the way, and `extern "C" { }`
stuff fixed that.

debrouxl:
The changes look sane to me as well, FWIW.
I added the same kind of casts on the return value of `malloc`-type functions in a project whose implementation I nearly entirely converted to C++ in order to benefit from the stricter C++ typing, keeping an external C API for compatibility.

src/control.c
Clownacy: Sorry about that - I should have checked what style the project uses.

sezero (approved):

sezero:
OK, do you plan to extend this to the full library or should we apply?

Clownacy:
I wasn't planning on it, but if I ever do, I'll make another PR.

debrouxl:
GCC / Clang `-Wc++-compat` makes it easy to detect 'implicit conversion from "void *" to a pointer to non-"void" type', but playing whack-a-mole based on its output isn't much fun.
Automatically applying the corresponding transform on the result of malloc-type functions would be best. However, either `clang-tidy` does not provide such a check, or its name does not contain "cast"; `coccinelle` can easily *remove* such casts, but inverting the semantic patch isn't trivial.

sezero:
I prefer manual method - do: `CC=g++ ./configure`, edit Makefile and
add `-Wno-sign-compare`, then `make -k 2> 2.txt` and read the failure
log, apply casts where it points. Very boring legwork, but it works.

PR #423, opened 2021-07-16
Remove stb_vorbis sanity checks that are no longer required.

AliceLR:
Removes the following checks that no longer do anything useful:

* `compute_codewords` `len[k] >= 32`: obsoleted by bounds checking lengths when they're read in `start_decoder`.
* `compute_codewords` `z >= 32`: obsoleted by bounds checking lengths when they're read in `start_decoder`.
* `draw_line` `adx==0`: obsoleted by `lx != hx` in `do_floor`.
* `draw_line` `x >= n*2`: obsoleted by clamping `x1` to `n`.
* `draw_line` `y >= 256` (two places): obsoleted by indexing with `y&255`.
* `draw_line` return value: obsoleted by removal of previous checks.

Removing these gets the libxmp copy of stb_vorbis a little closer to upstream. I haven't vetted all of the extra checks that libxmp adds, but these ones were fairly easy.

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -1110,7 +1110,6 @@ static int compute_codewords(Codebook *c, uint8 *len, int n, uint32 *values)
 
    // sanity check
    if (k == n) return (c->sorted_entries == 0);
-   if (len[k] >= 32) return FALSE;
 
    assert(len[k] < 32); // no error return required, code reading lens checks this
    // add to the list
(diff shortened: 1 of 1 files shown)


AliceLR:
The other checks libxmp introduces look less trivial to trace back so I think this is OK to review as is (I'll submit a separate patch for anything else I find that is no longer necessary).

sezero (approved):
OK

Issue #420, opened 2021-07-14
lite version lacking src/win32/debug.c


sezero:
The lite version is not including `src/win32/debug.c`, not in makefiles
nor in release-packaging, which would make MSVC debug builds fail.  For
simplicity, I suggest moving all contents of it into src/win32.c.  OK?

PR #425, opened 2021-07-16
move src/win32/debug.c contents into src/win32.c:

sezero:
The 'lite' packages lacked that file, so MSVC builds with DEBUG defined would fail.
Fixes: https://github.com/libxmp/libxmp/issues/420


AliceLR (approved):
This is OK. It always seemed weird to me that this was a separate compilation unit...

PR #429, opened 2021-07-18
Avoid the dllimport spec being enforced and provide the version calls as the working alternative to variables

### LINKED ISSUE #427: Using the dllimport() at the header adds an extra pain for static linking
I was found a small thing at the `xmp.h` that there were "dllimport()" markers that were been used. I usually do NOT use them, mainly because I want to allow easy use of the library with both static and shared form without the necessity to define extra macros at the user side, otherwise, when "dllimport()" was been used at the static library, the link will fail.

So, the little example I made at my libModPlug fork and there is an explanation why I made it:
https://github.com/WohlSoft/AudioCodecs/commit/042b59701adba4f6201f867e006a1f887367fcb2

So, I suggest never use `dllimport()` at libraries intended to be both static and shared to simplify the user's choice of which library to link. Otherwise, this adds an extra pain to developers that require them to specify extra macros to link the same library statically.

I heard that `dllimport()` helps to initialize the shared library faster, however, I do see not so much difference in the workflow when using the same application with or without the `dllimport()`. But, I do think, when the library is intended to be both shared and static, so, don't use `dllimport()` at all, otherwise this adds an extra pain for users, letting them figure why your static library won't link to their applications.

Can you explain the real reason to use them here?

Wohlstand:
Fixes #427

### DIFF
--- a/include/xmp.h
+++ b/include/xmp.h
@@ -16,12 +16,13 @@ extern "C" {
 #define XMP_VER_RELEASE 0
 
 #if defined(_WIN32) && !defined(__CYGWIN__)
-# if defined(BUILDING_STATIC)
-#  define LIBXMP_EXPORT
-# elif defined(BUILDING_DLL)
+# if defined(BUILDING_DLL)
 #  define LIBXMP_EXPORT __declspec(dllexport)
 # else
-#  define LIBXMP_EXPORT __declspec(dllimport)
+#  define LIBXMP_EXPORT
+#  if !defined(BUILDING_STATIC)
+#   define LIBXMP_EXPORT_VAR __declspec(dllimport)
+#  endif
 # endif
 #elif defined(__OS2__) && defined(__WATCOMC__) && defined(__SW_BD)
 #  define LIBXMP_EXPORT __declspec(dllexport)
--- a/include/xmp.h
+++ b/include/xmp.h
@@ -336,6 +336,9 @@ typedef char *xmp_context;
 LIBXMP_EXPORT_VAR extern const char *xmp_version;
 LIBXMP_EXPORT_VAR extern const unsigned int xmp_vercode;
 
+LIBXMP_EXPORT const char *xmp_get_version     (void);
+LIBXMP_EXPORT const unsigned int xmp_get_vercode (void);
+
 LIBXMP_EXPORT int         xmp_syserrno        (void);
 
 LIBXMP_EXPORT xmp_context xmp_create_context  (void);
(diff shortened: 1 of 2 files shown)


include/xmp.h
```diff
@@ -16,12 +16,11 @@ extern "C" {
...
 #if defined(_WIN32) && !defined(__CYGWIN__)
-# if defined(BUILDING_STATIC)
-#  define LIBXMP_EXPORT
-# elif defined(BUILDING_DLL)
+# if defined(BUILDING_DLL)
 #  define LIBXMP_EXPORT __declspec(dllexport)
 # else
-#  define LIBXMP_EXPORT __declspec(dllimport)
+#  define LIBXMP_EXPORT
+#  define LIBXMP_EXPORT_VAR __declspec(dllimport)
```
sezero, line 23: This patch is wrong: it attaches `dllimport` attribute to vars in static builds
which we don't want at all.

Wohlstand: You right, I should avoid this line being used when `BUILDING_STATIC` is enabled. I'll fix this tomorrow, now I'm at the bed from the phone.

Wohlstand: Fixed just now, please verify

sezero: And now you rely on `BUILDING_STATIC` again which you want to avoid..

Wohlstand: > And now you rely on BUILDING_STATIC again which you want to avoid..

The big problem was all calls in general, the last thing left were these constants. To deal with that I added calls to let users use them instead of variables, but, variables weren't removed to keep the compatibility with old code that already uses `BUILDING_STATIC` at their side. So, these variables should be just deprecated, not removed (until the next major)

Wohlstand: Also, the note: presence/absence of the `dllimport` affects only the client-side compilation and absolutely doesn't affects the compatibility with the existing DLL built. I mean, when there are two clients built with and without `dllimport`, they will still work fine.

Wohlstand:
Anything is needed to complete here, or the thing looks fine?

P.S. As an alternative way, we can inject the "static" macro into the header when building the "static only" package.

sezero:
This looks like a WONTFIX now, especially after https://github.com/libxmp/libxmp/pull/641. Closing.

Wohlstand:
I'll test the recent stuff, and will remake a thing. I still need to make sure the use of static builds should don't enforce user-side to declare the `LIBXMP_STATIC` macro on their end.

Wohlstand:
> This looks like a WONTFIX now, especially after https://github.com/libxmp/libxmp/pull/641. Closing.

The #641 only fixes the use of libxmp as an external project or as a sub-directory, or when using it via `find_package()` thing. However, it does NOT fixes the problem when using libxmp directly (with manual definition of the header path and linking flags). Espeically for cases when project had been built from sources and then being re-used in binary form through other project format like via Visual Studio's native project format.

Anyway, I will completely redo this, but with basing on the latest thing.

PR #437, opened 2021-07-19
Extract Galaxy Music System fuzz files packed in invalid DEFLATE.

AliceLR:
Should fix the tests broken in #436. These files were packed in MUSE with invalid DEFLATE streams that, for whatever reason, the old inflate algorithm allowed but miniz does not.


AliceLR:
I checked the *actual* coverage report and the issue was what I suspected: the old inflate implementation bails out when it encounters invalid block type 3, but doesn't reject it outright.

sezero:
If I understand correctly, the three test cases were compressed and this patch
extracts them, yes? Question:
- The extracted sizes seem smaller - because all the files are broken?
- The two are already supposed to fail, albeit with a different error code?
- One of them, test_fuzzer_gal4_env_point_bound, is supposed to load,
  was fuzzing done at the `muse` level or `gal[4|5]` level?

AliceLR:
> * The extracted sizes seem smaller - because all the files are broken?

The original MUSE files were broken, yes. The MUSE depacker allowed the partial decompressed "modules" to continue through to the loaders despite this. (FWIW before trying XZ compression I confirmed that the decompressed "modules" further gzip to about half of their size current, which should be fine for Git.)

> * The two are already supposed to fail, albeit with a different error code?

Yes. They're supposed to fail at particular points in the Galaxy loaders, but were failing due to the corrupted inflate streams instead.

> * One of them, test_fuzzer_gal4_env_point_bound, is supposed to load,
>   was fuzzing done at the `muse` level or `gal[4|5]` level?

The fuzz file was likely derived from a valid compressed MUSE file. I don't think there are any valid uncompressed Galaxy modules in libxmp's test files to be fuzzed (as such modules were always distributed as MUSE packed files with the .j2b extension).

sezero:
OK then.

sezero (approved):

Issue #434, opened 2021-07-19
warnings from depackers/vorbis.c:


sezero:
The warning shows up with `-O3` which is what cmake uses for release builds,
not with `-O2` which is what autotools use:
```
src/depackers/vorbis.c: In function 'stb_vorbis_get_frame_short_interleaved':
src/depackers/vorbis.c:5421: warning: 'output' may be used uninitialized in this function
src/depackers/vorbis.c: In function 'stb_vorbis_decode_memory':
src/depackers/vorbis.c:5421: warning: 'output' may be used uninitialized in this function
src/depackers/vorbis.c:5421: note: 'output' was declared here
```
The thing is, `stb_vorbis_get_frame_short_interleaved` calls `stb_vorbis_get_frame_float`
but checks `if (len)` instead of `if (len > 0)` because `stb_vorbis_get_frame_float` can
return `-1` on error, a libxmp modification. Other callers of `stb_vorbis_get_frame_float`
have the same issue, e.g., `stb_vorbis_get_samples_float`, `stb_vorbis_get_samples_short`,
`stb_vorbis_get_samples_float_interleaved` and `stb_vorbis_get_samples_short_interleaved`.

PR #441, opened 2021-07-20
vorbis depacker: respect stb_vorbis_get_frame_float() error returns

sezero:
Fixes: https://github.com/libxmp/libxmp/issues/434

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -5029,7 +5029,8 @@ int stb_vorbis_seek(stb_vorbis *f, unsigned int sample_number)
    if (sample_number != f->current_loc) {
       int n;
       uint32 frame_start = f->current_loc;
-      stb_vorbis_get_frame_float(f, &n, NULL);
+      if (stb_vorbis_get_frame_float(f, &n, NULL) < 0)
+          return error(f, VORBIS_seek_failed);
       assert(sample_number > frame_start);
       assert(f->channel_buffer_start + (int) (sample_number-frame_start) <= f->channel_buffer_end);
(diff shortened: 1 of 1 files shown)


AliceLR (approved):
This looks OK. Do you want to upstream this patch or should I?

sezero:
> This looks OK. Do you want to upstream this patch or should I?

You do - provided that `stb_vorbis_get_frame_float` error return is upstreamed too

AliceLR:
On further review, I'm not 100% sure libxmp returning -1 from `stb_vorbis_get_frame_float` is correct: the `error()` usage in that function that exists upstream returns 0, so 0 seems to be the intended return value in an error state. The warnings/potential errors may be our fault because we're using the wrong error return value.

sezero:
However, `stb_vorbis_get_frame_float` does return 0 without actually meaning error in there, am I wrong?

AliceLR:
That's correct, it may be able to return 0 through normal behavior. The only place where it explicitly returns 0 is when `vorbis_decode_packet` fails. This seems to result from either an error case in `vorbis_decode_packet_rest` or when `vorbis_decode_initial` detects an error state and rejects the packet. The latter appears to be in line with the Vorbis I spec, which I've been trying to make sense of: https://xiph.org/vorbis/doc/Vorbis_I_spec.html#x1-730004.3.1

My take on this is that if the explicit error state in the spec only results in rejection of the packet and if returning 0 results in this, that might be good enough. I'm going to try to figure out more about what's going on here, since those two error returns in `stb_vorbis_get_frame_float` are among the error checks I haven't submitted an upstream patch for yet because *I can't prove to myself whether or not they currently do anything meaningful*.

PR #443, opened 2021-07-21
Fix heap corruption in Coconizer loader with invalid sequences.

AliceLR:
A missing bounds check in the Coconizer sequence loading loop allows it to write well past the end of the `xmp_module` struct and into the pointers in `module_data`, predictably causing crashes and/or invalid frees. I also took the opportunity to fix the part of this loader affected by #400.

I'm not 100% sure whether or not libxmp's Coconizer sequence loading is correct due to vague documentation, so the test allows either `-XMP_ERROR_LOAD` or 0 as the return value (the important thing is it doesn't crash).


src/loaders/coco_load.c
```diff
@@ -276,20 +277,22 @@ static int coco_load(struct module_data *m, HIO_HANDLE *f, const int start)
...
-			event->ins = hio_read8(f);
-			event->note = hio_read8(f);
-			if (event->note)
-				event->note += 12;
-
-			if (hio_error(f)) {
-				return -1;
+		for (j = 0; j < 64; j++) {
+			for (k = 0; k < mod->chn; k++) {
+				event = &EVENT(i, k, j);
```
sezero, line 282: Unless I'm missing something (newly woke up), this is reading (0,0,0), (0,1,0), (0,2,0)
versus the old (0,0,0), (0,0,1), (0,0,2): are we sure that it's correct?

AliceLR: That is what the old version did too: the modulo is on the second parameter and the division is on the third. (I think if those had been reversed the format test would fail entirely due to different pattern data.)

sezero: Oh hell, as I'm still waking up..

sezero (approved):

PR #444, opened 2021-07-22
Remove unused sys/stat.h includes in depackers.

AliceLR:
Minor cleanup for the patches referenced in #442: these files no longer need to include `sys/stat.h`.


sezero (approved):
Huh.. forgot these, thanks.

PR #445, opened 2021-07-23
Fix ARC/ArcFS slow loads due to large allocations and slow RLE.

AliceLR:
Fixes two bugs in the ARC/ArcFS LZW decoder reported by @debrouxl in #383. 

* The LZW decoder attempted to allocate arbitrarily large buffers for its output based on the original size field in the header. There's no convincing way to heuristically calculate a maximum allowed original size for a packed size: the LZW approaches 2600:1 best case compression ratios for 13 bit codes, presumably much higher with RLE turned on or for 16 bit codes. A 2GB file under that best case compression rate would fit into a reasonable 700kb. I just added a hard cap of `(1 << 28)` for the output filesize, which isn't perfect, but anything better is much harder and not worth the effort for this contrib depacker code from 2001.
* Both the LZW code output and RLE expansion lacked the ability to bail out early at the end of the buffer, despite checking for the end of the buffer. This resulted in particular types of constructed files being VERY slow to load.


AliceLR:
(Wow, the codecov report is actually correct!)

src/depackers/readhuff.c
```diff
@@ -38,7 +38,7 @@ static int rawinput(struct data_in_out *io)
 	return -1;
 }
 
-static void rawoutput(int byte, struct data_in_out *io)
+static int rawoutput(int byte, struct data_in_out *io)
```
sezero, line 41: This doesn't return anything

AliceLR: Thanks, fixed. I missed it because libxmp doesn't actually use this file...

sezero (approved):

PR #447, opened 2021-07-23
replace stat() usage with target-specific code in test_ / load_module()

sezero:


AliceLR:
I think I'm OK with this overall. A couple of questions:

* Since there's a lot of `#ifdef`/platform-specific hackery here would it be better to move this to a different translation unit (similar to tempfile.c)?
* Earlier I was considering PRing a `memset()` hack to these to shut up the MSan false positives for `stat()` with LLVM 12. It isn't strictly necessary, but since it's only in one spot now, do you think it would be worth adding?

sezero:
>  * Since there's a lot of `#ifdef`/platform-specific hackery here would it be better to move this to a different translation unit (similar to tempfile.c)?

Sure - give me a file name

>  * Earlier I was considering PRing a `memset()` hack to these to shut up the MSan false positives for `stat()` with LLVM 12. It isn't strictly necessary, but since it's only in one spot now, do you think it would be worth adding?

Like `memset(&st,0,sizeof(struct stat))`? (Is it me, or is llvm getting more overzealous?)

AliceLR:
> Sure - give me a file name

Maybe `filetype.c`? Not sure...

> Like `memset(&st,0,sizeof(struct stat))`?

Yes. I think `st` is generally guaranteed to be initialized if `stat()` returns success, but oh well.

> (Is it me, or is llvm getting more overzealous?)

The weird part is the sanitzers are developed by [Google](https://github.com/google/sanitizers), but the code is part of LLVM and it's LLVM that knows about this issue but [keeps](https://lists.llvm.org/pipermail/llvm-dev/2021-May/150763.html) [releasing](https://lists.llvm.org/pipermail/llvm-dev/2021-June/151233.html) [without](https://groups.google.com/g/llvm-dev/c/G8ksPVTym3Q/m/ZXm5E_OcAwAJ) [fixing it](https://groups.google.com/g/llvm-dev/c/5hhRFLi3nAY/m/28uD2ppQCAAJ).

sezero:
Working on it

sezero:
Done - please review.

AliceLR (approved):
This looks good to me!

PR #453, opened 2021-07-24
vorbis: minor correction to gcc always_inline attribute use.

sezero:


AliceLR (approved):
This seems fine to me.

The earliest reference I can find to `__attribute__((always_inline))` in the GCC documentation is the 3.4.1 changelog, which implies it existed before, so great. I don't have builds of these ancient GCC versions around, do you have a source for `always_inline` having been added in this exact version?

sezero:
I remember, from memory, that it was introduced in gcc3.2, but now
I can't find an exact documentation.

Googling however, resulted in cases where gcc < 4 gave errors with
`-O0`, and an eigen header even enables it for gcc >= 4.2. Hmph...
https://www.boost.org/doc/libs/1_50_0/boost/config/suffix.hpp
http://eigen.tuxfamily.org/dox-3.2/Macros_8h_source.html
https://cs.fit.edu/code/projects/nothing_works/repository/revisions/680765abece33eab94e7054de4d1fd43373f1dd6/diff/src/node_http_parser.cc

Maybe we should enable it for gcc >= 4.2, or at least 4.0?

AliceLR:
> Maybe we should enable it for gcc >= 4.2, or at least 4.0?

I *may* still have an environment with GCC 4.1.x or 4.2.x available, let me check.

sezero:
Searched a little more, Amiga SDI headers enable it for gcc >= 3.0
https://github.com/adtools/SDI/blob/master/SDI_compiler.h#L143

The sys/cdes.h header in my linux setup enables it for gcc >= 3.2:
```
#if __GNUC_PREREQ (3,2)
# define __always_inline __inline __attribute__ ((__always_inline__))
#else
# define __always_inline __inline
#endif
```

AliceLR:
GCC 4.1.2 in my SPARC Debian 4.0 r4a VM was able to build vorbis.c both at -O0 and -O2 with no issues, so  gcc >= 4 *might* be fine.

sezero:
I have RedHat's gcc3.4.6, a cross-djgpp gcc3.4.6, an old mingw gcc3.4.5,
also just built gcc-3.3.2 from source: no failures with either of them, at `-O2`
or at `-O0`

So I'll just assume gcc >= 3.2 requirement of glibc is correct?

AliceLR:
OK, I think this can be merged as-is then.

PR #454, opened 2021-07-24
Fix UMRs, large allocations, unaligned pointers found by libFuzzer.

AliceLR:
Large allocation fixes:
* The StoneCracker 4.04 depacker now has a bounds check on the supposed original size to make sure it's actually possible to generate with a stream of the given input size. I did the bounding calculations manually, so feel free to double check them (see the comment in the diff).

Alignment fixes:
* (UBSan) The XPK depacker would load stored file checksums and calculate file checksums using unaligned pointers in some cases. It was also reading these as native endian instead of big endian (though since it's just an XOR this ended up cancelling out anyway).

UMR fixes:
* (MSan) The compress depacker did not reject maxbits values of less than 9, which could lead to it failing to generate referenced code data and thus referencing these later as undefined stack garbage.
* (Memcheck) The ABK test function did not initialize the module title buffer, causing UMRs in `test_module`.
* (Memcheck) The DBM, Galaxy 4.0/5.0, and PT 3.6 loaders didn't check the `hio_read` return values when reading module titles.

Other:
* The EMOD loader didn't sanitize instrument names with `libxmp_instrument_name` or check the `hio_read` return value when reading them.


src/depackers/uncompress.c
```diff
@@ -90,7 +90,7 @@ static int decrunch_compress(FILE * in, FILE * out, long inlen)
 	block_mode = inbuf[2] & BLOCK_MODE;
 	maxmaxcode = MAXCODE(maxbits);
 
-	if (maxbits > BITS) {
+	if (maxbits < INIT_BITS || maxbits > BITS) {
 		/*fprintf(stderr,
```
sezero, line 94: Is this, or an equivalent of it, applied in any Linux distros? Can't find it in https://github.com/vapier/liblzw/blob/main/lzw.c

src/depackers/unsqsh.c
```diff
@@ -45,12 +45,13 @@ static uint8 ctable[] = {
...
 
-static uint16 xchecksum(uint32 * ptr, uint32 count)
+static uint16 xchecksum(uint8 *ptr, uint32 count)
 {
-	register uint32 sum = 0;
+	uint32 sum = 0;
 
 	while (count-- > 0) {
-		sum ^= *ptr++;
+		sum ^= readmem32b(ptr);
```
sezero, line 53: this looks good

src/loaders/emod_load.c
```diff
@@ -104,7 +105,9 @@ static int get_emic(struct module_data *m, int size, HIO_HANDLE * f, void *parm)
 		hio_read8(f);	/* num */
 		sub->vol = hio_read8(f);
 		xxs->len = 2 * hio_read16b(f);
-		hio_read(xxi->name, 1, 20, f);
+		if (hio_read(name, 1, 20, f) < 20)
+			return -1;
+		libxmp_instrument_name(mod, i, name, 20);
```
sezero, line 110: this looks good

src/loaders/abk_load.c
```diff
@@ -435,6 +437,16 @@ static int abk_test(HIO_HANDLE *f, char *t, const int start)
         return -1;
     }
 
+    /* Attempt to read title. */
+    hio_read32b(f); /* instruments_offset */
+    song_section_offset = hio_read32b(f);
+
+    if (t != NULL && read_abk_song(f, &song, AMOS_MAIN_HEADER + song_section_offset) == 0)
+    {
+        memcpy(t, song.song_name, AMOS_STRING_LEN);
```
sezero, line 446: took me some hunting in the code, but looks OK.

Should you run libxmp_copy_adjust() on the name?

src/loaders/dbm_load.c
```diff
@@ -509,7 +509,8 @@ static int dbm_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	data.min_version = version & 0xFF;
 
 	hio_seek(f, 10, SEEK_CUR);
-	hio_read(name, 1, 44, f);
+	if (hio_read(name, 1, 44, f) < 44)
+		return -1;
```
sezero, line 513: OK.

src/loaders/gal4_load.c
```diff
@@ -67,7 +67,8 @@ static int get_main(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	char buf[64];
 	int flags;
 
-	hio_read(buf, 1, 64, f);
+	if (hio_read(buf, 1, 64, f) < 64)
+		return -1;
```
sezero, line 71: OK.

src/loaders/gal5_load.c
```diff
@@ -72,7 +72,8 @@ static int get_init(struct module_data *m, int size, HIO_HANDLE *f, void *parm)
 	char buf[64];
 	int flags;
 
-	hio_read(buf, 1, 64, f);
+	if (hio_read(buf, 1, 64, f) < 64)
+		return -1;
```
sezero, line 76: OK.

src/loaders/pt3_load.c
```diff
@@ -162,7 +162,8 @@ static int pt3_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	hio_read32b(f);		/* VERS */
 	hio_read32b(f);		/* VERS size */
 
-	hio_read(buf, 1, 10, f);
+	if (hio_read(buf, 1, 10, f) < 10)
+		return -1;
```
sezero, line 166: OK.

src/depackers/s404_dec.c
```diff
@@ -389,6 +389,18 @@ static int decrunch_s404(FILE *in, FILE *out, long inlen)
...
+   * Best case ratio of S404 sliding window:
+   *
+   *  2-3:  9b + (>=1b) -> 2-3B  ->  24:10
+   *  4-7:  9b + (>=3b) -> 4-7B  ->  56:12
+   *  8:22: 9b + (>=6b) -> 8-22B -> 176:15
+   *  23+:  9b + 3b + 8b * floor((n-23)/255) + 7b + (>=0b) -> n B -> ~255:1
+   */
+  if (pLen < (oLen / 255)) {
+    goto error;
+  }
```
sezero, line 402: Looks sane. Tested on valid files?

src/depackers/uncompress.c
AliceLR: I don't know, I haven't checked. A maxbits value under 9 is nonsensical because the initial bits for this type of LZW is always 9. [POSIX claims a valid under 9 for the maxbits argument](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/compress.html) to `compress` is invalid.

src/loaders/abk_load.c
AliceLR: That's probably a better idea, yes—will patch soon.

src/depackers/s404_dec.c
AliceLR: Only on whatever is in test-dev/. Do you have a link to other files I could test this on?

sezero: OK, tested with four files I had around, loads OK.

src/depackers/uncompress.c
sezero: OK.

sezero (approved):
OK.  Can push after you further patch abk title load.

AliceLR:
OK, pushed.

Issue #396, opened 2021-06-25
move USE_VERSIONED_SYMBOLS to configure?


sezero:
`USE_VERSIONED_SYMBOLS` (see in common.h) basically requires visibility
attributes and actually GNU ld as the linker so that `-Wl,--version-script,libxmp.map`
is supported. There is also a list of platforms that common.h opts out of versioned
symbols.

We can check GNU ld like the following and set `LD_VERSCRIPT` like the following:
```diff
diff --git a/configure.ac b/configure.ac
index 909cc96..4a06e99 100644
--- a/configure.ac
+++ b/configure.ac
@@ -11,6 +11,15 @@ AC_CANONICAL_HOST
 AC_PROG_CC
 AC_PROG_INSTALL
 AC_PROG_RANLIB
+AC_CHECK_TOOL([LD], [ld], [:])
+dnl based on libtool's lt_cv_prog_gnu_ld:
+AC_CACHE_CHECK([if the linker ($LD) is GNU ld], ac_cv_prog_gnu_ld,
+[case `$LD -v 2>&1 </dev/null` in
+  *GNU* | *'with BFD'*)
+     ac_cv_prog_gnu_ld=yes ;;
+  *) ac_cv_prog_gnu_ld=no  ;;
+esac])
+
 AC_C_BIGENDIAN
 
 dnl XMP_TRY_COMPILE(<message>,<cache-var>,<flags>,<program>,<ifyes>,<ifno>)
@@ -132,7 +141,9 @@ if test $ac_cv_c_flag_f_visibility_hidden = no; then
    CFLAGS="${old_CFLAGS}"
 else
    CFLAGS="${old_CFLAGS} -fvisibility=hidden -DXMP_SYM_VISIBILITY"
-   LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
+   if test $ac_cv_prog_gnu_ld = yes; then
+      LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
+   fi
 
 dnl https://github.com/libxmp/libxmp/issues/385
    old_CFLAGS="${CFLAGS}"
```
The setting of `LD_VERSCRIPT` also is in the visibility==yes case, so that one
is handled, which also filters out Amiga variants, dos and windows (see configure.ac).
OSX linked is not GNU ld, so the `__APPLE__` filter is handled too.

What is left is android, emscripten, beos, atheos, and mint.
(Why is android filtered out of versioned symbols use, btw?)
[...]

PR #456, opened 2021-07-27
move USE_VERSIONED_SYMBOLS decision to configuration time:

sezero:
Versioned symbols use requires symbols visilibity and GNU ld as linker,
which our configury checks for.  common.h has additional black-listing
against target OSs, like Android, Apple, etc:

- Apple: macOS uses its own non-GNU ld and is detected by configury, so
  it is handled.
- Android: The blacklist should have been because jni/Android.mk is set
  to produce a static library.  There is nothing else preventing it, so
  the new configury doesn't blacklist it.
- Emscripten, BeOS, Atheos, mint: New configury manually blacklists them
  to match original common.h, just in case.
- Win32, amiga: already filtered out by not having symbol visibility.

Differently: the new configury version enables the versioned symbols only
for shared libraries, not static.

Closes: https://github.com/libxmp/libxmp/issues/396

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -142,13 +142,25 @@ case $host_os in
 	;;
 esac
 
+have_versioned_symbols=no
 if test $ac_cv_c_flag_f_visibility_hidden = no; then
    CFLAGS="${old_CFLAGS}"
 else
    CFLAGS="${old_CFLAGS} -fvisibility=hidden -DXMP_SYM_VISIBILITY"
-   LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
 
-dnl https://github.com/libxmp/libxmp/issues/385
+   if test $ac_cv_prog_gnu_ld = yes; then
+      case "${host_os}" in
+        emscripten*|beos*|atheos*|*mint)
+          ;;
+        *) if test "${enable_shared}" != no; then
+              have_versioned_symbols=yes
+              CFLAGS_SHARED="${CFLAGS_SHARED} -DUSE_VERSIONED_SYMBOLS=1"
+         fi
+         LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
+         ;;
+      esac
+   fi
+
    old_CFLAGS="${CFLAGS}"
    if test $ac_cv_c_flag_w_error = yes; then
       CFLAGS="${CFLAGS} -Werror"
@@ -166,6 +178,9 @@ void foo(void) __attribute__((__symver__("foo@bar")));],
     [CFLAGS="${CFLAGS} -DHAVE_ATTRIBUTE_SYMVER"],
     [CFLAGS="${CFLAGS}"])
 fi
+
+AC_MSG_CHECKING(whether to enable versioned symbols support)
+AC_MSG_RESULT($have_versioned_symbols)
 
 dnl for clang
 XMP_TRY_COMPILE(whether compiler understands -Wunknown-warning-option,
--- a/lite/configure.ac
+++ b/lite/configure.ac
@@ -145,13 +145,25 @@ case $host_os in
 	;;
 esac
 
+have_versioned_symbols=no
 if test $ac_cv_c_flag_f_visibility_hidden = no; then
    CFLAGS="${old_CFLAGS}"
 else
    CFLAGS="${old_CFLAGS} -fvisibility=hidden -DXMP_SYM_VISIBILITY"
-   LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
 
-dnl https://github.com/libxmp/libxmp/issues/385
+   if test $ac_cv_prog_gnu_ld = yes; then
+      case "${host_os}" in
+        emscripten*|beos*|atheos*|*mint)
+          ;;
+        *) if test "${enable_shared}" != no; then
+              have_versioned_symbols=yes
+              CFLAGS_SHARED="${CFLAGS_SHARED} -DUSE_VERSIONED_SYMBOLS=1"
+         fi
+         LD_VERSCRIPT="-Wl,--version-script,libxmp.map"
+         ;;
+      esac
+   fi
+
    old_CFLAGS="${CFLAGS}"
    if test $ac_cv_c_flag_w_error = yes; then
       CFLAGS="${CFLAGS} -Werror"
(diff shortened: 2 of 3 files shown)


AliceLR (approved):
Tried the following things and it seems to be fine:

* MSYS2: doesn't do any of the versioned symbols checks.
* GCC 4.8.5 (CentOS 7): detects `externally_visible`, uses versioned symbols, builds and links fine.
* GCC 11.1.1 (Fedora 34, default `LD`, `LD=ld`, `LD=ld.gold`): detects both `externally_visible` and `__symver__`, uses versioned symbols, builds/links OK.
* GCC 11.1.1 (Fedora 34, `LD=lld`): detects both `externally_visible` and `__symver__`, does NOT use versioned symbols, builds/links OK.
* clang 12 (Fedora 34, CC=clang, LD=lld): doesn't detect `externally_visible` or `__symver__`, does NOT use versioned symbols, builds/links OK.

Not as sure whether this is correct:

* clang 12 (Fedora 34, default `LD`, `LD=ld`, `LD=ld.gold`): detects neither `externally_visible` nor `__symver__`, *claims to use versioned symbols*, builds/links OK.

sezero:
>     * MSYS2: doesn't do any of the versioned symbols checks.

Expected, good

>     * GCC 4.8.5 (CentOS 7): detects `externally_visible`, uses versioned symbols, builds and links fine.
>     * GCC 11.1.1 (Fedora 34, default `LD`, `LD=ld`, `LD=ld.gold`): detects both `externally_visible` and `__symver__`, uses versioned symbols, builds/links OK.

Expected, good.

>     * GCC 11.1.1 (Fedora 34, `LD=lld`): detects both `externally_visible` and `__symver__`, does NOT use versioned symbols, builds/links OK.

Does lld (clang's own linker, yes?) support `--version-script`? If it does,
then we should detect and support it.

>     * clang 12 (Fedora 34, `LD=lld`): doesn't detect `externally_visible` or `__symver__`,

That part is normal (clang doesn't support those attribs),

> does NOT use versioned symbols, builds/links OK.

And that's OK to build/link OK.  But the `lld` question above still applies

> Not as sure whether this is correct:
> 
>     * clang 12 (Fedora 34, default `LD`, `LD=ld`, `LD=ld.gold`): detects neither `externally_visible` nor `__symver__`, _claims to use versioned symbols_, builds/links OK.

That's _should_ be OK. Either `__symver__` or `externally_visible` seems to be
required by gcc lto.

sezero:
Another question about `lld`: How are you setting it up? I mean, if you
are exporting `LD=lld`, is the frontend (clang) actually calling lld?

sezero:
OK, I looked at online documents, `lld` should support `--version-script`,
But: lld is supposed to be invoked not directly, but target-specifically:

```
$ lld --version
lld is a generic driver.
Invoke ld.lld (Unix), ld64.lld (macOS), lld-link (Windows), wasm-ld (WebAssembly) instead
```

Then, with a clang-5.0.2 installation:
```
$ ld.lld -v
LLD 5.0.2 (compatible with GNU linkers)
ld.lld: error: no input files
ld.lld: error: target emulation unknown: -m or at least one .o file required

$ ld.lld --version
LLD 5.0.2 (compatible with GNU linkers)
```

Or, with a clang-11.1 installation:
```
$ ld.lld -v
LLD 11.1.0 (compatible with GNU linkers)

$ ld.lld --version
LLD 11.1.0 (compatible with GNU linkers)
```

So: `lld`, only if invoked as `ld.lld`, should be supported, and actually
be detected by configury -- see the grep for `GNU` in ld output at top of
of configure.ac.

However, invoking it directly won't be detected as GNU ld, and it won't
work either. If invoked as ld64.lld, then it pretends like Apple's ld64
and won't support `--version-script`.

Bottom line:  If you attempted to use lld only by setting `LD=lld`, it
wouldn't have worked, and the fact that linkage succeeded means the clang
or gcc frontend didn't use it. I yet don't know how to force clang or gcc
to use ld.lld instead of system's ld.

AliceLR:
> Another question about `lld`: How are you setting it up? I mean, if you
> are exporting `LD=lld`, is the frontend (clang) actually calling lld?

All places where I mentioned `LD=something`, I provided that exact setting to `./configure`, e.g. `LD=lld ./configure`. libxmp's Makefile seems to invoke `${LD}` rather than `${CC}` to link. The default for `${LD}` seems to be `${CC}` (so the "default" I mentioned above will be either `cc` or `clang`) but providing a different value to `./configure` overrides it in the Makefile.

I didn't try `LD=ld.lld`, so the results for that. With GCC, both `__symver__` and `externally_visible` are detected and versioned symbols support is enabled as expected. With clang, neither are detected, and it claims to enable versioned symbols support like with the other GNU linkers.

sezero:
> All places where I mentioned `LD=something`, I provided that exact setting to `./configure`, e.g. `LD=lld ./configure`. libxmp's Makefile seems to invoke `${LD}` rather than `${CC}` to link. The default for `${LD}` seems to be `${CC}` (so the "default" I mentioned above will be either `cc` or `clang`) but providing a different value to `./configure` overrides it in the Makefile.

Oh no: Do `LD=ld ./configure`, you'll still have `LD = $(CC)` in the Makefile,
because Makefile.in doesn't use `@LD@` for the LD assignment. However when you
run `make LD=ld`, the generated makefile will respect your override, and then
will fail.

I think, this is ready for pushing: Should I apply or do you want more research?

AliceLR:
> > All places where I mentioned `LD=something`, I provided that exact setting to `./configure`, e.g. `LD=lld ./configure`. libxmp's Makefile seems to invoke `${LD}` rather than `${CC}` to link. The default for `${LD}` seems to be `${CC}` (so the "default" I mentioned above will be either `cc` or `clang`) but providing a different value to `./configure` overrides it in the Makefile.
> 
> Oh no: Do `LD=ld ./configure`, you'll still have `LD = $(CC)` in the Makefile,
> because Makefile.in doesn't use `@LD@` for the LD assignment. However when you
> run `make LD=ld`, the generated makefile will respect your override, and then
> will fail.

You're right, sorry. I missed that somehow or mistook the different `./configure` output for that being the case.

> I think, this is ready for pushing: Should I apply or do you want more research?

I think this is OK to merge.

PR #457, opened 2021-07-28
Reduce performance overhead of the Protracker MOD loader.

AliceLR:
See: #400.

* Removes unnecessary integer division in the pattern event loading loop that would generate IDIV instructions or worse.
* The non-lite MOD loader now loads the entire pattern into a single allocated memory block to reduce seeking and IO overhead.
* Add const to `libxmp_decode_protracker_event`, `libxmp_decode_noisetracker_event` `mod_event` fields (to make it more obvious they won't modify the source data).

I haven't *personally* tested the lite MOD loader change but it should be OK. These changes dropped something like 10 seconds off of loading my entire ModLand Protracker folder with `xmp_load_module_from_memory`.


AliceLR:
Updated to include a second patch from #400 for division instruction usage in `mread`.

sezero:
Have you finalized this?

AliceLR:
Yes, I do not have any further things to add to this particular patch.

sezero (approved):

PR #461, opened 2021-07-29
XM and other misc. loading performance improvements.

AliceLR:
See #400. Fixes the following issues that affected loader performance:

* The XM pattern loader no longer relies on integer division instructions and allocates a single pattern work buffer instead of a new one for every pattern.
* The track pointers for the current pattern are now cached during `scan_module`, significantly improving scan performance by avoiding pointer aliasing overhead.
* The MD5 sum performed in `load_module` is delayed until after several sanity checks. This is mainly intended to help fuzzing performance. (Patch by @debrouxl.)

The XM diff is a little messy, which could maybe be worked around with an indentation hack for now if that's preferable.

### DIFF
--- a/src/scan.c
+++ b/src/scan.c
@@ -50,7 +50,9 @@ static int scan_module(struct context_data *ctx, int ep, int chain)
 {
     struct player_data *p = &ctx->p;
     struct module_data *m = &ctx->m;
-    struct xmp_module *mod = &m->mod;
+    const struct xmp_module *mod = &m->mod;
+    const struct xmp_track *tracks[XMP_MAX_CHANNELS];
+    const struct xmp_event *event;
     int parm, gvol_memory, f1, f2, p1, p2, ord, ord2;
     int row, last_row, break_row, row_count, row_count_total;
     int orders_since_last_valid, any_valid;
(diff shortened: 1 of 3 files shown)


src/scan.c
```diff
@@ -239,10 +245,11 @@ static int scan_module(struct context_data *ctx, int ep, int chain)
 	    pdelay = 0;
 
 	    for (chn = 0; chn < mod->chn; chn++) {
-		if (row >= mod->xxt[mod->xxp[pat]->index[chn]]->rows)
+		if (row >= tracks[chn]->rows)
 		    continue;
 
-		event = &EVENT(mod->xxo[ord], chn, row);
+		//event = &EVENT(mod->xxo[ord], chn, row);
+		event = &tracks[chn]->event[row];
```
sezero, line 252: Can't think straight right now - we're no longer referencing mod->xxo here anymore?

AliceLR: Right— `mod->xxo[ord]` is stored to `pat` at line 149 and neither `ord` nor `pat` are modified after that part of the loop. This means the track pointers can be determined once instead of every single time an event is read. This is what `EVENT()` does, for reference:
```C
#define TRACK_NUM(a,c)	m->mod.xxp[a]->index[c]
#define EVENT(a,c,r)	m->mod.xxt[TRACK_NUM((a),(c))]->event[r]
```
Every single event it was dereferencing the pattern to get an index, then dereferencing the `mod->xxt` to get the track pointer, which is plenty of opportunities for a paranoid compiler to stick in explicit loads because it doesn't know `m->scan_cnt` and `p->sequence_control` don't somehow alias these.

sezero: I see, OK then.

sezero (approved):

PR #466, opened 2021-07-31
Update stb_vorbis EOF checks to match upstream PR.

AliceLR:
Updates the EOF hacks added in previous patches to match the [current upstream PR](https://github.com/nothings/stb/pull/1174), which fixes additional leaks in `start_decoder` that the current checks missed. Also removes two libxmp-specific checks in `start_decoder` that I'm *mostly* certain do nothing of value.

* The removed check at line 4129 in this diff is easy to verify: upstream stb_vorbis now performs an identical check at line 4107→4096 in this diff.
* The removed check at line 4096 is a little harder to verify, but I can't find anywhere in the [standard](https://xiph.org/vorbis/doc/Vorbis_I_spec.html) that indicates `r->end - r->begin` should be 1024 or less apart. There are a few things going on here that makes me suspect this check is not helping very much, though:
  * Line 4102→4091: cases where `r->begin` is greater than `r->end` are already filtered by the upstream code.
  * Line 4256→4239: the first time these are used, they're bounded to `f->blocksize_1 / 2` as per the specification. `f->blocksize_1` is never larger than 8192, so the effective upper bound of their difference is already 4096 i.e. not much bigger in scale than the libxmp check enforced.
  * Line ?→2150: the other place where `r->begin` and `r->end` are used, a similar bounding is performed. You can trace the value it is being bounded by back to `vorbis_decode_packet_rest`, where you'll find it is once again `f->blocksize_1 / 2`. ~~I suspect some of the `r->begin` uses in `decode_residue` are *maybe* supposed to be `limit_r_begin`, which would make the libxmp-specific hack in that function no longer necessary, but I'm not 100% sure.~~ (edit: nevermind, this is fine and that check can be removed in another patch—the usage of `limit_r_begin` to derive `part_read` makes it impossible for the later calculations to be out of bounds.)

That second removed check should probably be verified by someone with more Vorbis knowledge than I have.


AliceLR:
Force pushed to fix a minor style inconsistency. This patch should be OK for review.

sezero (approved):
I followed the code as best as I could, this looks OK.

PR #468, opened 2021-08-01
Add -Wextra -Wno-unused-parameter -Wno-sign-compare to CFLAGS.

AliceLR:
Adds -Wextra, minus the warnings it enables that are useless (-Wunused-parameter) or too hard to fix right now (-Wsign-compare), to CFLAGS.

Warning fixes required for the new warnings:
* The non-lite Protracker MOD loader used a sentinel for the MOD magic list that wasn't fully defined. I replaced the sentinel with `ARRAY_SIZE()`. (-Wmissing-field-initializers)
* The `xmp_inject_event` regression test had the same warning for the `xmp_event` it defines.
* Every ProWizard loader had the same warning for a `list_head` field in the loader definition. I just commented out the field because nothing uses it.
* Misc. places in the XZ depacker required `/* fall-through */` to silence `-Wimplicit-fallthrough`. The copy of this depacker in the Linux kernel does something similar.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -106,6 +106,12 @@ XMP_TRY_COMPILE(whether compiler understands -Wall,
   int main(void){return 0;}],
   CFLAGS="${CFLAGS} -Wall")
 
+XMP_TRY_COMPILE(whether compiler understands -Wextra,
+  ac_cv_c_flag_w_extra,
+  -Wextra -Wno-unused-parameter -Wno-sign-compare,[
+  int main(void){return 0;}],
+  CFLAGS="${CFLAGS} -Wextra -Wno-unused-parameter -Wno-sign-compare")
+
 XMP_TRY_COMPILE(whether compiler understands -Wwrite-strings,
   ac_cv_c_flag_w_write_strings,
   -Wwrite-strings,[
--- a/src/loaders/mod_load.c
+++ b/src/loaders/mod_load.c
@@ -80,7 +80,6 @@ const struct mod_magic mod_magic[] = {
 	{"FA06", 1, TRACKER_DIGITALTRACKER, 6},	/* Atari Falcon */
 	{"FA08", 1, TRACKER_DIGITALTRACKER, 8},	/* Atari Falcon */
 	{"NSMS", 1, TRACKER_UNKNOWN, 4},	/* in Kingdom.mod */
-	{"", 0}
 };
 
 static int mod_test(HIO_HANDLE *, char *, const int);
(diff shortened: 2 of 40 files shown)


AliceLR:
Going through the regression test CI logs, I found the dev-check build for `linux (ubuntu-18.04)` demonstrates the `-Wunused-result` pedantry for stdio that glibc had at some point I mentioned in the last warning PR. libxmp itself seems to be unaffected but I can add a commit to fix those too when I'm done checking the other logs.

src/loaders/mod_load.c
```diff
@@ -80,7 +80,6 @@ const struct mod_magic mod_magic[] = {
 	{"FA06", 1, TRACKER_DIGITALTRACKER, 6},	/* Atari Falcon */
 	{"FA08", 1, TRACKER_DIGITALTRACKER, 8},	/* Atari Falcon */
 	{"NSMS", 1, TRACKER_UNKNOWN, 4},	/* in Kingdom.mod */
-	{"", 0}
```
sezero, line 83: you could just complete the initializer, but OK

configure.ac
```diff
@@ -106,6 +106,12 @@ XMP_TRY_COMPILE(whether compiler understands -Wall,
   int main(void){return 0;}],
   CFLAGS="${CFLAGS} -Wall")
 
+XMP_TRY_COMPILE(whether compiler understands -Wextra,
+  ac_cv_c_flag_w_extra,
+  -Wextra -Wno-unused-parameter -Wno-sign-compare,[
+  int main(void){return 0;}],
+  CFLAGS="${CFLAGS} -Wextra -Wno-unused-parameter -Wno-sign-compare")
```
sezero, line 113: saving space and extra lines? OK :)

sezero (approved):
This is OK.

configure.ac
AliceLR: The idea was more like if the annoying warnings can't be turned off, just don't add -Wextra at all. `-Wno-sign-compare` is temporary, really—the problem is when I *did* go through looking at places the warning crops up, there are uses where the unsigned side can be pretty arbitrarily large and the signed side intentionally uses negatives to indicate failure states. Untangling them all will probably have to be done gradually.

As for `-Wno-unused-parameter`, it spams out the build process for things that aren't really problems. *Maybe* some of them could be fixed, but most of them are things like `start` which is provided to the format load functions (and only really matters for loaders that can be invoked by other loaders i.e. UMX -> MOD/S3M/XM/IT).

src/loaders/mod_load.c
AliceLR: I didn't really see the point in using sentinel termination for this list when it is static, fixed length, and accessible to every place that uses it. I had to change it anyway, so I went with the option I thought was better. Don't know that it saves any time/space, but it seems conceptually cleaner to me.

Also, I accidentally leave off array sentinels sometimes. :(

configure.ac
sezero: I meant checking individual switches, well, individually - but they being not activated without -Wextra, all is good

PR #473, opened 2021-08-02
Remove stb_vorbis sanity checks that are not required (again).

AliceLR:
This removes more of the libxmp-specific checks and other changes that do absolutely nothing. I don't blame anyone for thinking that these *could* do something because the backtracking for some of these cases is getting absurd.

I'll comment each change with the justification this time instead of dumping it all in the initial post.

edit: OK, each removed check has a comment thread now. As a final note, I checked the commit history for vorbis.c and most of these were added circa 2015 and their commit messages don't point to any reported issue they fixed (this doesn't mean they don't exist). I'm marking this for review but I do want to fuzz it a bit before merge.

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -1115,10 +1115,7 @@ static int compute_codewords(Codebook *c, uint8 *len, int n, uint32 *values)
    memset(available, 0, sizeof(available));
    // find the first entry
    for (k=0; k < n; ++k) if (len[k] < NO_CODE) break;
-
-   // sanity check
-   if (k == n) return (c->sorted_entries == 0);
-
+   if (k == n) { assert(c->sorted_entries == 0); return TRUE; }
    assert(len[k] < 32); // no error return required, code reading lens checks this
    // add to the list
    add_entry(c, 0, k, m++, len[k], values);
@@ -1195,7 +1192,7 @@ static int STBV_CDECL uint32_compare(const void *p, const void *q)
 
 static int include_in_sort(Codebook *c, uint8 len)
 {
-   if (c->sparse) { return (len != NO_CODE); }
+   if (c->sparse) { assert(len != NO_CODE); return TRUE; }
    if (len == NO_CODE) return FALSE;
    if (len > STB_VORBIS_FAST_HUFFMAN_LENGTH) return TRUE;
    return FALSE;
@@ -2139,7 +2136,7 @@ static int residue_decode(vorb *f, Codebook *book, float *target, int offset, in
 
 // n is 1/2 of the blocksize --
 // specification: "Correct per-vector decode length is [n]/2"
-static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int rn, uint8 *do_not_decode)
+static void decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int rn, uint8 *do_not_decode)
 {
    int i,j,pass;
    Residue *r = f->residue_config + rn;
@@ -2299,14 +2296,7 @@ static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int 
                      float *target = residue_buffers[j];
                      int offset = r->begin + pcount * r->part_size;
                      int n = r->part_size;
-                     Codebook *book;
-
-                     /* Sanity check */
-                     if (offset + n >= f->blocksize_1) {
-                        return FALSE;
-                     }
-
-                     book = f->codebooks + b;
+                     Codebook *book = f->codebooks + b;
                      if (!residue_decode(f, book, target, offset, n, rtype))
                         goto done;
                   }
@@ -2315,10 +2305,6 @@ static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int 
          }
          #ifndef STB_VORBIS_DIVIDES_IN_RESIDUE
          ++class_set;
-         /* Sanity check */
-         if (class_set >= part_read) {
-             return FALSE;
-         }
          #endif
       }
    }
@@ -2330,8 +2316,6 @@ static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int 
    temp_free(f,classifications);
    #endif
    temp_alloc_restore(f,temp_alloc_point);
-
-   return TRUE;
 }
 
 
@@ -3320,10 +3304,6 @@ static int vorbis_decode_packet_rest(vorb *f, int *len, Mode *m, int left_start,
             }
 
 #ifdef STB_VORBIS_NO_DEFER_FLOOR
-	    /* Sanity check */
-	    if (n >= f->blocksize_1)
-		return FALSE;
-
             do_floor(f, map, i, n, f->floor_buffers[i], finalY, step2_flag);
 #else
             // defer final floor computation until _after_ residue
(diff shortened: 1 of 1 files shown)


src/depackers/vorbis.c
```diff
@@ -1115,10 +1115,7 @@ static int compute_codewords(Codebook *c, uint8 *len, int n, uint32 *values)
    memset(available, 0, sizeof(available));
    // find the first entry
    for (k=0; k < n; ++k) if (len[k] < NO_CODE) break;
-
-   // sanity check
-   if (k == n) return (c->sorted_entries == 0);
-
+   if (k == n) { assert(c->sorted_entries == 0); return TRUE; }
```
AliceLR, line 1118: The variable of interest here is `c->sorted_entries`. `compute_codewords` is called by `start_decoder`. (line 3886)

## Case 1: ordered
1) `sorted_count` = 0 (line 3856)
2) The following loop scans for values != `NO_CODE` of a certain length; `sorted_count++` when found
3) `c->sorted_entries` = `sorted_count` (line 3864)
4) `c->sorted_entries` will always be 0 if all codes are `NO_CODE`, so `compute_codewords` returns `TRUE`. (line 1120)
Also, it seems like there isn't even a mechanism for `NO_CODE` to get inserted in this case, which makes sense.

## Case 2: unordered, non-sparse.
1) `total` incremented only if a code exists, not when `NO_CODE` is inserted (line 3834)
2) Because `total` is unused, this case ends up being identical to case 1.

## Case 3: unordered, sparse.
1) `total` = 0 (line 3787)
2) `total` incremented only if a code exists, not when `NO_CODE` is inserted (line 3834)
3) side case: sparse can be converted to non-sparse, which will follow the same path as case 1. (line 3839-3850)
4) `sorted_count` = `total` (line 3854)
5) `c->sorted_entries` = `sorted_count` (line 3864)
6) `c->sorted_entries` will always be 0 if all codes are `NO_CODE`, so `compute_codewords` returns `TRUE`. (line 1120)

src/depackers/vorbis.c
```diff
@@ -1195,7 +1192,7 @@ static int STBV_CDECL uint32_compare(const void *p, const void *q)
 
 static int include_in_sort(Codebook *c, uint8 len)
 {
-   if (c->sparse) { return (len != NO_CODE); }
+   if (c->sparse) { assert(len != NO_CODE); return TRUE; }
```
AliceLR, line 1195: 1) Usage of `include_in_sort` at line 1216 is irrelevant to sparse codebooks.
2) Usage of `include_in_sort` later in `compute_sorted_huffman` with `huff_len` is relevant for all codebooks. (line 1235)
3) `huff_len` = `lengths[values[i]]` for sparse codebooks. (line 1234)
4) 0 <= `i` < `c->sorted_entries`. (lines 1227-1233)
5) `compute_sorted_huffman` is called from `start_decoder`. (line 3900)
6) `values` is initialized in `start_decoder` via `compute_codewords` (line 3886)
7) These are initialized by `add_entry`, which for sparse codebooks, inserts them incrementally and only for non-`NO_CODE` lengths. The value will always be an index to that length in the lengths table. (lines 1124, 1147)
8) One value is inserted into values in `compute_codewords` for every non-`NO_CODE` length.
9) `c->sorted_entries` is always equal to the number of non-`NO_CODE` lengths inserted in sparse codebooks (see previous comment).

`len` will never be `NO_CODE` in `include_in_sort` for sparse codebooks, so line 1198 will always return `TRUE`.

src/depackers/vorbis.c
```diff
@@ -2299,14 +2296,7 @@ static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int
...
                      int n = r->part_size;
-                     Codebook *book;
-
-                     /* Sanity check */
-                     if (offset + n >= f->blocksize_1) {
-                        return FALSE;
-                     }
-
-                     book = f->codebooks + b;
+                     Codebook *book = f->codebooks + b;
```
AliceLR, line 2299: I alluded to this check not helping in #466, but here's the reason why:

1) `offset` = `r->begin + pcount * r->part_size`. (line 2300)
2) 0 <= `pcount` < `part_read`. (line 2289)
3) `part_read` = `n_read / r->part_size`. (line 2153)
4) `n_read` = `limit_r_end - limit_r_begin`, both clamped to range [0, `blocksize`]. (lines 2149-2152)
5) `n` (in this particular block) = `r->part_size`. (line 2301)
6) `(part_read - 1)` (maximum value of `pcount`) ` * r->part_size + r->part_size` (value of `n`) == `part_read * r->part_size` <= `n_read` <= `blocksize`, so this check isn't necessary, and worst case may be able to refuse valid offset+length pairs due to off-by-one in the check equality.

src/depackers/vorbis.c
```diff
@@ -2315,10 +2305,6 @@ static int decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int
          }
          #ifndef STB_VORBIS_DIVIDES_IN_RESIDUE
          ++class_set;
-         /* Sanity check */
-         if (class_set >= part_read) {
-             return FALSE;
-         }
```
AliceLR, line 2321: `class_set` will always be less than `part_read`:

1) The containing loop iterates `pcount` = 0, `class_set` = 0; `pcount` < `part_read`. (line 2270)
2) The loop that increments `pcount` will always iterate at least once: (line 2289)
	1) The condition `pcount` < `part_read` is trivially true for the initial check.
	2) `i` < `classwords` is true for the initial check:
		1) `classwords` = `f->codebooks[c].dimensions`. (line 2148)
		2) codebook dimension count of 0 is filtered by `start_decoder` (see below). (line 3802)
3) `class_set` will never be incremented more times than `pcount`, so this check never matters in most cases, and worst case the off-by-one in the check equality may refuse valid cases.

Re: line 3802, there is an exception to this case, which is when the codebook has zero entries. It seems to me that codebooks with zero entries should be an error, or floors/residues referencing them should be, but I haven't seen a place where these get filtered out (and the spec doesn't mention this edge case). If this is an stb_vorbis bug then I want a case that will flag ASan for it.

src/depackers/vorbis.c
```diff
@@ -3320,10 +3304,6 @@ static int vorbis_decode_packet_rest(vorb *f, int *len, Mode *m, int left_start,
             }
 
 #ifdef STB_VORBIS_NO_DEFER_FLOOR
-	    /* Sanity check */
-	    if (n >= f->blocksize_1)
-		return FALSE;
-
```
AliceLR, line 3326: An easy one for a change:

1) `n` = `f->blocksize[m->blockflag]` -> `n` is either `blocksize_0` or `blocksize_1`.
2) `blocksize_0` is always less than `blocksize_1`. (line 3696)
3) This case will never occur or, worst case, will filter valid floor lengths of `blocksize_1` due to off-by-one.

sezero (approved):

AliceLR:
Followup for fuzzing: I fuzzed this branch a bit with ASan and nothing new came up. I also fuzzed stb_vorbis directly (which should exhibit the same bugs as removing these checks) and it never came up. Every job instead aborted on a different issue in `codebook_decode_deinterleave_repeat`. I'll diagnose and report or patch that issue upstream soonish.

PR #475, opened 2021-08-03
lzw.c: replace variadic macro LZWD_ with LZW_DEBUG ifdefs.

sezero:


AliceLR:
From what I can gather, the oldest version of MSVC that does not support variadic macros is MSVC 2003. It seems like this version has good target support and theoretically could be useful for targeting e.g. Windows 95 (though so can old MinGW builds, and unlike MSVC they bother to implement `snprintf` correctly). This is a debug macro, though, so my question is this: is there a legitimate reason to use MSVC 2003 *for development*? The only thing that comes to mind is maybe ReactOS. This question isn't that important, but I am curious since I'd pretty much written off MSVC versions prior to MSVC 2015 due to their bad standards compliance.

More to the point, I think the variadic macro could be avoided by just doing something like what the MED4 loader does:
```C
#ifdef LZW_DEBUG
printf("whatever");
#endif 
```

sezero:
Force-pushed a version to use `LZW_DEBUG` ifdefs: OK?

AliceLR (approved):

Issue #474, opened 2021-08-03
cmake: pkgconfig file is not generated


sezero:
As title says: libxmp.pc isn't generated/installed.

The documentation isn't generated, either.

PR #476, opened 2021-08-03
cmake fixes and improvements:

sezero:
- Use GNUInstallDirs, replace hardcoded install destinations
  with CMAKE_INSTALL_xxxDIR
- Generate and install libxmp.pc (bug #474)
- MSVC now outputs libxmp.dll as it should, not xmp.dll
- MSVC now outputs libxmp-static.lib not xmp-static.lib
- Tweak libxmp-lite.pc installdir

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,14 +52,12 @@ endif()
 
 set(XMP_INSTALLS)
 
-# ============= FULL =============
 if(BUILD_STATIC)
     add_library(xmp_static STATIC ${LIBXMP_SRC_LIST})
     list(APPEND XMP_INSTALLS xmp_static)
     set_target_properties(xmp_static PROPERTIES C_STANDARD 90)
-
-    if(MSVC AND BUILD_SHARED)
-        set_target_properties(xmp_static PROPERTIES OUTPUT_NAME xmp-static)
+    if(MSVC)
+        set_target_properties(xmp_static PROPERTIES OUTPUT_NAME libxmp-static)
     else()
         set_target_properties(xmp_static PROPERTIES OUTPUT_NAME xmp)
     endif()
@@ -82,7 +80,12 @@ endif()
 if(BUILD_SHARED)
     add_library(xmp_shared SHARED ${LIBXMP_SRC_LIST})
     list(APPEND XMP_INSTALLS xmp_shared)
-    set_target_properties(xmp_shared PROPERTIES OUTPUT_NAME xmp C_STANDARD 90)
+    set_target_properties(xmp_shared PROPERTIES C_STANDARD 90)
+    if(MSVC)
+        set_target_properties(xmp_shared PROPERTIES OUTPUT_NAME libxmp)
+    else()
+        set_target_properties(xmp_shared PROPERTIES OUTPUT_NAME xmp)
+    endif()
 
     if(HAVE_GCC OR HAVE_CLANG)
         if(APPLE)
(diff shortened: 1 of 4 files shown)


sezero:
@Wohlstand: please review if you can.

AliceLR (approved):
These changes look sensible to me. (Same disclaimer as usual about my inexperience with CMake.)

CMakeLists.txt
```diff
@@ -151,12 +154,28 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/examples/CMakeLists.txt")
 endif()
 
 
-install(TARGETS ${XMP_INSTALLS}
-        RUNTIME DESTINATION "bin"
-        LIBRARY DESTINATION "lib"
-        ARCHIVE DESTINATION "lib"
-        INCLUDES DESTINATION "include")
+include(GNUInstallDirs)
```
Wohlstand, line 157: Please put include into the top of the file to make an order

Wohlstand, line 157: Please put includes into the top of the file to make an order

lite/CMakeLists.txt
```diff
@@ -145,12 +148,28 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/build_examples.cmake")
 endif()
 
 
-install(TARGETS ${XMP_INSTALLS}
-        RUNTIME DESTINATION "bin"
-        LIBRARY DESTINATION "lib"
-        ARCHIVE DESTINATION "lib"
-        INCLUDES DESTINATION "include")
+include(GNUInstallDirs)
```
Wohlstand, line 151: Please put includes into the top of the file to make an order

CMakeLists.txt
```diff
@@ -151,12 +154,28 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/examples/CMakeLists.txt")
...
 
+if(LIBM_REQUIRED)
+  set(LIBM "-lm")
+endif()
```
Wohlstand, line 161: I suggest to add this line into this place to make an order:
https://github.com/libxmp/libxmp/blob/master/cmake/libxmp-checks.cmake#L118

lite/CMakeLists.txt
```diff
@@ -145,12 +148,28 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/build_examples.cmake")
...
 
+if(LIBM_REQUIRED)
+  set(LIBM "-lm")
+endif()
```
Wohlstand, line 155: I suggest to add this line into this place to make an order:
https://github.com/libxmp/libxmp/blob/master/cmake/libxmp-checks.cmake#L118

Wohlstand (changes requested):
Looks good, however, there are some small tunes needed.

CMakeLists.txt
sezero: I put it where it's needed, but OK will do.

Question: Do we really need `install`ing for MSVC builds?

Wohlstand: Yes, needed, the direct example of use:
- Taking this library and others as ExternalProject
- Making the dedicated hive of libraries and set it as "CMAKE_INSTALL_PREFIX"
- All built libraries will appear in that hive of libraries
- The main project will use these libraries directly

Wohlstand: As an example, my TheXTech is built also for Windows ARM64, uses MSVC, and it does the same as other platforms and toolchains: all libraries were installed into sub-directory of the build directory that got set as "CMAKE_INSTALL_PREFIX", and all these libraries getting easily used by the main project itself.

Wohlstand: Also, the `install()` is useful while crafting various packages, and there is the `cpack` tool that allows you quickly make the binary package of the library if you specify all flags required for certain package type, and the `install()` will be needed

sezero: > Yes, needed, [...]

OK

sezero: > I put it where it's needed, but OK will do.

Well, looking at it more, I think it is good where it is

CMakeLists.txt
sezero: Done this (will force-push shortly.)

lite/CMakeLists.txt
sezero: Done this (will force-push shortly.)

CMakeLists.txt
```diff
@@ -151,12 +154,26 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/examples/CMakeLists.txt")
...
 
+set(prefix ${CMAKE_INSTALL_PREFIX})
+set(exec_prefix "\${prefix}")
+set(includedir "${CMAKE_INSTALL_FULL_INCLUDEDIR}")
+set(libdir     "${CMAKE_INSTALL_FULL_LIBDIR}")
+set(bindir     "${CMAKE_INSTALL_FULL_BINDIR}")
+configure_file("${CMAKE_CURRENT_SOURCE_DIR}/libxmp.pc.in"
+               "${CMAKE_CURRENT_BINARY_DIR}/libxmp.pc" @ONLY
+)
+install(TARGETS ${XMP_INSTALLS}
```
Wohlstand, line 168: I'd added the space between `configure_file()` and `install()` here to make the thing look nicer, however, fine.

PR #478, opened 2021-08-06
use LT_PATH_LD to detect GNU ld, which uses compiler's ld program:

sezero:
extracted LT_PATH_LD and its dependencies with minor modifications
from libtool for use in libxmp and placed as m4/ld.m4, and with an
aclocal.m4 pointing to it.  generating the configure script is the
same as before, simply running 'autoconf' (and _not_ autoreconf .)


AliceLR (approved):
This seems OK to me.

PR #269, opened 2020-12-11
Fix out-of-bounds write in vorbis.c due to missing error check.

AliceLR:
`stb_vorbis_get_frame_short_interleaved` et al. can return -1 when an error state occurs, but `stb_vorbis_decode_{filename,memory}` were treating this like a valid return value, which could cause out-of-bounds writes to the heap on subsequent calls. Fixes #268.

As a side note, since it's 60k I tried compressing the test "OXM" file but it looks like libxmp doesn't handle nested depackers (probably for the best...).

PR #479, opened 2021-08-07
Remove remaining stb_vorbis sanity checks that are not required.

AliceLR:
More of the same as #473, except this time it should be the last one. Like last time, I'm going to stick the explanations in review comments because some of them are long. This reverts a couple of recent patches (#269 and #441) fixing bugs caused by negative return values that were actually libxmp's fault in the first place—at some point the libxmp fork was modified to use Unix-style 0/-1 return values in some places where stb_vorbis consistently used `TRUE`/`FALSE` or lengths.

edit: all major issues have review comments explaining them now. Places that aren't commented are generally related to the `vorbis_pump_first_frame`/`vorbis_finish_frame` issue or are from the aforementioned patches that are no longer necessary.

### DIFF
--- a/src/depackers/vorbis.c
+++ b/src/depackers/vorbis.c
@@ -1303,14 +1303,8 @@ static void compute_bitreverse(int n, uint16 *rev)
 {
    int ld = ilog(n) - 1; // ilog is off-by-one from normal definitions
    int i, n8 = n >> 3;
-   for (i=0; i < n8; ++i) {
-      /* CID 128660 (#1 of 1): Bad bit shift operation (BAD_SHIFT)
-       * large_shift: right shifting by more than 31 bits has undefined
-       * behavior. The shift amount, 32 - ld + 3, is 36.
-       */
-      int s = 32-ld+3;
-      rev[i] = s > 31 ? 0 : (bit_reverse(i) >> s) << 2;
-   }
+   for (i=0; i < n8; ++i)
+      rev[i] = (bit_reverse(i) >> (32-ld+3)) << 2;
 }
 
 static int init_blocksize(vorb *f, int b, int n)
@@ -3341,9 +3335,6 @@ static int vorbis_decode_packet_rest(vorb *f, int *len, Mode *m, int left_start,
       int r;
       uint8 do_not_decode[256];
       int ch = 0;
-
-      memset(do_not_decode, 0, 256);
-
       for (j=0; j < f->channels; ++j) {
          if (map->chan[j].mux == i) {
             if (zero_channel[j]) {
@@ -3549,7 +3540,7 @@ static int vorbis_pump_first_frame(stb_vorbis *f)
    int len, right, left, res;
    res = vorbis_decode_packet(f, &len, &left, &right);
    if (res)
-      res = vorbis_finish_frame(f, len, left, right);
+      vorbis_finish_frame(f, len, left, right);
    return res;
 }
 
@@ -3620,7 +3611,6 @@ static int start_decoder(vorb *f)
    uint8 header[6], x,y;
    int len,i,j,k, max_submaps = 0;
    int longest_floorlist=0;
-   int val;
 
    // first page, first packet
    f->first_decode = TRUE;
@@ -3655,9 +3645,8 @@ static int start_decoder(vorb *f)
    if (!vorbis_validate(header))                    return error(f, VORBIS_invalid_first_page);
    // vorbis_version
    if (get32(f) != 0)                               return error(f, VORBIS_invalid_first_page);
-   val = get8(f); if (val == 0)                     return error(f, VORBIS_invalid_first_page);
-   if (val > STB_VORBIS_MAX_CHANNELS)               return error(f, VORBIS_too_many_channels);
-   f->channels = val;
+   f->channels = get8(f); if (!f->channels)         return error(f, VORBIS_invalid_first_page);
+   if (f->channels > STB_VORBIS_MAX_CHANNELS)       return error(f, VORBIS_too_many_channels);
    f->sample_rate = get32(f); if (!f->sample_rate)  return error(f, VORBIS_invalid_first_page);
    get32(f); // bitrate_maximum
    get32(f); // bitrate_nominal
@@ -4565,12 +4554,6 @@ int stb_vorbis_decode_frame_pushdata(
 
    // success!
    len = vorbis_finish_frame(f, len, left, right);
-
-   /* Sanity check */
-   if (len < 0) {
-      return -1;
-   }
-
    for (i=0; i < f->channels; ++i)
       f->outputs[i] = f->channel_buffers[i] + left;
 
@@ -4975,8 +4958,7 @@ int stb_vorbis_seek_frame(stb_vorbis *f, unsigned int sample_number)
          return 1; // the next frame will contain the sample
       } else if (f->current_loc + frame_samples + max_frame_samples > sample_number) {
          // there's a chance the frame after this could contain the sample
-         if (vorbis_pump_first_frame(f) < 0)
-            return -1;
+         vorbis_pump_first_frame(f);
       } else {
          // this frame is too early to be relevant
          f->current_loc += frame_samples;
@@ -4998,8 +4980,7 @@ int stb_vorbis_seek(stb_vorbis *f, unsigned int sample_number)
    if (sample_number != f->current_loc) {
       int n;
       uint32 frame_start = f->current_loc;
-      if (stb_vorbis_get_frame_float(f, &n, NULL) < 0)
-          return error(f, VORBIS_seek_failed);
+      stb_vorbis_get_frame_float(f, &n, NULL);
       assert(sample_number > frame_start);
       assert(f->channel_buffer_start + (int) (sample_number-frame_start) <= f->channel_buffer_end);
       f->channel_buffer_start += (sample_number - frame_start);
@@ -5110,18 +5091,7 @@ int stb_vorbis_get_frame_float(stb_vorbis *f, int *channels, float ***output)
       return 0;
    }
 
-   /* Sanity check */
-   if (len > f->blocksize_1) {
-      return -1;
-   }
-
    len = vorbis_finish_frame(f, len, left, right);
-
-   /* Sanity check */
-   if (len < 0) {
-      return -1;
-   }
-
    for (i=0; i < f->channels; ++i)
       f->outputs[i] = f->channel_buffers[i] + left;
 
@@ -5147,8 +5117,7 @@ stb_vorbis * stb_vorbis_open_file_section(FILE *file, int close_on_free, int *er
       f = vorbis_alloc(&p);
       if (f) {
          *f = p;
-         if (vorbis_pump_first_frame(f))
-            return NULL;
+         vorbis_pump_first_frame(f);
          return f;
       }
    }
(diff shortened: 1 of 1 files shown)


src/depackers/vorbis.c
```diff
@@ -1303,14 +1303,8 @@ static void compute_bitreverse(int n, uint16 *rev)
...
-   for (i=0; i < n8; ++i) {
-      /* CID 128660 (#1 of 1): Bad bit shift operation (BAD_SHIFT)
-       * large_shift: right shifting by more than 31 bits has undefined
-       * behavior. The shift amount, 32 - ld + 3, is 36.
-       */
-      int s = 32-ld+3;
-      rev[i] = s > 31 ? 0 : (bit_reverse(i) >> s) << 2;
-   }
+   for (i=0; i < n8; ++i)
+      rev[i] = (bit_reverse(i) >> (32-ld+3)) << 2;
```
AliceLR, line 1307: 1) Out of bounds shift occurs when `ld` is 3 or less, resulting in a shift of 32 or greater.
2) `ld` = `ilog(n) - 1` = log₂(`n`) or -1 (if `n`<=0).
3) Side note: cited shift value 36 would occur with `ld`=-1, presumably from an `n` of 0.
4) `compute_bitreverse` is only ever called from `init_blocksize` (line 1329). The `n` parameter to `compute_bitreverse` is the `n` parameter to `init_blocksize` and is never modified.
5) `init_blocksize` is called once each for both blocksizes in `start_decoder` (line 4195).
6) The blocksizes are calculated from stored log₂() values in the Vorbis stream, also in `start_decoder` (line 3670). These log values are bounded between 6 and 13, meaning the blocksizes will never be less than 64 or greater than 8192.
7) `ilog(64)` = `(5 + log2_4[64 >> 5])` = `(5 + log2_4[2])` = 7, so `ld` = 6 is the minimum value for `ld`.

This hack was added in this patch: https://github.com/libxmp/libxmp/commit/db5cad087d0baf36c5368bdfb9f2aa3ccf02f8b4.

The same analysis holds up for the copy of stb_vorbis that was originally patched, meaning this fix never actually did anything and was just to shut up a static analyzer false positive.

src/depackers/vorbis.c
```diff
@@ -3341,9 +3335,6 @@ static int vorbis_decode_packet_rest(vorb *f, int *len, Mode *m, int left_start,
       int r;
       uint8 do_not_decode[256];
       int ch = 0;
-
-      memset(do_not_decode, 0, 256);
-
```
AliceLR, line 3346: 1) `do_not_decode` and `ch` are declared in `vorbis_decode_initial` (line 3342). `ch` has an initial value of 0.
2) A loop on `f->channels` determines which channel to decode according to a map. Before `ch` is incremented, `do_no_decode[ch]` is initialized.
3) `decode_residue` is called with `ch` and `do_not_decode` being passed to parameters with identical names (line 3360).
4) Usages of `do_not_decode` in `decode_residue`:
    1) Line 2161: iterated with `i` which is bounded by `ch`.
    2) Line 2166: iterated with `j` which is bounded by `ch`.
    3) Line 2270: iterated with `j` which is bounded by `ch`.
    4) Line 2288: iterated with `j` which is bounded by `ch`.
5) Indices of `do_not_decode` >= `ch` do not need to be initialized and indices <`ch` already are being initialized.

Introduced by this patch: https://github.com/libxmp/libxmp/commit/3b6214b3b9e138eba14d0725bd8e99853c7b1786.

src/depackers/vorbis.c
```diff
@@ -3549,7 +3540,7 @@ static int vorbis_pump_first_frame(stb_vorbis *f)
    int len, right, left, res;
    res = vorbis_decode_packet(f, &len, &left, &right);
    if (res)
-      res = vorbis_finish_frame(f, len, left, right);
+      vorbis_finish_frame(f, len, left, right);
```
AliceLR, line 3543: An old libxmp-specific hack (https://github.com/libxmp/libxmp/commit/fb4b9dda1deb864e64ed1a862767870b5eedbd90) was removed in #408 that could cause `vorbis_finish_frame` to return -1. This was obsoleted by an upstream check that was added that returns 0 instead.

Based on upstream usage, `vorbis_pump_first_frame` should not be returning the return value of `vorbis_finish_frame` (which is the number of samples written), but instead a `TRUE` or `FALSE` value. To add to the confusion, at some point the libxmp copy of stb_vorbis was modified to use 0/-1 return values, so a later check in deadcode incorrectly returns NULL when the return value of this is NOT zero. I've just reverted this entire mess. This includes one of the two checks in `stb_vorbis_get_frame_float` that have caused a lot of trouble elsewhere.

src/depackers/vorbis.c
```diff
@@ -3655,9 +3645,8 @@ static int start_decoder(vorb *f)
    if (!vorbis_validate(header))                    return error(f, VORBIS_invalid_first_page);
    // vorbis_version
    if (get32(f) != 0)                               return error(f, VORBIS_invalid_first_page);
-   val = get8(f); if (val == 0)                     return error(f, VORBIS_invalid_first_page);
-   if (val > STB_VORBIS_MAX_CHANNELS)               return error(f, VORBIS_too_many_channels);
-   f->channels = val;
+   f->channels = get8(f); if (!f->channels)         return error(f, VORBIS_invalid_first_page);
+   if (f->channels > STB_VORBIS_MAX_CHANNELS)       return error(f, VORBIS_too_many_channels);
```
AliceLR, line 3649: This one is interesting because at first glance it looks like it does *absolutely nothing of value*, but actually did fix invalid free bugs in the old stb_vorbis.

1) `start_decoder` is only called from `stb_vorbis_open_memory` (line 5199) on a stack-allocated `stb_vorbis` that is zero initialized by `vorbis_init`.
2) `start_decoder` returns 0 (line 3658-9) due to a zero or too large channel count. Assume that the channels count is written directly to `f->channels` like upstream does it.
3) Quick check: `error`, defined at line 935, does in fact always return 0.
4) On failure, `stb_vorbis_open_memory` sets `*error`, calls `vorbis_deinit`, and returns NULL.
5) In `vorbis_deinit` at line 4306, `p->channels` is iterated on. This is combined with a check to prevent `i` from ever being >=`STB_VORBIS_MAX_CHANNELS` so no frees are performed on out-of-bounds array positions. The reason for the libxmp hack is that this second check was missing in the old stb_vorbis.
6) Since there's an upstream safety check here, the old check can be removed.

Bonus: `stb_vorbis_open_pushdata` and `stb_vorbis_open_file_section`, the other places that use `start_decoder` (deadcode in our copy), also behave in similar ways, so the only thing using `f->channels` for this error state is that loop in `vorbis_deinit`.

This check was added here: https://github.com/libxmp/libxmp/commit/957e4a61d6f3798937dbca69e2e3716b111563a8.

src/depackers/vorbis.c
```diff
@@ -5110,18 +5091,7 @@ int stb_vorbis_get_frame_float(stb_vorbis *f, int *channels, float ***output)
       return 0;
    }
 
-   /* Sanity check */
-   if (len > f->blocksize_1) {
-      return -1;
-   }
-
```
AliceLR, line 5117: 1) `&len` passed to `vorbis_decode_packet`
2) `vorbis_decode_packet` first calls `vorbis_decode_initial`, receiving return values to `mode`, `p_left`, `left_end`, `p_right`, and `right_end`.
3) `len` is then passed to `vorbis_decode_packet_rest` along with the window bounds.

## `vorbis_decode_initial`

All window bounds are >= 0, <= `blocksize_1`.

### Case 1: `blockflag` == 0 or `blockflag` == 1 and the relevant `prev` or `next` flag == 1.
1) `n` = `blocksize_0` (line 3191) or `blocksize_1` (line 3186).
2) `*p_left_start` (aka `*p_left`) = 0 (line 3201).
3) `*p_left_end` (aka `*left_end`) = `window_center` = `n`/2 (line 3202).
4) `*p_right_start` (aka `*p_right`) = `window_center` = `n`/2 (line 3208).
5) `*p_right_end` (aka `*right_end`) = `n` (line 3209).
6) Since `blocksize_0` <= `blocksize_1` (line 3674), all values are less than or equal to `blocksize_1`.

### Case 2: `blockflag` == 1, `prev` == 0
1) `*p_left_start` (aka `*p_left`) = (`n` - `blocksize_0`) / 4 (line 3198) is trivially less than `blocksize_1`.
2) `*p_left_end` (aka `*left_end`) = (`n` + `blocksize_0`) / 4 (line 3199) caps out at `blocksize_1`/2.

### Case 3: `blockflag` == `, `next` == 0
1) `*p_right_start` (aka `*p_right`) = (`n`*3 - `blocksize_0`) / 4 (line 3205) is trivially less than `blocksize_1`.
2) `*p_right_end` (aka `*right_end`) = (`n`*3 + `blocksize_0`) / 4 (line 3206) caps out at `blocksize_1`.

## `vorbis_decode_packet_rest`.

`*len` is assigned in two ways. `vorbis_decode_packet_rest` returns `TRUE` after either case. `vorbis_decode_packet_rest` only returns in one other place, which is an error case where the value of `*len` won't matter.

### Normal case.

1) `*len` = `right_end` (line 3481), so <=`blocksize_1`.

### End of current stream.

If the window would cause `f->current_loc` to advance beyond `current_end`, some special handling occurs.

1) The check that triggers this: `current_end` < `f->current_loc` + (`right_end` - `left_start`).
2) If `current_end` < `f->current_loc`: `*len` is set to 0 (line 3459) + `left_start` (line 3463).
3) Otherwise: `*len` is `current_end` - `f->current_loc` (line 3461) + `left_start` (line 3463). Since (`right_end` - `left_start`) is at most `blocksize_1`, this difference shouldn't ever be larger than `blocksize_1`.
4) `*len` is immediately bounded to `right_end` (line 3464), and is thus never larger than `blocksize_1` (even if it somehow was before).
5) `*len` is always >=0, <=`blocksize_1`.

## `stb_vorbis_get_frame_float`

`*len` is always >=0, <=`blocksize_1` in cases where `vorbis_decode_packet` returns success, per above. This check (and all of the hacks that had to be added as a result of it and the other check) can now be removed because `stb_vorbis_get_frame_float` no longer returns -1 values added by libxmp-specific checks.

Check was added here: https://github.com/libxmp/libxmp/commit/a1c922ad54ecc0ba4468ea60a8db233038daddd5.
Modifications that had to be added because of negative return values from `stb_vorbis_get_frame_float`:
* https://github.com/libxmp/libxmp/commit/b4a1d365324e8171b5ed4ff91711828d0cb0d096
* https://github.com/libxmp/libxmp/commit/d815172614f0c6c5ff89f4eb06098ffcdac58349

sezero: AFAICS, this doesn't revert https://github.com/libxmp/libxmp/commit/d815172614f0c6c5ff89f4eb06098ffcdac58349 in its entirety.

AliceLR: I probably should have detailed this better, but as far as I can tell this shouldn't return -1 anymore, so both patches are no longer necessary:

* 5106: returns `error()` -> 0
* 5110: returns 0
* 5115: removed (the hardest part of this to wrap my brain around to prove :( )
* 5122: removed (`vorbis_finish_frame` never returns <0)
* 5133: returns `len`; 0 <= `len` <= `blocksize_1`

AliceLR: (I wish stb_vorbis would use unsigned types for these return codes when it doesn't use negative values for errors, so this kind of ambiguity wouldn't exist in the first place.)

sezero (approved):
I wish we had testcases for those old sanity check additions.

Go ahead and apply this.

PR #480, opened 2021-08-25
Fixed CMake build

Wohlstand:
The quick fix of the error that causes next problems:
- resulted builds are can't be linked (the whole main part is absent and won't be linked)
- it will fail with "no sources at the target" when both Prowizard and de-packers disabled


Wohlstand:
Wait... seems there is my local fault, I had an old copies of `libxmp-sources.cmake`, damn

PR #483, opened 2021-09-23
Implement Ultra Tracker tone portamento to replace broken hack.

AliceLR:
Ultra Tracker uses persistent tone portamento instead of MOD-style portamento. Instead of properly supporting this, every player  I've checked either tries to unroll it (libxmp, OpenMPT) or treats it like MOD portamento (MikMod, libmodplug). It's not clear where this behavior originates from but it's bad and needs to be replaced. As it turns out, libxmp already supports persistent tone portamento, so this isn't difficult.

Fixes playback of this test module: [PORTA.ULT.zip](https://github.com/libxmp/libxmp/files/7221305/PORTA.ULT.zip). Found while reviewing a MikMod pull request that adds a similar behavior https://github.com/sezero/mikmod/pull/40.

~~I'm marking this as a draft for now since I haven't verified libxmp's persistent portamento effect matches what UltraTracker expects for varying speeds, interrupting ongoing portamento, etc. I also need to add a regression test once I have a more thorough test module.~~ edit: there were enough minor quirks with Ultra Tracker's tone portamento that I just added a new ULT tone portamento effect instead of using the regular one.


AliceLR:
Marking this ready for review, but prior to merge, it should probably be compared against real Ultra Tracker modules and not just my test modules.

sezero:
@sagamusix, @cmatsuoka: Please have a look.

AliceLR:
Two Ultra Tracker modules I was able to find that rely on continuous tone portamento:

[the earth's last defender theme.ult](http://modland.com/pub/modules/Ultratracker/Killer%20Fish/the%20earth%27s%20last%20defender%20theme.ult). Affected portamento occurs between orders 9 and 10.
[consolidation.ult](http://modland.com/pub/modules/Ultratracker/LaLa/consolidation.ult). Affected portamento occurs between orders 23 and 24.

"The Earth's Last Defender Theme" also reveals libxmp suffers from finetune bugs like MikMod did (in this case, the field is not handled whatsoever). I'll go ahead and fix that while I'm at it, I guess.

(Another very minor issue I noticed that I'll patch is that in two places in the loader, the Ultra Tracker format version is written as "V000x" instead of "V00x". A quick look at any Ultra Tracker file in a hex editor confirms there are two 0s, not three.)

AliceLR:
I went ahead and just fixed the version numbers. The finetune issue is significant enough that it needs a separate patch (if it gets addressed at all). It seems that Ultra Tracker finetune is added directly to the final sample rate after octave/note calculations, but I haven't confirmed this yet. This can be reviewed as-is.

src/loaders/ult_load.c
```diff
@@ -151,18 +148,13 @@ static int ult_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	uih.sizeend = hio_read32l(f);
 	uih.volume = hio_read8(f);
 	uih.bidiloop = hio_read8(f);
+	uih.c2spd = (ver >= 4) ? hio_read16l(f) : 0; /* Incorrect in ult_form.txt */
```
sezero, line 151: Where is c2spd handled for `ver < 4` cases?

AliceLR: The Ultra Tracker format does not store c2spd prior to V004, so it's always 8363Hz. This variable is only used with `libxmp_c2spd_to_note` later in the loader to set instrument transpose and finetune when `ver > 3` (not added by me, I prefer `>=` to  `>` for version checks). The default for those is presumably 0, which should be equivalent to 8363Hz.

sezero: OK.

sezero:
>  This can be reviewed as-is.

This is OK by me.  @sagamusix ?

sagamusix:
Sorry, I don't really have any time to review this in-depth at the moment.

sezero (approved):

Issue #357, opened 2021-06-03
S3M: It's possible to tell GUS and SB apart!


sagamusix:
Today I made an amazing discovery while looking at some S3Ms in a hex editor: It is in fact possible to tell whether an S3M file was last saved with the GUS driver loaded in ST3, or with the SB driver: In the sample headers, there are some reserved / internal values, in particular the `Int:Gp` field that contains the sample address in GUS memory.

- If the file was last saved with a SB driver loaded, this value is 0 for empty sample slots and 1 otherwise.
- If the file was last saved with the GUS driver loaded, this value is 0 for empty sample slots, and for other sample slots it's some value greater than or equal to 1 (the lowest address).
- Early versions of ScreamTracker 3.00 didn't have GUS support, so in S3M files saved with those versions all of those `Int:Gp` values are 0, which is okay because that would indicate a SoundBlaster driver anyway (e.g. party.s3m by Skaven - which was used in The Party '92 Invitation which indeed only supports SB output as well). However there are also *some* ScreamTracker 3.00 modules with the correct GUS pointer values in them, so this feature was added before the ScreamTracker 3.01 release.

Luckily, re-saving the same file in different ST3 sessions with GUS or SB loaded updates to values properly, so if you first save a file with the GUS driver loaded you get the GUS pointers in the file, and if you later resave it with the SB driver loaded all those pointer values become 0 and 1 again. So you get a pretty reliable fingerprint of whether the module is inteded to use GUS quirks or SB quirks.

As I recall, xmp already supports some GUS-specific quirks for S3M. By applying this fingerprint detection, those quirks could be disabled for S3Ms that were detected to be saved with the SB drivers loaded.

sagamusix:
Note that the `Int:Gp` value being 0 for all samples despite sample data being present can also be used for fingerprinting other software disguising as ST3.20 (but not for ST3.00 due to early versions lacking GUS support, as explained above). For example there is some PSM to S3M conversion tool that otherwise disguises perfectly as ST3, but can still be identified due to `Int:Gp` being 0 for all samples.

AliceLR:
Slightly relatedly, I updated my tools to dump S3Ms, and I noticed classic MPT modules seem to save with `Cwt/v`=0x1320 and `Int:Gp`=0 for all instruments—is this reliable fingerprinting? (libxmp currently doesn't try to detect classic MPT, but it seems like it should all things considered.)

AliceLR:
[...]

PR #485, opened 2021-10-03
Fix >1MB S3M modules relying on the sample segment high byte.

AliceLR:
Found via running checks for #357. libxmp completely ignores the sample segment high byte in S3Ms, leading to large S3Ms sounding very bad.

[Blatant example](http://modland.com/pub/modules/Screamtracker%203/DJ%20Meas/sunday%20train%20ride.s3m), though ModLand has at least 39 of these (and Mod Archive probably has far more).

No test because I don't think Claudio would appreciate me adding a 2.5 MB module to the repository. :-)

### DIFF
--- a/src/loaders/s3m.h
+++ b/src/loaders/s3m.h
@@ -82,7 +82,8 @@ struct s3m_file_header {
 };
 
 struct s3m_instrument_header {
-	uint8 dosname[13];	/* DOS file name */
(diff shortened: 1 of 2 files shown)


sezero (approved):

sezero:
Does libmikmod do this right?

AliceLR:
> Does libmikmod do this right?

Surprisingly, yes!

PR #486, opened 2021-10-03
Fix ASYLUM Music Format restart byte.

AliceLR:
The ASYLUM Music Format from the Crusader games, being an `8CHN` MOD clone, supports a restart byte. This isn't necessarily obvious due to the nonexistent documentation on the format. The restart byte is stored between the order count and order table, just like a NoiseTracker or FastTracker MOD.

Only the Todd Parsons credited AMFs on ModLand rely on this. libxmp's loader test module is fortunately one of these, so this doesn't need a new test.

To verify: `xmp -l test-dev/data/m/m07.amf` and let it repeat. A channel will hang for two orders because it's actually supposed to restart to order 2, where the appropriate `EC0` note cut can be found.


sezero (approved):

sezero:
libmikmod needs fixing too, as far as I can tell.

AliceLR:
I have a patch in-progress for this in MikMod (and for the `TDZ#`/`LARD`/`NSMS` MOD magics as well). Holding off on splitting and PRing these until I've figured out what the deal with the `FA##` MOD magics are and if they're worth detecting. Another interesting thing is it looks like it does not support the restart byte for MODs at all, but properly supporting it for `M.K.` MODs is tricky enough I don't want to touch it right now.

PR #489, opened 2021-10-10
Allow xmp_smix_play_* to play key off, cut, and fade events.

### LINKED ISSUE #106: Unable to Cut Looped SMIX Instrument
```c
xmp_smix_play_instrument(ctx, 3,          60, 0, 0); // Instrument #3 contains a loop.
sleep(10);
xmp_smix_play_instrument(ctx, 3, XMP_KEY_CUT, 0, 0);
```

The above code does not work as expected, as pitches passed to `xmp_smix_play_instrument` are incremented by one.  Currently, there is no clean way to cut a looped instrument.  (If I am wrong, please let me know!)

AliceLR:
Fixes #106. This needs better testing and is probably not 100% correct currently.

### DIFF
--- a/src/read_event.c
+++ b/src/read_event.c
@@ -1540,7 +1540,7 @@ static int read_event_smix(struct context_data *ctx, struct xmp_event *e, int ch
 	struct xmp_module *mod = &m->mod;
 	struct channel_data *xc = &p->xc_data[chn];
 	struct xmp_subinstrument *sub;
-	int is_smix_ins;
+	struct xmp_instrument *xxi;
 	int ins, note, transp, smp;
 
 	xc->flags = 0;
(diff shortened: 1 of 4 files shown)


AliceLR:
After some testing against my modified xmp-cli and a few tweaks, this works now roughly as it ought to. Marking this ready for review.

sezero:
This patch does not apply cleanly anymore - needs a rebase.

PING @cmatsuoka for review?

sezero (approved):
This is OK by me. If @cmatsuoka and/or @NoSuck doesn't respond, we can merge after about a week.

cmatsuoka (approved):
The smix_* calls are pretty much a bonus feature that don't (or shouldn't) affect module replaying, so it adding functionality there should be harmless.

sezero:
OK then, let's get this in.

PR #491, opened 2021-10-24
Fix Digital Symphony effects and pattern loading issues.

### LINKED ISSUE #347: Digital Symphony findings
As mentioned in a PR, I have just finished writing my own [feature-complete Digital Symphony loader](https://github.com/OpenMPT/openmpt/blob/master/soundlib/Load_dsym.cpp) for OpenMPT.

Based on a quick glance, the xmp loader needs to address at least the following issues to be able to load more files (and play them):
- Missing support for sigma-delta compression. Feel free to take the decoder from OpenMPT.
- Missing support for several other non-compressed sample formats that xmp already supports, so adding them should be trivial.
- Several pattern commands are not imported even though xmp should be able to support them (e.g. panning). I couldn't find any real-world modules using effects 0x2B and 0x32 yet but they are implemented in OpenMPT the way I think they should work.
- Suggested sanity checks: bail out early if number of channels is 0, or if there are more than 4096 tracks / orders.
- Supposedly patterns / tracks are grouped into chunks of 2000 tracks (i.e. chunks of 512000 uncompressed bytes), but I have no modules to verify this.
- There's two files with a questionable LZW stream [this collection](https://archive.org/details/zenta-digital-symphony-collection-102) (FULLEFFECT and NARCOSIS) that my code originally didn't load; xmp also doesn't load them but I didn't verify whether it's for the same reason. TimPlayer's LZW routines get these files wrong, too. They require a special case to load correctly to not increase the bit width once the decompressed data has reached the expected size (the last insertion into the dictionary causes it to grow to 512 entries, so the bit width should be increased from 9 to 10, but if you do that, the next thing you read will be 769 instead of the expected 257). I have added that special handling to OpenMPT and the files load fine now; and while it smells like a bug in the LZW encoder to me, all the other 3000+ files are unaffected by the fix, so I think it's safe.

AliceLR:
Fixes #347. (Unless I missed something.)

* Digital Symphony stores patterns in blocks of 2000. libxmp previously did not handle this quirk.

* Implemented !Tracker effect 16 Line Jump, which is also used by Digital Symphony. This effect acts like Pattern Break, except it jumps into the current pattern instead of the next pattern. It can also be used in conjunction with Position Jump and Pattern Break to change their target row. This effect required minor flow hacks to work without clobbering running time.

* Fixed a scan bug related to `FX_S3M_TEMPO` where tempo values under 32 would be ignored despite them being higher than the minimum allowed tempo.

* Fixed a memory leak in the Digital Symphony loader.

* Digital Symphony effects added:
  - Implemented 1F Invert Loop. This required a new quirk as invert loop is implemented in a way that assumes QUIRK_PROTRACK may be changed by xmp_set_player.
  - Implemented 2A Volume Slide + Fine Slide Down.
  - Implemented 2B Line Jump (same as !Tracker Line Jump).
  - Implemented 30 Set Stereo.
  - Implemented 32 Unset Sample Repeat. This required switching Digital Symphony modules to IT sustain loops instead of regular loops. This also required adding support for sustain loops to invert loop.

* Digital Symphony effects fixed:
  - Portamento effects 01 02 21 and 22 were incorrectly using effect memory.
  - Fine portamento effects 11 12 1A and 1B were being converted to Exx instead of their equivalent full byte parameter effects.
  - Vibrato and tremolo effects 04 06 07 now update on tick 0.
  - The highest 3 bits of 09 Set Sample Offset are now translated to `FX_HIOFFSET` instead of being ignored.
  - 0B Position Jump uses a hex parameter in Digital Symphony, but libxmp was treating it as decimal.
  - 0F Set Speed now ignores parameter 0 and clamps values >255.
  - 2F Set Tempo now only ignores parameter 0, and clamps other valid tempos instead of ignoring them.
  - Extended effects 16 1C 1D and 1E (which libxmp currently does not have dedicated effects for) are now bounded with a clamp instead of a modulo, which makes them slightly less wrong.

TODO:
- [x] Better Digital Symphony effects coverage should be possible with existing modules. ~~I can't verify any of them in SymphPlay or the Digital Symphony demo though, since neither work in emulation.~~ Nevermind, this is possible now, so a custom module will be possible (with some extra effort since the demo can't save).
- [x] Effect 32 Unset Sample Repeat ~~would be reasonable to implement but I have no modules that use it (yet).~~ is fixed.
- [x] Need `test_effect_line_jump` or similar.

edit: rewrote the changes list to better match the squashed commit message. All of the same information should still be there.
edit: updated for `QUIRK_INVLOOP` being added back in.


AliceLR:
Now to figure out how to untangle this branch so Unset Sample Repeat can use sustain loops instead...

AliceLR:
OK, this should now be about where it was before, except that Unset Sample Repeat is now implemented with sustain loops instead of an ugly hack. I also went ahead and implemented a [sigma-delta sample compressor](https://gist.github.com/AliceLR/f74052999dde1db3a7dcda86c324b138#file-dsymgen-cpp) for my Digital Symphony module writer, which significantly reduced the size of plastic.dsym even compared to gzip.

sezero:
How far is this from being complete?

AliceLR:
I think this just needs extra tests, which I've been slow about implementing due to depacker junk and IMO fixing IT bugs being a higher priority.

AliceLR:
Still working on the effects test module. It turns out there's some nuance to how line jump works that's made it annoying to get right, but the following model *seems* correct for Digital Symphony. I haven't tested how !Tracker handles these.

* `0b` Position Jump, `0d` Pattern Break, and `2b` Line Jump share two variables I'll call `order_target` and `line_target`. In libxmp these are roughly `flow_control::jump` and `flow_control::jumpline`.
* At the start of an order:
  `order_target` := -1
  `line_target` := 0
* Relevant effect behaviors:
  * `0b` Position Jump:
    `order_target` := `param`
    `line_target` := 0
  * `0d` Pattern Break:
    if(`order_target` < 0) {
        `order_target` := `current_order` + 1
    }
    `line_target` := ((`param` & 0xff) < 0x40 ? `param` & 0xff : 0)
  * `2b` Line Jump
    if(`order_target` < 0) {
      `order_target` := `current_order`
    }
    `line_target` := (`param` < 0x40 ? `param` & 0xff : 0)
* Observed combinations of these effects on the same row that seems to support the above:
  * `0b` followed by `0d`: jumps to order specified by position jump, row specified by pattern break.
  * `0b` followed by `2b`: jumps to order specified by position jump, row specified by line jump.
  * `0d` followed by `0b`: jumps to order specified by position jump, row 0.
  * `0d` followed by `2b`: jumps to `current_order`+1, row specified by line jump.
  * `2b` followed by `0b`: jumps to order specified by position jump, row 0.
  * `2b` followed by `0d`: jumps to `current order`, row specified by pattern break.
* Loops seem to take precedence over all jump target handling.

Since this needed to be implemented with pattern jumps I had to add an extra flow variable so libxmp doesn't clobber the current player time with scan data.

AliceLR:
OK, I *think* this is fine now. I've squashed it but my original commit history is here: https://github.com/libxmp/libxmp/compare/master...AliceLR:sym-fix-effects-unsquashed

I also updated the initial post to better reflect the new squashed commit, since I missed some things before.

src/effects.c
```diff
@@ -470,7 +470,9 @@ void libxmp_process_fx(struct context_data *ctx, struct channel_data *xc, int ch
 		case EX_PATT_DELAY:	/* Pattern delay */
 			goto fx_patt_delay;
 		case EX_INVLOOP:	/* Invert loop / funk repeat */
-			xc->invloop.speed = fxp;
+			if (HAS_QUIRK(QUIRK_PROTRACK)) {
```
sezero, line 473: Is this correct?

sezero: .. especially with the lite version?

AliceLR: I missed this, but there is a way for this quirk to be changed mid-playback (`xmp_set_player` with `XMP_PLAYER_MODE`) which if changed at a bad time could leave invert loop stuck on. I don't know *why* someone would do that but it's allowed by the API and probably the reason this was done the old way... ugh.

AliceLR: I'll change it back to a quirk. After reviewing `QUIRK_MODRNG`, it seems safe to place `QUIRK_INVLOOP` in its position. It was the old version of `PERIOD_MODRNG` and is only still referenced in two dead loaders. Likewise, there was a `QUIRK_LINEAR` which is now `PERIOD_LINEAR`.

AliceLR: OK, reverted.

sezero: OK then. AFAICS, the playback result is identical.

Let's get this in.

sezero (approved):

PR #495, opened 2021-10-29
hio_seek should reset the HIO_HANDLE stream error status on success.

AliceLR:
See https://github.com/libxmp/libxmp/pull/493#discussion_r738902930. `hio_seek` doesn't clear the error flag (which also stores the `EOF` state), which makes it semantically different from `fseek`, which is guaranteed to clear the `EOF` state. The result is that hitting `EOF` followed by an `hio_seek` will result in `hio_eof` returning 0 and `hio_error` returning `EOF`.

The fix is to just clear the error state on a successful `hio_seek` if it's `EOF`.

### DIFF
--- a/src/hio.c
+++ b/src/hio.c
@@ -287,17 +287,26 @@ int hio_seek(HIO_HANDLE *h, long offset, int whence)
 		if (ret < 0) {
 			h->error = errno;
 		}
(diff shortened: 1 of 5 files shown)


sezero (approved):
OK

PR #496, opened 2021-11-02
Don't increment voice position by step value at loop/tick end.

AliceLR:
Since commit b7d6a6e, libxmp has added an extra step to voice positions if they are close to or at the end of their loop on a tick boundary. This results in samples with loops being very slightly too fast. This patch corrects this behavior so samples with loops play at the correct speed.

The issue can be demonstrated with two looping samples, where one sample is two inverted copies of the other sample. When played back at the same pitch, they should cancel entirely and be silent, but this extra step causes them to gradually diverge instead.

Test module: [sample_sync.it.zip](https://github.com/libxmp/libxmp/files/7460792/sample_sync.it.zip). This is slightly different from the one linked in a different issue to force it to diverge more quickly without this patch. With this patch, a slight buzz will play with cubic interpolation due to an unrelated bug.

Apologies for the very large diff, but I feel like this patch was necessary :(. The vast majority is mixer test data files where the sample positions drifted by small amounts due to this bug. The important part of the diff is 3 lines long in mixer.c.


sezero:
i686-mingw tests are still failing..

sezero:
`make check` fails on my i686-linux, too:
```
$ make check
CC test/test.o
CC src/md5.o
LD test/libxmp-test
cd test; LD_LIBRARY_PATH=../lib DYLD_LIBRARY_PATH=../lib LIBRARY_PATH=../lib:$LIBRARY_PATH PATH=$PATH:../lib ./libxmp-test
Testing ................................................................................................................................................................................................................................................rendering error
 fail
make: *** [check] Error 1
```

AliceLR:
> `make check` fails on my i686-linux, too:

Can you try with `CFLAGS="-msse -mfpmath=sse -march=pentium4" ./configure`? This fixes the bug when I compile with MinGW32 locally, which suggests to me it's an x87 floating point bug :(

AliceLR:
For reference, this is the *single sample frame* that was different in the output between SSE and x87.

image

sezero (approved):

sezero:
This is OK by me, although I do wonder where in the code the difference comes into being.
Merge?

AliceLR:
> This is OK by me, although I do wonder where in the code the difference comes into being.

I didn't investigate this but given that it's caused by floats and it only affects one sample frame, it has to be in the main mixer loop and is likely caused by using a double precision float for the current voice position. There are several double->integer conversions that could cause it, as well as the calculation for the number of samples to render for a frame. IMO it probably isn't related to computing the step value or periods, otherwise it would affect a lot more of the output. It probably also isn't in the mixing inner loops since they use fixed point.

> Merge?

Since I already pinged Claudio about this patch (and the pending bidi patch) via email I think we should wait a little bit to see if he can respond. Like the anticlick patch, after further investigation I'm more confident that the changes are correct (and 99% of the diff is just noise), so if he doesn't respond, I think it's OK to just merge it.

sezero:
PING?

AliceLR:
FYI this *probably* needs to be rebased due to #500 (the data for the new tests probably needs to be updated). Will do that soon.

sezero:
Rebase? PING?

AliceLR:
I gave both of these patches a quick look and it *looked* like the fuzzing patch shouldn't have broken them, but I can rebase them anyway. The mixer.c changes somehow completely avoided both patches, and the tests affected are different too.

AliceLR:
> PING?

I'm getting crashes with bidirectional sustain loops now, so I went ahead and emailed Claudio about these again... : )

AliceLR:
Should I send another email to Claudio's spam box or should we just merge this?

sezero:
> Should I send another email to Claudio's spam box or should we just merge this?

He's super busy I guess.. We can merge and deal with any possible fallout later. Your call. 

> edit: for context, I found even more—you guessed it—mixer issues while processing my fuzzing corpus with UBSan.

Heh..

AliceLR:
> He's super busy I guess.. We can merge and deal with any possible fallout later. Your call.

I *suspect* this one is just a minor bug and won't cause any serious playback issues, I mainly tagged Claudio to see if he remembered any context re: the patch that added this. I'll follow up once I diagnose these UBSan errors and can tell whether or not they'll impact the diffs for #496 and #498. A little bit of fuzzing here probably wouldn't hurt either.

sezero:
PING? Your decision?

AliceLR:
OK, double checked and I don't think I need to verify this one against my partial fuzzing patch or anything. I'm just merging this thing.

PR #498, opened 2021-11-03
Move interpolation wraparound handling out of sample loader.

AliceLR:
libxmp modifies the end of the sample loop to improve continuity between sample loops when using interpolating mixers, but it does this in a way that's flawed in a couple of ways:

* It only modifies the end of the loop, so cubic interpolation will still be inconsistent at the first sample of the loop (as it will reference a single sample prior to the loop). This can be noticed as a quiet buzz in the test module linked to #496 when that patch is applied.
* It permanently modifies sample data, which affects sample hashes in the regression tests, and partially corrupts data after the loop. This isn't noticeable usually, but my patch to overhaul bidirectional and sustain loops ran into this problem because sustain loops in that patch now use the same sample data as the regular loop—one loop or the other will contain data that sounds wrong unless the two loops have perfectly overlapping endpoints.
* Permanently modifying sample data might also be problematic for users attempting to use libxmp as a tracker playback library, but I'm not entirely sure about that (libxmp needs to keep its own samples internally anyway to maintain prologue/epilogue bytes for interpolation).

This patch changes this to only temporarily modify and then revert sample data during mixing. This is very (infinitesimally) slower than before, but future changes to interpolation will no longer affect sample hashes, and modifying the sample prior to the loop for cubic interpolation is now possible.

This is different from how OpenMPT handles this problem (it uses a separate buffer for the wraparound section). As far as I can tell, MikMod and libmodplug don't handle this problem *at all*. I think modifying the sample once or twice per tick per channel and reverting it *probably* works out to faster than doubling the number of invocations of the mixing loops, but I didn't verify. libxmp master, libxmp with this patch, and OpenMPT all take roughly the same amount of time to render various reference modules I used to test this patch. The slowest one I tested is here: https://modarchive.org/index.php?request=view_by_moduleid&query=41557.

TODO:

- [x] Uncomment the cubic loop start fix. Currently this is disabled because it will require changes to the cubic mixing regression tests and possibly test/test.c. I want to make sure the tests reliably pass without this, first.
- [x] Test module (based on the module I linked below).


AliceLR:
Test module that reproduces this without relying on the loop timing patch: [interpolate_loop.it.zip](https://github.com/libxmp/libxmp/files/7483497/interpolate_loop.it.zip)

With `xmp -i nearest` or `xmp -i linear`, a short note plays, followed by silence. With `xmp -i spline`, a quiet buzz will follow the note for the rest of the module.

A quick look at the waveform shows that the sample prior to the loop is the issue (though that was obvious already): when entering the loop, the output dips below 0 due to interpolating `127 0 0 0` (this is expected). Each time the loop repeats after, there's a nearly identical dip.

image

With the fix portion of this patch uncommented, the former is present and the latter absent. Most of the module is silent like you'd expect.

image

I confirmed that this *does* break the test module (ew) in addition to the cubic mixer tests, but these are all easy to patch.

AliceLR:
I uncommented the fix and fixed all of the tests for it. I also added a test based on the module above. Other things I found: I confirmed Modplug 1.16 has the same playback bug as libxmp (without this patch) with both its cubic and fir interpolation.

This should be ready for review now.

sezero:
Build-tested on i686-linux, check and devcheck OK.

Tried crudely rebasing #696 an top of it, check fails, devcheck fails in
test_mixer_mono_8bit_spline and test_mixer_stereo_8bit_spline

Will test more later.

AliceLR:
> Tried crudely rebasing #496 an top of it, check fails, devcheck fails in test_mixer_mono_8bit_spline and test_mixer_stereo_8bit_spline

I mentioned this in the email I sent to you and Claudio but not here, sorry—both patches change those tests and I'll have to rebase one of them with updated test.c hash(es?) and 8-bit spline mixer data after the other is merged because of that :(

sezero (approved):
OK by me.

Merge, or wait for @cmatsuoka (and/or @sagamusix) for additional review?

AliceLR:
I think it'd be good to give Claudio a chance to look at this patch as well. Like #496 I think this one is fairly clear cut as being a bug, but there's a chance he might find the methodology questionable... 😅

sagamusix:
I welcome this change but I have too little insigt into xmp's inner workings to provide any meaningful review.

AliceLR:
FYI this *probably* needs to be rebased due to #500 (the data for the new tests probably needs to be updated). Will do that soon.

sezero:
Rebase? PING?

sezero:
PING? Decision?

AliceLR:
Fixing the test hashes (and potentially other test breakage) right now, then this can be merged.

PR #500, opened 2021-11-08
Fix support for most Farandole Composer effects.

AliceLR:
Work in progress patch to fix support for most Farandole Composer (FAR) effects that were broken, particularly the tempo effects and persistent tone portamento. This patch is not remotely close to finished and hasn't been thoroughly tested. Due to requiring several new module and channel fields, part of this patch is implemented with format extras.

- [x] ~~Implement effect `03` Fulfill loop.~~ It's supposed to be sustain release but it's very buggy (presumably due to incorrectly configuring the GUS voice) and the vast majority of the time works like a note cut instead. For now it's implemented with `FX_KEYOFF` unless there's a version of Farandole Composer without this broken that can be tested.
- [x] Implement effect `A?` "Slide-to-vol" (persistent volslide to volume column value). This will probably need another new effect.
- [x] Effect `4?` Retrigger is still buggy in old tempo mode, should fix.
- [x] ~~Try to figure out what the deal is with effect `03` and 16-bit samples that makes them very difficult to use in Farandole Composer (I suspect misuse of GUS). If `03`'s behavior can reliably be reproduced, sustain release from it could be supported in the future.~~ lol :(

Testing:

- [x] Manually test "fulfill loop" effect `03`.
- [x] Manually test new FAR tempo mode (`05`) with tempo effects `D?`, `E?`, `F?`.
- [x] Manually test old FAR tempo mode (`04`) with tempo effects `D?`, `E?`, `F?`.
- [x] Manually test tone slide effects `1?` `2?`.
- [x] Manually test persistent tone portamento effect `3?`.
- [x] Manually test retrigger effect `4?`.
- [x] Manually test vibrato effects `5?` `6?` `9?`.
- [x] Manually test volslide effects `7?` `8?`.
- [x] Manually test persistent volslide effect `A?`.
- [x] Manually test balance effect `B?`.
- [x] Manually test note offset effect `C?`.
- [x] Verify all existing FAR modules against this patch (only ~36 of them...).
- [x] Do some fuzzing...

Regression tests:

- [x] FAR loader regression test(s) for effects coverage.
- [x] Effects regression tests for new FAR effects.
- [x] ~~Further coverage required?~~ The main missing coverage is related to old tempo mode retrigger, not worth it.

Misc:

- [ ] Part of this patch (the "new" FAR tempo translation code) is based directly on the FAR replayer by Daniel Potter due to handling for it being very unusual and convoluted. It's not clear what the license on that is (the intent seems to be that it can be used to write a custom player with no restriction).
- [x] The FAR format provides a header length field that libxmp was ignoring. OpenMPT, MikMod, and libmodplug all handle this as specified by the format document.
- [x] libxmp ignored the FAR format's default channel panning table and muted channel flags.
- [x] Load module comment text.
- [x] Fixed FAR restart position.
* Effects `01` and `02` are effects for controlling volume ramping in Farandole Composer's replayer. This is presumably a slow operation and these effects can temporarily defer handling it. libxmp can safely ignore them.


AliceLR:
Looks like effect `03` "Fulfill loop" is another sustain release style effect, so this patch is going to be blocked on the bidi patch (which is blocked on two other patches) too. 🥲

sagamusix:
I don't think xmp handles this yet, and I don't see the magic constants in your code yet, so here's a comment regarding pitch slide up / down: As can be seen from the documentation, the slide units here are GUS frequency units. The precision of those units depends on the number of initializeds GUS channels, which curiously is 17 and not 16 as one might expect, which means that one pitch slide unit adds or subtracts <sup>36318</sup>/<sub>1024</sub> Hz. Needless to say FAR should be handled with Hz and not Amiga periods internally.

AliceLR:
> I don't think xmp handles this yet, and I don't see the magic constants in your code yet, so here's a comment regarding pitch slide up / down: As can be seen from the documentation, the slide units here are GUS frequency units. The precision of those units depends on the number of initialized GUS channels, which curiously is 17 and not 16 as one might expect, which means that one pitch slide unit adds or subtracts 36318/1024 Hz. Needless to say FAR should be handled with Hz and not Amiga periods internally.

I've been putting off fixing the pitch effects since I don't think libxmp currently has handling for it. I'm running out of other things to fix in this patch, though...

> Edit: Similarly, I think all the weird tempo stuff is probably based on GUS timers? I never quite finished figuring that out.

I *think* it might be intended to be the Programmable Interval Timer due to the constant and the usage of divisors, but the constant is off (is 1197255 Hz, should be 1193182 Hz). This is pure speculation, but apparently the PIT ports are 0x40 through 0x43, which lines up with what's in Daniel Potter's replayer code.

sagamusix:
I tried command 03 here and it doesn't seem to finish the loop; it seems to stop the sample immediately?

AliceLR:
> I tried command 03 here and it doesn't seem to finish the loop; it seems to stop the sample immediately?
> 
> Edit: The test file I'm using: [FULFLOOP.zip](https://github.com/libxmp/libxmp/files/7508371/FULFLOOP.zip)

I haven't implemented this effect yet since it's going to be yet another patch modifying the mixer code... right now it just uses `FX_KEYOFF` :(

Unless you mean in Farandole Composer, in which case maybe it's another Farandole Composer bug. It's *supposed* to finish the loop (and doesn't play anything after the loop), and I've had success there with a shorter sample.

sagamusix:
Sorry, I did mean in Farandole Composer. Guess I'll try again with a bunch of different samples tomorrow...

AliceLR:
> Sorry, I did mean in Farandole Composer. Guess I'll try again with a bunch of different samples tomorrow...

On further review of my test module: when used *on the same row* as a note, it works as intended. When used on a different row, it does seem to act like a note cut. Very strange.

AliceLR:
I caved and looked at the replayer code for `03` and I feel like I understand even less.

Here's how the replayer starts a sample:
```C
          if (m&(1<<2))
            PlaySample(Sample[q].Seg/2,
                       Sample[q].Off/2,
                       Sample[q].Rep/2,
                       Sample[q].RepEnd/2,
                       ch,m);
          else
            PlaySample(Sample[q].Seg,
                       Sample[q].Off,
                       Sample[q].Rep,
                       Sample[q].RepEnd,
                       ch,m);
```

Here's what `03` does:
```C
            case 3:
              q=CurChSmp[CurVoice];
              SetLoops(Sample[q].Seg,Sample[q].Off,Sample[q].Len,
                Sample[q].LoopMode&0xf7,CurVoice);
              break;
```

The implementations of `PlaySample` and `SetLoops` are not provided, but this makes it seem like `03` is *supposed* to be sustain release i.e. play the sample to the end without looping. ugh

AliceLR:
I'm just leaving `03` as `FX_KEYOFF` for now without a sustain loop. If the sustain release effect can be reliably reproduced this can be changed to use a sustain loop later and it should work correctly. (So this patch isn't blocked on other patches, just on me being slow.)

AliceLR:
I confirmed libxmp actually does have what should be GUS-compatible period mode (`PERIOD_CSPD`, which sets the "period" in Hz). I tried it before and it didn't work, but it turns out this is because it uses integer division when it should instead be using float division (whoops). Not sure how this wasn't noticed before—669 is the only format that currently uses `PERIOD_CSPD`, so maybe it just wasn't tested adequately.

AliceLR:
Effect `40` crashes Farandole Composer. i don't think i'll be adding support for this feature

sagamusix:
Oh, where did you find that player code? I wasn't aware it even exists! My copy of Farandole only comes with FARLOAD.CPP, which is pretty useless on its own...

AliceLR:
Here's the archive it's in since I don't remember exactly where I got it. I'm fairly certain I got it from a legitimate source, though: [far100.zip](https://github.com/libxmp/libxmp/files/7511091/far100.zip)

I had to copy the tempo handling almost verbatim from it because it's so bizarre (and by extension, part of the retrigger/note offset handling, because they're derived from tempo and equally as strange). Hope the license is friendly, but due to the absurdity this probably falls under fair use in the unlikely event it wasn't intended to be public domain.

AliceLR:
Had to completely readjust the module timing because I found out testing retrigger that `OCount` is decremented after checking, meaning an `OCount` of 3 actually corresponds to 4 "ticks". So e.g. tempo 4 works out to speed=4, bpm=32, rate=50/128. (Previously I had it as speed=3, bpm=32, rate=50/96...)

AliceLR:
Apparently FAR saves any blank patterns as size 0, even if they're reference in the order list. `Guillermo Luijk Delgado/to the polish of valencia.far` and apparently some other modules rely on this.

AliceLR:
I finally finished verifying libxmp against all of the existing FAR modules I have and found a few more issues to fix. Current results for libxmp, OpenMPT, and MikMod, as well as some effects implementation info, are [here](https://gist.github.com/AliceLR/a5b7ace9d678ab407f974d9663236cc2).

AliceLR:
I figured out 16-bit samples (but not sustain release yet): Farandole Composer seems to load samples one after another to the GUS without any alignment. 16-bit samples reliably always break when loaded to an odd offset (i.e. an odd number of odd-length samples have been loaded prior) and reliably always work when loaded to an even offset.

AliceLR:
NOPing 96 bytes of Farandole Composer (`0x23459` through `0x234B8`) sort-of-fixes sustain release for all samples (it finishes the loop but doesn't play anything after).

These bytes are the part of the function `SetLoops` which sets the sample loop end, and removing them leaves it so that function just clears the loop flag. This suggests it *is* something buggy with how Farandole Composer sets the sample loop end (or alternatively, what it does worked on real hardware but not in DOSBox for whatever reason). Still don't know entirely what the issue is there, but at least that narrows it down.

AliceLR:
Okay, after more careful inspection of what `PlaySample` does vs. `SetLoops`, I understand why it only works for 16-bit samples except for the first loaded.

`PlaySample` computes values to give to the GUS from `Seg`, `Off`, `Rep`, and `RepEnd` with what looks like the following (sane) code:
```C
static int BASE;
static long addrOffset;
static long addrStart;
static long addrEnd;

static void _PlaySample_GUS(int channel, int mode)
{
  outportb(BASE + 0x102, channel);

  outportb(BASE + 0x103, 0xA);
  outport(BASE + 0x104, addrOffset >> 7);
  outportb(BASE + 0x103, 0xB);
  outport(BASE + 0x104, addrOffset << 9);

  outportb(BASE + 0x103, 0x2);
  outport(BASE + 0x104, addrStart >> 7);
  outportb(BASE + 0x103, 0x3);
  outport(BASE + 0x104, addrStart << 9);

  outportb(BASE + 0x103, 0x4);
  outport(BASE + 0x104, addrEnd >> 7);
  outportb(BASE + 0x103, 0x5);
  outport(BASE + 0x104, addrEnd << 9);

  outportb(BASE + 0x103, 0x0);
  outportb(BASE + 0x105, mode);
}

void PlaySample(int segment, int offset, long repeat_start, long repeat_end, int channel, int mode)
{
  ...
  addrOffset = ((long)segment << 16) | offset;
  addrStart = addrOffset + repeat_start;
  addrEnd = addrOffset + repeat_end;
  _PlaySample_GUS(channel, mode);
}
```

`SetLoops` doesn't bother to add the sample length to the computed end value, so the loop end value is the START of the sample. I confirmed that the original ASM routine I translated here *is* called directly from the playercode, so this is almost certainly the source of this issue. Furthermore, the segment value gets shifted out entirely and the upper word of `addrEnd` is always 0(!).
```C
static void _SetModeLoop_GUS(int channel, int mode)
{
  int current;

  outportb(BASE + 0x102, channel);

  outportb(BASE + 0x103, 0x80);
  current = inportb(BASE + 0x105);

  mode = (current & 0xE7) | mode;
  outportb(BASE + 0x103, 0);
  outportb(BASE + 0x105, mode);
}

void SetLoops(int segment, int offset, long length, int mode, int channel)
{
  addrEnd = ((int)segment << 16) /* !?!? */ | offset;
  outportb(BASE + 0x102, channel);

  outportb(BASE + 0x103, 4);
  outport(BASE + 0x104, addrEnd >> 7);

  outportb(BASE + 0x103, 5);
  outport(BASE + 0x104, addrEnd << 9);

  _SetModeLoop(channel, mode);
}
```

This works for 16-bit samples because of another bug, which is that the handler for `03`  (in an above comment) never divides the segment/offset/length by 2 for 16-bit samples, so these all inadvertently end up being double what they should be, allowing some 16-bit samples to continue normally. The first loaded 16-bit sample doesn't work because it has an offset of 0, and later or large 16-bit samples won't work because of the missing cast when calculating `addrEnd`.

This was a lot of effort to say "probably it should just be left as note cut for now". Ugh.

sagamusix:
And here I thought FT2 was bad...

AliceLR:
Nothing new has come up during fuzzing and I don't *think* I missed anything else, so I'm marking this ready for review.

sezero (approved):

Issue #450, opened 2021-07-23
four test cases fail with vs6


sezero:
As title says, this is with MSVC6 - I printed the differences resulting in failure. Here:
```
test 78: test_loader_s3m_adpcm: at util.c:271: subinst transpose: **fail** <== sub->xpo: 11, x: 12
test 86: test_loader_it_oldins: at util.c:271: subinst transpose: **fail** <== sub->xpo: 23, x: 24
test 97: test_loader_dbm_venv: at util.c:271: subinst transpose: **fail**  <== sub->xpo: 11, x: 12
test 400: test_openmpt_it_filter_7f: at compare_mixer_data.c:64: period mismatch: **fail**  <== 
								period: 876544, ci->period: 876939
```

This is certainly not a big issue, newer MSVC doesn't fail, but would be nice
to understand if we can, as time permits - if not, we can close this.

PR #508, opened 2021-12-22
strip down miniz:

sezero:
This greatly strips down our miniz after @ccawley2011's PR #436 merge:

* remove everything that libxmp-specific config already ifdefs-out.
* remove procs not used in libxmp.
* comment out some of the procs in 'Misc utils' section.
* remove m_pFile, m_file_archive_start_ofs and also m_mem_size and
  m_mem_capacity from struct mz_zip_internal_state_tag.
* make miniz_def_alloc_func & co, and mz_crc32 static to miniz.c.

The actual butchery happens in the zip file handler: miniz_tinfl.c
functionality stays there 100%.

Next step to this would be splitting miniz.[ch] into miniz_tinfl.[ch]
and miniz_zip.[ch]:  If this is acceptable, I would do that too but 
after #504 is decided in order to avoid conflicts.


AliceLR:
I'll start uhh, sifting through this when I can. :(

this "amalgamation" thing might have an excuse if it was JavaScript, not C 🤦‍♀️

sezero:
Done splitting into miniz_tinfl and miniz_zip (should help with moving muse.c into main tree).

sezero:
.. And merged two tiny updates from mainstream.

There is also https://github.com/richgel999/miniz/commit/2beb69d0a5633682148cfc0722f07329612a4478 which disables unaligned read/writes by default - don't know whether we want that.

sezero:
I'm pretty much finished with this patchset (can squash if it is to be accepted.)
To summarize: we split our miniz into tinfl and miniz_zip, tinfl is unchanged as
compared to mainstream, miniz_zip is butchered-down (the only & only reason
we use miniz_zip is for only one unzip depacker.)

I'm still interested about the mainstream PR/159 (https://github.com/richgel999/miniz/pull/159)
Can you look at it? (I am attaching here a version that applies cleanly to this
branch: [159.zip](https://github.com/libxmp/libxmp/files/7771641/159.zip) )

src/depackers/miniz_zip.h
```diff
@@ -0,0 +1,246 @@
+
+#pragma once
+#include "miniz.h"
```
AliceLR, line 3: This applies to miniz.h too, but include-guards should be more portable than `#pragma once`, right?

sezero:
@AliceLR: Please review.

I can merge https://github.com/richgel999/miniz/pull/159 patchset
to this too (which passes our CI tests), or I can make another PR for
it later.

src/depackers/miniz_zip.h
sezero: Nice, will change shortly.

What is the thing with the CI macos failure???

AliceLR:
That tinfl patch looks fine, though this seems like a write-only codebase, so I'm not sure. Conceptually I'm fine with this, but the diff isn't particularly helpful. I'm going to peek at the coverage to figure out why it's only at 50% for miniz_zip.c.

Also, I think we found a crash bug in clang on macOS :/

src/depackers/miniz_zip.h
AliceLR: It's a clang crash bug that (I thought) happens infrequently with parallel builds. It also happened in one of my AliceLR/libxmp regression test runs, but so far it doesn't seem to happen with the other clang-based regression tests.

I see it sometimes when building with clang + sanitizers in Fedora, so it might specifically be a thing for newer clang versions. It *only* ever happens with test-dev (not the main executable) in my experience.

sezero: Replaced those `#pragma once` with header guards.

sezero: > It's a clang crash bug that (I thought) happens infrequently with parallel builds. It also happened in one of my AliceLR/libxmp regression test runs, but so far it doesn't seem to happen with the other clang-based regression tests.
> 
> I see it sometimes when building with clang + sanitizers in Fedora, so it might specifically be a thing for newer clang versions. It _only_ ever happens with test-dev (not the main executable) in my experience.
> 
> edit: I don't know _why_ it would only happen with parallel builds since that should only matter for Make, not clang.

Never happened here before, and now it didn't happen after my header guard commit.
Does apple and/or llvm guys know about it, I wonder..

AliceLR: Yeah, it only happens *sometimes*. I don't know what causes it, but this gets spammed a bunch, so maybe it's a hint?
```
warning: null character ignored [-Wnull-character]
```

See #526 is this keeps being a problem going forward.

AliceLR:
Okay, the miniz_zip.c coverage thing is actually easy to resolve, and is caused by two things:

* miniz supports ZIP64(!). I can't imagine this ever being useful for libxmp but if there's a way to force 7z or whatever to emit one for smaller archives that might be good for testing.
* There are a handful of unused exposed functions remaining. Not sure if they're useful to keep around and/or `#if 0` or something:
  * `mz_zip_reader_locate_file`
  * `mz_zip_reader_locate_file_v2`
  * `mz_zip_reader_extract_to_callback` (This one might actually be useful if we want to allow extracting to `FILE` *or* memory interchangeably at some point, otherwise, it can probably go.)

sezero:
> Okay, the miniz_zip.c coverage thing is actually easy to resolve, and is caused by two things:
> 
>     * miniz supports ZIP64(!). I can't imagine this ever being useful for libxmp but if there's a way to force 7z or whatever to emit one for smaller archives that might be good for testing. Removal looks like it'd be intrusive.

Removal of zip64 support will be _very_ intrusive (even though I was tempted about it multiple times because I'd be happier with 32 bit offsets.) 

>     * There are a handful of unused exposed functions remaining. Not sure if they're useful to keep around and/or `#if 0` or something:
>       
>       * `mz_zip_reader_locate_file`
>       * `mz_zip_reader_locate_file_v2`

These I can happily `#if 0` out (or delete): Confirm?

>       * `mz_zip_reader_extract_to_callback` (This one might actually be useful if we want to allow extracting to `FILE` _or_ memory interchangeably at some point, otherwise, it can probably go.)

Our original use was with this, and yes I thought that it may be of use in the future. What to do?

AliceLR:
> Removal of zip64 support will be _very_ intrusive (even though I was tempted about it multiple times because I'd be happier with 32 bit offsets.)

I think let's keep it and maybe add a test for it in another commit. It's nice for libxmp to be aware of it even if it's not going to be used much—things like the giant Mod Archive ZIPs come to mind.

> These I can happily `#if 0` out (or delete): Confirm?
> Our original use was with this, and yes I thought that it may be of use in the future. What to do?

OK, let's keep these around and `#if 0` them, maybe? I can imagine the specific filename locating might also be useful in the future if libxmp gets some way to target specific files in archives (that would be nice but I don't know what it'd look like).

sezero:
Done commenting out the unused zip procs.

sezero:
I also just pushed some MSVC6 build fixes. 5 of the test cases fail (c.f.: #450)
```
test 79: test_loader_s3m_adpcm: at util.c:274: subinst transpose: **fail**
test 87: test_loader_it_adpcm: at util.c:274: subinst transpose: **fail**
test 88: test_loader_it_oldins: at util.c:274: subinst transpose: **fail**
test 100: test_loader_dbm_venv: at util.c:274: subinst transpose: **fail**
test 419: test_openmpt_it_filter_7f: at compare_mixer_data.c:64: period mismatch: **fail**
```

P.S.: Do we want to avoid `stdint.h`, or live with it?

AliceLR:
I didn't even notice the `stdint.h` usage. Generally I like `stdint.h` and I really dislike scrubbing it from things like XZ Embedded or my ARC extractor. I'm not very familiar with which older targets do not support it at all. (I thought it was around as far back as GCC 2.95 but apparently it was only generally available around the time 3+ came out and only guaranteed to exist as of 4.5...) Other old compilers and `#ifndef HAVE_STDINT_H` would obviously need a support header.

sezero:
OK, we can live with stdint.h

As an alternative: since we are customizing to libxmp, we can include common.h
in miniz.h and typedef `mz_xxx` to libxmp's `xxx`

sezero:
I mean like:
```diff
diff --git a/src/depackers/miniz.h b/src/depackers/miniz.h
index e02c05b..e745e3d 100644
--- a/src/depackers/miniz.h
+++ b/src/depackers/miniz.h
@@ -277,19 +277,25 @@ typedef void *(*mz_realloc_func)(void *opaque, void *address, size_t items, size
 #include <string.h>
 
 /* ------------------- Types and macros */
+
+#if 1 /* libxmp-specific typedefs: */
+#include "../common.h"
+typedef uint8 mz_uint8;
+typedef int16 mz_int16;
+typedef uint16 mz_uint16;
+typedef uint32 mz_uint32;
+typedef int64 mz_int64;
+typedef uint64 mz_uint64;
+#else
+#include <stdint.h>
 typedef unsigned char mz_uint8;
 typedef signed short mz_int16;
 typedef unsigned short mz_uint16;
 typedef unsigned int mz_uint32;
-typedef unsigned int mz_uint;
-#if defined(_MSC_VER) && (_MSC_VER < 1600)
-typedef signed __int64 mz_int64;
-typedef unsigned __int64 mz_uint64;
-#else
-#include <stdint.h>
 typedef int64_t mz_int64;
 typedef uint64_t mz_uint64;
 #endif
+typedef unsigned int mz_uint;
 typedef int mz_bool;
 
 #define MZ_FALSE (0)
```

AliceLR:
> I mean like:

Actually, I think this is OK (and means we can worry about the compatibility for `stdint.h` some other day if we need to).

sezero:
OK pushed the change.

I don't think anything else significant is left with this (Except for that miniz/pr159
patchset, which I can create another pull request for.)

AliceLR (approved):
I agree, I think this patch is OK for now.

PR #510, opened 2021-12-23
Replace bidirectional loop unrolling with reverse sample rendering.

AliceLR:
libxmp currently unrolls bidirectional loops, which causes a few nasty problems to arise:

* Bidirectional samples take up to twice as much RAM as they should.
* IT sustain loops have to be maintained in a separate structure complete with a *duplicate* sample buffer, wasting more RAM.
* IT bidirectional sustain loops are predictably completely broken due to the above hackery.
* Anything that wants to use IT sustain loop semantics (#347, #491) has to buy into this mess and duplicate all of its samples.

As it turns out, the way stepping in the software mixer functions is written libxmp *already* supports reverse stepping samples, meaning the mixer just needs to generate/bound negative step values to be able to handle bidirectional loops (and eventually other reverse effects) properly.

This is a work in progress right now:

- [x] Normal and bidirectional loop rendering that should be identical isn't always unless nearest interpolation is used (round position up in mixer functions when step is negative?).
- [x] ~~The extra sample dropped for IT bidirectional loops may technically be the wrong one (I think it should be fine though).~~ It was the wrong one, and it was NOT fine.
- [x] Bidirectional loops still gradually diverge from forward loops after about 10 seconds. It turns out this was caused by `usmp` being decremented from the remaining samples count.
- [x] Nearest neighbor mixing isn't actually nearest neighbor, which is why it *seemed* to work before.
- [x] Add a mixing test module. (Need variants for both IT and XM.)
- [x] Still fails 8 tests. These just need to have their sample position data updated.
- [x] Squash.

Note: merging this will make the remaining bug in #145 appear to get worse, but it actually won't—the broken sustain loop bounds fixed by this patch were partially masking the issue there, which *seems* to be `VOICE_RELEASE` getting dropped in cases where it shouldn't.


AliceLR:
The latest sample divergence bug seems to be related to `usmp`, which *seems* to be intended to force the mixer loop to exit in the unlikely event a sample gets stuck past its bound. It does this by decrementing the number of samples to render if the current sample position is past the bound, but the result is on some ticks there are a few samples that get skipped even in normal cases.

I'm going to try replacing this with something else that doesn't affect the number of output samples rendered (might need to be yet another PR depending on the number of affected tests).

AliceLR:
I probably mentioned something about this elsewhere before and forgot about it: 7 of the 8 affected OpenMPT tests were testing things unrelated to bidi loops and this branch does not significantly change them. The one relevant test is Bidi-Loops.it and it still plays back correctly.

I've added a bidi regression test that covers both XM and IT. Once the regression tests are confirmed to be OK, this should just need squashing and review.

sezero (approved):
Looks good. Let's replace the utf8 use by simple ascii in two places
and OK with that change:
```diff
diff --git a/src/mixer.c b/src/mixer.c
index 3d603fa..c93b987 100644
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -437,7 +437,7 @@ static int loop_reposition(struct context_data *ctx, struct mixer_voice *vi,
 
 		/* Wrap voice position around endpoint */
 		if (vi->flags & VOICE_REVERSE) {
-			/* OpenMPT Bidi-Loops.it: "In Impulse Tracker’s software
+			/* OpenMPT Bidi-Loops.it: "In Impulse Tracker's software
 			 * mixer, ping-pong loops are shortened by one sample."
 			 */
 			vi->pos = vi->end * 2 - ctx->s.bidir_adjust - vi->pos;
@@ -516,7 +516,7 @@ void libxmp_mixer_softmixer(struct context_data *ctx)
 #endif
 
 #ifndef LIBXMP_CORE_DISABLE_IT
-	/* OpenMPT Bidi-Loops.it: "In Impulse Tracker’s software
+	/* OpenMPT Bidi-Loops.it: "In Impulse Tracker's software
 	 * mixer, ping-pong loops are shortened by one sample."
 	 */
 	s->bidir_adjust = IS_PLAYER_MODE_IT() ? 1 : 0;
```

AliceLR:
> Looks good. Let's replace the utf8 use by simple ascii in two places and OK with that change:

Oh, yuck, I didn't even notice that while editing. Fixed. I copy/pasted that comment with minor tidying only so I'm *fairly* sure it existed before this patch...

sezero:
Patch is in

AliceLR:
> Patch is in

Didn't think I'd ever get to finish it considering all of the bugs that kept coming up...

PR #511, opened 2021-12-23
stb_vorbis: change namespace from stb_ to libxmp_ for public functions.

sezero:


AliceLR (approved):
This seems like a reasonably non-intrusive way to handle this issue. Should probably wait until after #512 is merged so the seek defines can be dropped.

sezero:
> This seems like a reasonably non-intrusive way to handle this issue.

Agreed

> Should probably wait until after #512 is merged so the seek defines can be dropped.

No need: if in the unlikely case we re-enable anything, they will still be mangled properly.

Issue #481, opened 2021-08-31
Remaining stb_vorbis diff.


AliceLR:
This is the remaining diff between the local libxmp stb_vorbis and upstream + misc. patches I contributed pending merge. I've removed irrelevant portions and regrouped bits of it.

libxmp specific changes that probably shouldn't be upstreamed:
```diff
69a69,73
> /* libxmp customizations: */
> #define STB_VORBIS_C
> #include "vorbis.h"
> 
> 
355a360
> #ifndef STB_VORBIS_NO_FLOAT_CONVERSION
357a363
> #endif
650a653,656
> /* libxmp-specific change */
> #if 1
> #include "../common.h"
> #else
656a663
> #endif
975c982
<    return sz ? malloc(sz) : NULL;
---
>    return sz ? calloc(sz, 1) : NULL;
992c999
<    return malloc(sz);
---
>    return calloc(sz, 1);
5469a5482
> #ifndef STB_VORBIS_NO_FLOAT_CONVERSION
5521a5535
> #endif // STB_VORBIS_NO_FLOAT_CONVERSION
```

`alloca` usage tweaks and `STB_FORCEINLINE` appear to be interconnected:
```diff
593,600d598
< 
<    // find definition of alloca if it's not in stdlib.h:
<    #if defined(_MSC_VER) || defined(__MINGW32__)
<       #include <malloc.h>
<    #endif
<    #if defined(__linux__) || defined(__linux) || defined(__sun__) || defined(__EMSCRIPTEN__) || defined(__NEWLIB__)
<       #include <alloca.h>
<    #endif
607a606,620
> /* we need alloca() regardless of STB_VORBIS_NO_CRT,
>  * because there is not a corresponding 'dealloca' */
> #if !defined(alloca)
> # if defined(HAVE_ALLOCA_H)
> #  include <alloca.h>
> # elif defined(__GNUC__)
> #  define alloca __builtin_alloca
> # elif defined(_MSC_VER)
> #  include <malloc.h>
> #  define alloca _alloca
> # elif defined(__WATCOMC__)
> #  include <malloc.h>
> # endif
> #endif
> 
610,628c623,630
< #ifdef __MINGW32__
<    // eff you mingw:
<    //     "fixed":
<    //         http://sourceforge.net/p/mingw-w64/mailman/message/32882927/
<    //     "no that broke the build, reverted, who cares about C":
<    //         http://sourceforge.net/p/mingw-w64/mailman/message/32890381/
[...]

PR #512, opened 2021-12-23
more miscellaneous stuff: dead code

sezero:
1. generate a map file from watcom builds.
2. remove asif.c from build:
   It is only used by ssmt_load.c, which hasn't been built since commit 34354ab7
3. med4_load: remove global symbol MAX_CHANNELS
   its use has been removed as of commit 246b7c7
4. disable alloc_pattern_tracks_long & schism_tracker_string in lite builds
5. disable the seek api in stb_vorbis for libxmp
   Bound to the new STB_VORBIS_NO_SEEK_API define.
   The rest of the libxmp-unused functions in stb_vorbis are:
   stb_vorbis_get_info, stb_vorbis_get_error`, stb_vorbis_get_sample_offset,
   stb_vorbis_get_samples_short, stb_vorbis_get_samples_short_interleaved,
   and stb_vorbis_stream_length_in_seconds, which possibly aren't worth the
   trouble.  (Also see: https://github.com/libxmp/libxmp/issues/481)

There are also other code-dead-to-lite builds, such as `libxmp_virt_setnote`
and `hio_open_file2`, but they aren't worth the trouble of making an effort.

Real dead code is `move_data`: It's now only used by `test_write_file_move_data.c`.
Otherwise, the prowizard code is using `pw_move_data` and the depacker code
is using `depacker_move_data`, which are actually identical. What to do about
that?

### DIFF
--- a/src/loaders/loader.h
+++ b/src/loaders/loader.h
@@ -30,7 +30,9 @@ int	libxmp_alloc_pattern		(struct xmp_module *, int);
 int	libxmp_alloc_track		(struct xmp_module *, int, int);
 int	libxmp_alloc_tracks_in_pattern	(struct xmp_module *, int);
 int	libxmp_alloc_pattern_tracks	(struct xmp_module *, int, int);
+#ifndef LIBXMP_CORE_PLAYER
 int	libxmp_alloc_pattern_tracks_long(struct xmp_module *, int, int);
+#endif
 char	*libxmp_instrument_name		(struct xmp_module *, int, uint8 *, int);
 
 char	*libxmp_copy_adjust		(char *, uint8 *, int);
@@ -47,7 +49,9 @@ void	libxmp_set_type			(struct module_data *, const char *, ...);
 int	libxmp_load_sample		(struct module_data *, HIO_HANDLE *, int,
 					 struct xmp_sample *, const void *);
 void	libxmp_free_sample		(struct xmp_sample *);
+#ifndef LIBXMP_CORE_PLAYER
 void	libxmp_schism_tracker_string	(char *, size_t, int, int);
+#endif
 
 extern uint8		libxmp_ord_xlat[];
 extern const int	libxmp_arch_vol_table[];
--- a/src/period.c
+++ b/src/period.c
@@ -262,6 +262,7 @@ void libxmp_c2spd_to_note(int c2spd, int *n, int *f)
 	*f = c % 128;
 }
 
+#ifndef LIBXMP_CORE_PLAYER
 /* Gravis Ultrasound frequency increments in steps of Hz/1024, where Hz is the
  * current rate of the card and is dependent on the active channel count.
  * For <=14 channels, the rate is 44100. For 15 to 32 channels, the rate is
@@ -281,3 +282,4 @@ double libxmp_gus_frequency_steps(int num_steps, int num_channels_active)
 	CLAMP(num_channels_active, 14, 32);
 	return (num_steps * GUS_rates[num_channels_active - 14]) / 1024.0;
 }
+#endif
--- a/Makefile.vc
+++ b/Makefile.vc
@@ -58,7 +58,6 @@ OBJS	= \
  src\loaders\common.obj \
  src\loaders\iff.obj \
  src\loaders\itsex.obj \
- src\loaders\asif.obj \
  src\loaders\lzw.obj \
  src\loaders\voltable.obj \
  src\loaders\sample.obj \
--- a/cmake/libxmp-sources.cmake
+++ b/cmake/libxmp-sources.cmake
@@ -31,7 +31,6 @@ set(LIBXMP_SRC_LIST
     src/loaders/common.c
     src/loaders/iff.c
     src/loaders/itsex.c
-    src/loaders/asif.c
     src/loaders/lzw.c
     src/loaders/voltable.c
     src/loaders/sample.c
--- a/lite/watcom.mif.in
+++ b/lite/watcom.mif.in
@@ -7,6 +7,8 @@ STATICFLAGS=-DBUILDING_STATIC
 
 DLLNAME=libxmp.dll
 EXPNAME=libxmp.exp
+# Note: not libxmp.map...
+MAPNAME=xmp-lite.map
 LIBNAME=libxmp.lib
 LIBSTATIC=xmp_static.lib
 TESTNAME=libxmp-test.exe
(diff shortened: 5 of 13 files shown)


src/loaders/loader.h
```diff
@@ -30,7 +30,9 @@ int	libxmp_alloc_pattern		(struct xmp_module *, int);
 int	libxmp_alloc_track		(struct xmp_module *, int, int);
 int	libxmp_alloc_tracks_in_pattern	(struct xmp_module *, int);
 int	libxmp_alloc_pattern_tracks	(struct xmp_module *, int, int);
+#ifndef LIBXMP_CORE_PLAYER
 int	libxmp_alloc_pattern_tracks_long(struct xmp_module *, int, int);
+#endif
```
AliceLR, line 35: IMO just let anything that tries to use these fail to link...

src/period.c
```diff
@@ -262,6 +262,7 @@ void libxmp_c2spd_to_note(int c2spd, int *n, int *f)
 	*f = c % 128;
 }
 
+#ifndef LIBXMP_CORE_PLAYER
```
AliceLR, line 265: This one might actually be needed by core (#357). @sagamusix should S3M in GUS mode be using this table?

sagamusix: Probably not. You can observe a lower precision in the lower frequencies when using the GUS driver in ST3, but that can probably be implemented in a different way if you really want to emulate it (OpenMPT does for the time being).

AliceLR:
Commented a couple of places that might need to be tweaked but generally this looks good to me. I think the header changes are a little extra/cluttery and the GUS table *might* be required for proper S3M support.



> 2. remove asif.c from build:
>    It is only used by ssmt_load.c, which hasn't been built since commit [34354ab](https://github.com/libxmp/libxmp/commit/34354ab71386dee61c5a339bb72406a87cd520d3)

Good. Re: rewriting this loader with a permissive license, I bugged someone in the MZX community who got an Apple IIgs to test SoundSmith and they never got back to me... :^)

> 5. disable the seek api in stb_vorbis for libxmp
>    Bound to the new STB_VORBIS_NO_SEEK_API define.
>    The rest of the libxmp-unused functions in stb_vorbis are:
>    stb_vorbis_get_info, stb_vorbis_get_error`, stb_vorbis_get_sample_offset,
>    stb_vorbis_get_samples_short, stb_vorbis_get_samples_short_interleaved,
>    and stb_vorbis_stream_length_in_seconds, which possibly aren't worth the
>    trouble.  (Also see: [Remaining stb_vorbis diff. #481](https://github.com/libxmp/libxmp/issues/481))

Has this one been contributed upstream? It seems like it might be useful enough. I agree defining out some of the other unused functions could be messy and wouldn't decrease the executable size much...

> There are also other code-dead-to-lite builds, such as `libxmp_virt_setnote` and `hio_open_file2`, but they aren't worth the trouble of making an effort.

Agreed. This also came up in the heap depackers PR but I wonder if it's worth combining `hio_open_file` and `hio_open_file2` with an added release-on-close param.

> Real dead code is `move_data`: It's now only used by `test_write_file_move_data.c`. Otherwise, the prowizard code is using `pw_move_data` and the depacker code is using `depacker_move_data`, which are actually identical. What to do about that?

Ughhh... I think this was originally used by the depackers and got replaced for ccawley2011's hio depackers patch. Possibly `move_data` could be replaced with `depacker_move_data` and `pw_move_data` since, as far as I can tell, nothing relies on unabstracted stdio streams anymore.

src/loaders/loader.h
sezero: It _will_ fail linkage in `lite` builds

src/period.c
AliceLR: OK, then disregard this complaint. :D

sagamusix: Typo corrected: OpenMPT *doesn't* emulate reduced GUS frequency precision. Either way, if the table would required for that, it can always be ifdeffed in again, right?

src/loaders/loader.h
AliceLR: Oh, is this a Watcom thing? I haven't seen GCC fail to link due to undefined function declarations that aren't used.

sezero:
> > 5. disable the seek api in stb_vorbis for libxmp
> >    Bound to the new STB_VORBIS_NO_SEEK_API define.
> > [...]
> 
> Has this one been contributed upstream? It seems like it might be useful enough.

I have not. I can try if you really want, but mainstream seems slow to react

sezero:
> > Real dead code is `move_data`: It's now only used by `test_write_file_move_data.c`. Otherwise, the prowizard code is using `pw_move_data` and the depacker code is using `depacker_move_data`, which are actually identical. What to do about that?
> 
> Ughhh... I think this was originally used by the depackers and got replaced for ccawley2011's hio depackers patch. Possibly `move_data` could be replaced with `depacker_move_data` and `pw_move_data` (and then those could be removed) since, 

Easy peasy, we can combine the two into an `hio_move_data`
What to with `test_write_file_move_data.c`? Do we want it?

>   since, as far as I can tell, nothing relies on unabstracted stdio streams anymore.

Aren't file writes still stdio? (Maybe I'm missing something.)

AliceLR:
> What to with `test_write_file_move_data.c`? Do we want it?

I think this test is reasonable, it'd just need to be updated to take an hio input, I think...

> Aren't file writes still stdio? (Maybe I'm missing something.)

Yes, I meant for *input* specifically, sorry for being unclear. Writes for depackers and ProWizard are still stdio...

src/loaders/loader.h
sezero: I actually meant any internal use of the procedure will fail linkage.
Otherwise, this (and the other non-`LIBXMP_CORE_PLAYER` procs)
are not public procs: they won't be used by any clients either way.
(Am I reading you correctly?)

AliceLR: Oh, I see. What I meant is that leaving the declarations as-is and letting linking fail if anyone actually tries to use them (which *shouldn't* currently be happening in `lite` builds) would probably be okay. It doesn't really matter that much though. Let's just leave the #ifdefs in the headers.

src/period.c
AliceLR: Right.

AliceLR (approved):
Unless `move_data` and `hio_open_file#` are going to be addressed with this patch (IMO they can have a separate one)), this looks OK to me.

src/loaders/loader.h
sezero: OK

sezero:
> Unless `move_data` and `hio_open_file#` are going to be addressed with this patch (IMO they can have a separate one), 

Will do in a separate patch

>  this looks OK to me.

OK, applying.

sezero:
> > > 5. disable the seek api in stb_vorbis for libxmp
> > >    Bound to the new STB_VORBIS_NO_SEEK_API define.
> > >    [...]
> > 
> > 
> > Has this one been contributed upstream? It seems like it might be useful enough.
> 
> I have not. I can try if you really want, but mainstream seems slow to react

I just submitted it to mainstream as https://github.com/nothings/stb/pull/1266

PR #513, opened 2021-12-23
Disable building shared for PSP by default

sharkwouter:
The Playstation Portable actually does not support building shared libraries. We currently build with ``BUILD_SHARED=OFF``, but it would be nice if we didn't need to set that.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@ else()
 endif()
 # -----------------------------------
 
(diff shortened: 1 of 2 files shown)


sezero (commented):
You should do the same to `lite/CMakeLists.txt` -- OK with that change.

sharkwouter:
Ah, I had not seen that file. I've updated it now.

sezero (approved):

sezero:
Thanks.

PR #514, opened 2021-12-24
combine all move_data procedures

sezero:
The temp file writes are annoying, hope there were a better way..

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -68,7 +68,7 @@ typedef unsigned short int uint16;
 typedef unsigned int uint32;
 #endif
 
-#ifdef _MSC_VER				/* MSVC++6.0 has no long long */
+#ifdef _MSC_VER /* MSVC++ 6.0 has no long long */
 typedef signed __int64 int64;
 typedef unsigned __int64 uint64;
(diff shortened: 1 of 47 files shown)


sezero:
Thinking about it a bit more, the only other user of this procedure other
than the whole prowizard suite, is depackers/mmcmp: as mentioned in #509,
mmcmp can be changed to operate on heap only, and then we can keep prowiz
as is and remove [depacker_]move_data and test_write_file_move_data.c

AliceLR:
> Thinking about it a bit more, the only other user of this procedure other than the whole prowizard suite, is depackers/mmcmp: as mentioned in #509, mmcmp can be changed to operate on heap only, and then we can keep prowiz as is and remove [depacker_]move_data and test_write_file_move_data.c

This patch looks good, but yes, I think you're right. I was under the impression a lot more depackers than mmcmp used it, for whatever reason. I think I can revise the mmcmp depacker and then instead this patch can just drop the non-PW version. This would reduce the diff significantly and at the very least keep one more non-PW function out of PW.

sezero:
OK, forfeiting this one and closing - the solution will be changing mmcmp.

PR #516, opened 2021-12-24
Depack MMCMP to buffer, remove move_data and depacker_move_data.

AliceLR:
Updates the MMCMP depacker to use an intermediate buffer instead of file IO, which should make it much easier to adapt for #509.

Also removes the now unused functions `move_data` and `depacker_move_data`. Since the ProWizard version of this function isn't easily accessible from the regression tests right now, I just dropped the relevant regression test. (This function is already getting tested a ton from the ProWizard loader tests anyway.)

PLEASE test this against some real MMCMP files, as far as I know I don't have any and I'm just going off of the regression tests :(


AliceLR:
(Coverage drop is due to missing MMCMP error case coverage combined with several new safety checks.)

sezero:
Tested by running with modules compressed with the mmcmp tool.
valgrind whines:
```
==16007== Syscall param write(buf) points to uninitialised byte(s)
==16007==    at 0x9A0CFB: ??? (syscall-template.S:82)
==16007==    by 0x93A05F: _IO_file_xsputn@@GLIBC_2.1 (fileops.c:522)
==16007==    by 0x92FBDD: fwrite (iofwrite.c:45)
==16007==    by 0x4071A9E: decrunch_mmcmp (mmcmp.c:444)
==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
==16007==    by 0x40256F6: xmp_load_module (load.c:371)
==16007==    by 0x804B180: main (main.c:407)
==16007==  Address 0x40deb72 is 298 bytes inside a block of size 772,338 alloc'd
==16007==    at 0x40072B2: malloc (vg_replace_malloc.c:270)
==16007==    by 0x407176B: decrunch_mmcmp (mmcmp.c:346)
==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
==16007==    by 0x40256F6: xmp_load_module (load.c:371)
==16007==    by 0x804B180: main (main.c:407)
```
Changing malloc to calloc fixes it.
```diff
diff --git a/src/depackers/mmcmp.c b/src/depackers/mmcmp.c
index bb1a8ec..e33888b 100644
--- a/src/depackers/mmcmp.c
+++ b/src/depackers/mmcmp.c
@@ -343,7 +343,7 @@ static int decrunch_mmcmp(HIO_HANDLE *in, FILE *out, long inlen)
 	if (table == NULL) {
 		goto err;
 	}
-	outbuf.buf = (uint8 *) malloc(h.filesize);
+	outbuf.buf = (uint8 *) calloc(1, h.filesize);
 	if (outbuf.buf == NULL) {
 		goto err2;
 	}
```

sezero:
PS: mmcmp in libmikmod too runs on a memory buffer, might be helpful
to compare for sanity checks, etc, if you haven't already.

AliceLR:
> PS: mmcmp in libmikmod too runs on a memory buffer, might be helpful to compare for sanity checks, etc, if you haven't already.

libmodplug's sanity checks are famously very bad, so I don't know how much that would help (IIRC I had to fix a couple there before...). Switching to `calloc` there is a good idea sine MMCMP jumps around the stream, fixed.

sezero:
> libmodplug's sanity checks are famously very bad, so I don't know how much that would help

Then it's a good thing that libmikmod's version diverged from libmodplug's quite a while ago :)

Anyways, applying this.

sezero (approved):

AliceLR:
FWIW I'm looking at libmodplug now just in case and have found:

* Despite the extra data in the buffer, literal blocks are bounded to `dwFileSize` in the same manner I implemented here.
* Both packed types bound the unpacked subblock size against `dwFileSize` as well, then bound the unpacked contents of each block against the unpacked subblock size.
* In each of these cases when the data would extend past the file size, rather than erroring, it just just breaks. I don't know whether or not that's actually important.
* Still not actually clear what the extra runover space is for.

sezero:
> ```
> ==16007== Syscall param write(buf) points to uninitialised byte(s)
> ==16007==    at 0x9A0CFB: ??? (syscall-template.S:82)
> ==16007==    by 0x93A05F: _IO_file_xsputn@@GLIBC_2.1 (fileops.c:522)
> ==16007==    by 0x92FBDD: fwrite (iofwrite.c:45)
> ==16007==    by 0x4071A9E: decrunch_mmcmp (mmcmp.c:444)
> ==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
> ==16007==    by 0x40256F6: xmp_load_module (load.c:371)
> ==16007==    by 0x804B180: main (main.c:407)
> ==16007==  Address 0x40deb72 is 298 bytes inside a block of size 772,338 alloc'd
> ==16007==    at 0x40072B2: malloc (vg_replace_malloc.c:270)
> ==16007==    by 0x407176B: decrunch_mmcmp (mmcmp.c:346)
> ==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
> ==16007==    by 0x40256F6: xmp_load_module (load.c:371)
> ==16007==    by 0x804B180: main (main.c:407)
> ```
> 
> Changing malloc to calloc fixes it.

I'm actually worried as to why some parts of the malloc'ed dest buffer
are left untouched..

sezero:
OK, here is the file that makes valgrind whine, whereas libmikmod is OK: [AGONY.zip](https://github.com/libxmp/libxmp/files/7773751/AGONY.zip)

And this one fails loading with libxmp (with this patch or before it), whereas libmikmod loads it: [EVERSMOK.zip](https://github.com/libxmp/libxmp/files/7773754/EVERSMOK.zip)

sezero:
Attached problem files above (they are good in libmikmod).

The first one that results in valgrind warnings (with malloc instead of calloc)
was initially a wrong one, corrected in the edited version.

sezero:
Created ticket #517 for the issues found here

Issue #517, opened 2021-12-24
issues with mmcmp depacker


sezero:
Originally noted by https://github.com/libxmp/libxmp/issues/516#issuecomment-1000755649 :

This file that makes valgrind to report uninitialized read warnings when one changes calloc to malloc at line 346 for outbuf.buf, whereas libmikmod is OK: [FLYBY.zip](https://github.com/libxmp/libxmp/files/7773771/FLYBY.zip)

And this one fails loading with libxmp (with PR #516 or before it), whereas libmikmod loads it: [EVERSMOK.zip](https://github.com/libxmp/libxmp/files/7773754/EVERSMOK.zip)

sezero:
Here are the valgrind warnings with malloc instead of calloc:
```
==16007== Syscall param write(buf) points to uninitialised byte(s)
==16007==    at 0x9A0CFB: ??? (syscall-template.S:82)
==16007==    by 0x93A05F: _IO_file_xsputn@@GLIBC_2.1 (fileops.c:522)
==16007==    by 0x92FBDD: fwrite (iofwrite.c:45)
==16007==    by 0x4071A9E: decrunch_mmcmp (mmcmp.c:444)
==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
==16007==    by 0x40256F6: xmp_load_module (load.c:371)
==16007==    by 0x804B180: main (main.c:407)
==16007==  Address 0x40deb72 is 298 bytes inside a block of size 772,338 alloc'd
==16007==    at 0x40072B2: malloc (vg_replace_malloc.c:270)
==16007==    by 0x407176B: decrunch_mmcmp (mmcmp.c:346)
==16007==    by 0x40708AF: libxmp_decrunch (depacker.c:295)
==16007==    by 0x40256F6: xmp_load_module (load.c:371)
==16007==    by 0x804B180: main (main.c:407)
```

AliceLR:
I don't think the parts of the dest buffer that aren't being written to are an issue—MMCMP streams can arbitrarily seek all over the place and probably intentionally skip stretches of 0 bytes. `GlobalAllocPtr` in libmodplug uses `malloc`+`memset` (???) and `MikMod_malloc` uses `calloc`.

AliceLR:
[...]

PR #518, opened 2021-12-24
Fix bad seeking behavior in MMCMP compressed blocks.

AliceLR:
Fixes one of the issues in #517: files ending with an 8-bit compressed block shorter than 256 bytes would be rejected because the ptable handling was incorrectly adapted from libmodplug.

### DIFF
--- a/src/depackers/mmcmp.c
+++ b/src/depackers/mmcmp.c
@@ -163,7 +163,7 @@ static int block_unpack_16bit(struct block *block, struct sub_block *sub,
 	if (mem_seek(out, sub->unpk_pos) < 0) {
 		return -1;
 	}
-	if (hio_seek(in, block->tt_entries, SEEK_SET) < 0) {
+	if (hio_seek(in, block->tt_entries, SEEK_CUR) < 0) {
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


AliceLR:
Ugh, MikMod actually does try to load the entire 0x100 bytes as well. Surely nothing *actually* relies on table entries past the bound, right? Thanks, libmodplug...

AliceLR:
Revised to be closer to Olivier Lapicque's, uhhhh, vision. 🤦‍♀️

src/depackers/mmcmp.c
```diff
@@ -232,8 +232,13 @@ static int block_unpack_8bit(struct block *block, struct sub_block *sub,
 	uint32 numbits = block->num_bits;
 	uint32 j, oldval = 0;
 	uint8 ptable[0x100];
+	long seekpos = hio_tell(in) + block->tt_entries;
 
-	if (hio_read(ptable, 1, 0x100, in) != 0x100) {
+	/* The way the original libmodplug depacker is written allows values
+	 * to be read from the compressed data. It's impossible to tell if this
+	 * was intentional or yet another bug. Nothing seems to rely on it. */
+	memset(ptable, 0, sizeof(ptable));
```
sezero, line 240: do we need the memset ?

AliceLR: Yes. The input stream can index any position in that array, though valid data *seems* to never index beyond `block->tt_entries`. The read is not guaranteed to initialize the entire array (which is why EVERSMOK.S3M failed to load).

sezero: >  (which is why EVERSMOK.S3M failed to load)

The thing is, I can load and play EVERSMOK.S3M with that memset commented out

AliceLR: That's because before the patch it would reject the module unless the read returned 0x100.

sezero: OK then.

sezero (approved):
This is good for one of the failures, so let's apply.

I guess we go back to malloc instead calloc for `outbuf.buf` and
continue investigating.

PR #519, opened 2021-12-25
Fix uninitialized reads caused by odd length 16-bit MMCMP blocks.

AliceLR:
libmodplug is the gift that keeps giving


sezero (approved):

sezero:
> libmodplug is the gift that keeps giving

Heh :)))

PR #521, opened 2021-12-25
Fix nearest neighbor to actually round to the nearest neighbor.

AliceLR:
Nearest neighbor interpolation in libxmp isn't actually nearest neighbor—the rounding step was missing, so it was effectively just flooring the sample position. This is a problem with the reverse sample rendering added in #510, which makes the rounding actually relevant! Fixing this shouldn't significantly impact performance.

Play the attached module using the latest commit of #510 with linear and cubic interpolation, and it will be silent (until it diverges about 10 seconds in due to a different bug). Play it using #510 with nearest interpolation, and it will produce a very annoying buzz. Merge this patch and it goes away. This bug *only* affects #510 right now.

For extra fun, try this module with OpenMPT's "no interpolation" setting, which has the same issue. It doesn't actually claim to be nearest neighbor though. :)

[bidi.zip](https://github.com/libxmp/libxmp/files/7775874/bidi.zip)


sezero:
>  This bug _only_ affects #510 right now.

Should we wait until #510 finalizes, or apply now?

AliceLR:
> Should we wait until #510 finalizes, or apply now?

I *think* now would be better since then the regression tests in #510 could just be checks to make sure all output samples are 0±ε and then nearest neighbor could be included in that before it's merged.

sezero:
OK then

sezero (approved):

sagamusix:
Clever way of hoisting the rounding out of the inner loop, I think I'm gonna do something similiar as well in OpenMPT :)

PR #522, opened 2021-12-27
Replace unxz.c with an MIT reimplementation, update XZ Embedded.

AliceLR:
Since unxz.c was LGPL and didn't support extracting to a buffer, I went ahead and rewrote it to address both issues.

I also updated XZ Embedded to the latest version, but there's not a whole lot different: CRC-64 support we don't really care about, a small handful of minor(?) bugfixes, and some commenting. I reverted some stylistic changes to be closer to their original counterparts (like the CRC function names), but left everything that looked like a portability hack.

Some extra testing should be done on this before it's merged.

### DIFF
--- a/src/depackers/xz.h
+++ b/src/depackers/xz.h
@@ -2,7 +2,7 @@
  * XZ decompressor
  *
  * Authors: Lasse Collin <lasse.collin@tukaani.org>
- *          Igor Pavlov <http://7-zip.org/>
+ *          Igor Pavlov <https://7-zip.org/>
  *
  * This file has been put into the public domain.
  * You can do whatever you want with this file.
@@ -29,7 +29,7 @@ extern "C" {
  * enum xz_mode - Operation mode
  *
  * @XZ_SINGLE:              Single-call mode. This uses less RAM than
- *                          than multi-call modes, because the LZMA2
+ *                          multi-call modes, because the LZMA2
  *                          dictionary doesn't need to be allocated as
  *                          part of the decoder state. All required data
  *                          structures are allocated at initialization,
@@ -244,6 +244,18 @@ XZ_EXTERN void xz_dec_end(struct xz_dec *s);
 #		define XZ_INTERNAL_CRC32 1
 #endif
 
+/*
+ * If CRC64 support has been enabled with XZ_USE_CRC64, a CRC64
+ * implementation is needed too.
+ */
+#ifndef XZ_USE_CRC64
+#	undef XZ_INTERNAL_CRC64
+#	define XZ_INTERNAL_CRC64 0
+#endif
+#ifndef XZ_INTERNAL_CRC64
+#		define XZ_INTERNAL_CRC64 1
+#endif
+
 #if XZ_INTERNAL_CRC32
 /*
  * This must be called before any other xz_* function to initialize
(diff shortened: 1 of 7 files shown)


src/depackers/xz.h
```diff
@@ -241,7 +241,27 @@ XZ_EXTERN void xz_dec_end(struct xz_dec *s);
  * care about the functions below.
  */
 #ifndef XZ_INTERNAL_CRC32
+#	ifdef __KERNEL__
```
sezero, line 244: Do we need to add these back?

src/depackers/xz.h
```diff
@@ -241,7 +241,27 @@ XZ_EXTERN void xz_dec_end(struct xz_dec *s);
...
+/*
+ * If CRC64 support has been enabled with XZ_USE_CRC64, a CRC64
+ * implementation is needed too.
+ */
+#ifndef XZ_USE_CRC64
+#	undef XZ_INTERNAL_CRC64
+#	define XZ_INTERNAL_CRC64 0
+#endif
+#ifndef XZ_INTERNAL_CRC64
+#	ifdef __KERNEL__
```
sezero, line 260: same

src/depackers/xz_dec_stream.c
```diff
@@ -567,7 +590,7 @@ static enum xz_ret dec_main(struct xz_dec *s, struct xz_buf *b)
 	 */
 	s->in_start = b->in_pos;
 
-	for ( ;; ) {
+	while (xz_true) {
```
sezero, line 593: ??

src/depackers/xz.h
AliceLR: No, but it reduces the (sizeable) diff between libxmp and XZ Embedded a little bit.

src/depackers/xz.h
AliceLR: This one wasn't even there in the last version of XZ Embedded.

src/depackers/xz_dec_stream.c
AliceLR: Yeah, I don't know why this was changed either. This is the original.
```C
static enum xz_ret dec_main(struct xz_dec *s, struct xz_buf *b)
{
	enum xz_ret ret;

	/*
	 * Store the start position for the case when we are in the middle
	 * of the Index field.
	 */
	s->in_start = b->in_pos;

	while (true) {
		switch (s->sequence) {
```

AliceLR:
unrelated: of course it's the MSYS2 checks that fail, because they're the only ones that forces CI to install an entire userspace from scratch every time from an unreliable upstream. ugh

src/depackers/xz.h
AliceLR: Note the vast majority of the diff was changing stdint.h types to libxmp's custom fixed-width types, so this isn't even that big, and I can just remove these.

sezero: I'd suggest keeping the changes to minimum

AliceLR: OK, I ditched the `__KERNEL__` parts of these.

src/depackers/xz.h
AliceLR: I removed the `ifdef __KERNEL__` branch of this but left it otherwise, since "theoretically" CRC-64 might be useful if we end up with multiple formats that need it.

sezero (approved):

PR #523, opened 2021-12-27
minor updates to umx loader:

sezero:
- replace strcasecmp use with a more universally available solution.
- don't overzealously restrict the loader to known upkg versions.


AliceLR (approved):
This looks OK to me.

PR #527, opened 2021-12-28
Don't allow libxmp_decrunch to return a NULL HIO_HANDLE.

AliceLR:
This seems like a regression: `libxmp_decrunch` should never allow the `HIO_HANDLE` to be `NULL` because various places in load.c will attempt to close it, resulting in a crash.

Found with this Zip64 archive. The returned length ends up being `(size_t)UINT32_MAX`, which overflows to `(long)-1` in Windows, causing `hio_open_mem` to fail. That length is definitely a Zip64 placeholder for the 32-bit length field rather than a real length, but I haven't diagnosed it yet (the bug could plausibly be in either unzip.c or in miniz).

[ponylips.64.zip](https://github.com/libxmp/libxmp/files/7784434/ponylips.64.zip)

edit: it's absolutely a miniz bug: `mz_zip_reader_extract_to_heap` reads the 32-bit values out of the central directory header without so much as checking them for Zip64.

### DIFF
--- a/src/depackers/depacker.c
+++ b/src/depackers/depacker.c
@@ -205,6 +205,7 @@ static int decrunch_command(HIO_HANDLE **h, const char * const cmd[], char **tem
 	/* Don't use external helpers in android */
 	return 0;
 #else
+	HIO_HANDLE *tmp;
 	FILE *t;
 
 	D_(D_WARN "Depacking file... ");
@@ -227,10 +228,12 @@ static int decrunch_command(HIO_HANDLE **h, const char * const cmd[], char **tem
 		goto err2;
 	}
 
-	hio_close(*h);
-	*h = hio_open_file2(t);
+	if ((tmp = hio_open_file2(t)) == NULL)
+		return -1;  /* call closes on failure. */
 
-	return (*h == NULL)? -1 : 0;
+	hio_close(*h);
+	*h = tmp;
+	return 0;
 
     err2:
 	fclose(t);
@@ -241,6 +244,7 @@ static int decrunch_command(HIO_HANDLE **h, const char * const cmd[], char **tem
 
 static int decrunch_internal_tempfile(HIO_HANDLE **h, struct depacker *depacker, char **temp)
 {
+	HIO_HANDLE *tmp;
 	FILE *t;
 
 	D_(D_WARN "Depacking file... ");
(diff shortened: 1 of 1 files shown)


sezero:
Current master (or #508) errors out with `Error depacking file` for me on i686-linux.

The patch is OK though.

sezero (approved):

AliceLR:
> Current master (or #508) errors out with `Error depacking file` for me on i686-linux.
> 
> The patch is OK though.

This is because there's a check in miniz to make sure `size_t` can hold the length. The underflow is really libxmp's fault for using `long` there instead of `size_t`.

I reported the Zip64 bug upstream (https://github.com/richgel999/miniz/issues/218). Notably, this bug is absent from `mz_zip_reader_extract_to_callback`, so we might need to switch back to that.

sezero:
> This is because there's a check in miniz to make sure `size_t` can hold the length. The overflow is really libxmp's fault for using `long` there instead of `size_t`.

We should have something like the following, I guess?
```diff
diff --git a/src/depackers/unzip.c b/src/depackers/unzip.c
index f510d18..fef0299 100644
--- a/src/depackers/unzip.c
+++ b/src/depackers/unzip.c
@@ -84,6 +84,11 @@ static int decrunch_zip(HIO_HANDLE *in, void **out, long inlen, long *outlen)
 		}
 
 		mz_zip_reader_end(&archive);
+		if (pSize > LONG_MAX) {
+			free(pBuf);
+			return -1;
+		}
+
 		*out = pBuf;
 		*outlen = pSize;
 		return 0;
```

> 
> I reported the Zip64 bug upstream ([richgel999/miniz#218](https://github.com/richgel999/miniz/issues/218)). Notably, this bug seems to be absent from `mz_zip_reader_extract_to_callback`, so we might need to switch back to that.

OK.  (Do you have a patch for it?)

AliceLR:
> We should have something like the following, I guess?

Maybe. Another option would be `mz_zip_reader_file_stat`, which would allow skipping files over other size thresholds.

> OK. (Do you have a patch for it?)

No, I don't have a patch (either for the upstream bug or for switching back to `mz_zip_reader_extract_to_callback`). Ideally the upstream bug would get fixed, which I don't think is *too* hard.

sezero:
We should create a ticket for this with relevant details, so that it doesn't get lost.

sezero:
> > This is because there's a check in miniz to make sure `size_t` can hold the length. The overflow is really libxmp's fault for using `long` there instead of `size_t`.
> 
> We should have something like the following, I guess?
> 
> ```diff
> diff --git a/src/depackers/unzip.c b/src/depackers/unzip.c
> index f510d18..fef0299 100644
> --- a/src/depackers/unzip.c
> +++ b/src/depackers/unzip.c
> @@ -84,6 +84,11 @@ static int decrunch_zip(HIO_HANDLE *in, void **out, long inlen, long *outlen)
>  		}
>  
>  		mz_zip_reader_end(&archive);
> +		if (pSize > LONG_MAX) {
> +			free(pBuf);
> +			return -1;
> +		}
> +
>  		*out = pBuf;
>  		*outlen = pSize;
>  		return 0;
> ```

> 
> > We should have something like the following, I guess?
> 
> Maybe. Another option would be `mz_zip_reader_file_stat`, which would allow skipping files over other size thresholds.

What is the best solution for this?

PR #528, opened 2021-12-28
more miscellenaous small stuff..

sezero:


AliceLR (approved):
This seems OK. (IMO Watcom can die mad about unreachable code but in this case it's not a very intrusive patch.)

PR #529, opened 2021-12-28
miniz: apply mainstream PR/159 to reduce inflate memory usage

sezero:
See:  https://github.com/richgel999/miniz/pull/159


AliceLR (approved):
This looks OK.

PR #530, opened 2021-12-29
more miscellaneous small stuff ...

sezero:
- list.h, vorbis.h: use inline instead of `__inline__`
  common.h already has fixes for it where necessary.
- add a libxmp_strdup() helper procedure.
- med2_load.c: fix static declaration followed by non-static definition.
- readlzw.c: fix static declaration followed by non-static definition.


AliceLR (approved):
This looks OK.

PR #531, opened 2021-12-29
Fix mixer crash from not checking sptr before wraparound init.

AliceLR:
There is a `NULL` check for `vi->sptr` before mixing sample data but I forgot to add one before initializing the sample wraparound data. (Found thanks to Win32 text mode `stdout` mangling the module from test-dev/data/arcfsdata.)


AliceLR:
I think codecov might be confused again 🥶

sezero (approved):

PR #532, opened 2021-12-31
vorbis: eliminate type-punned pointer use

sezero:
Fixes strict-aliasing warnings from older gcc versions:

src/loaders/vorbis.c: In function 'start_decoder':
src/loaders/vorbis.c:3831: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:3888: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:3889: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:3890: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:3969: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c: In function 'vorbis_deinit':
src/loaders/vorbis.c:4328: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:4329: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:4330: warning: dereferencing type-punned pointer will break strict-aliasing rules
src/loaders/vorbis.c:4331: warning: dereferencing type-punned pointer will break strict-aliasing rules

### DIFF
--- a/src/loaders/vorbis.c
+++ b/src/loaders/vorbis.c
@@ -929,10 +929,10 @@ struct stb_vorbis
    int channel_buffer_end;
 
   // temporary buffers
-   uint8 *temp_lengths;
-   uint32 *temp_codewords;
-   uint32 *temp_values;
-   uint16 *temp_mults;
+   void *temp_lengths;
+   void *temp_codewords;
+   void *temp_values;
+   void *temp_mults;
 };
 
 #if defined(STB_VORBIS_NO_PUSHDATA_API)
@@ -3786,8 +3786,10 @@ static int start_decoder(vorb *f)
       if (c->dimensions == 0 && c->entries != 0)    return error(f, VORBIS_invalid_setup);
       if (f->valid_bits < 0)                        return error(f, VORBIS_unexpected_eof);
 
-      if (c->sparse)
-         lengths = f->temp_lengths = (uint8 *) setup_temp_malloc(f, c->entries);
+      if (c->sparse) {
+         lengths = (uint8 *) setup_temp_malloc(f, c->entries);
+         f->temp_lengths = lengths;
+      }
       else
          lengths = c->codeword_lengths = (uint8 *) setup_malloc(f, c->entries);
 
@@ -3828,7 +3830,7 @@ static int start_decoder(vorb *f)
          c->codeword_lengths = (uint8 *) setup_malloc(f, c->entries);
          if (c->codeword_lengths == NULL) return error(f, VORBIS_outofmem);
          memcpy(c->codeword_lengths, lengths, c->entries);
-         setup_temp_free(f, (void **)&(f->temp_lengths), c->entries); // note this is only safe if there have been no intervening temp mallocs!
+         setup_temp_free(f, &f->temp_lengths, c->entries); // note this is only safe if there have been no intervening temp mallocs!
          lengths = c->codeword_lengths;
          c->sparse = 0;
       }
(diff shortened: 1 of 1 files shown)


AliceLR:
Which GCC versions emit this warning?

sezero:
gcc4.2 in my amigaos4 and aros toolchains

sezero:
Note: Newer gcc (e.g. gcc-9) do warn but with `-Wstrict-aliasing=1` or 2.

AliceLR:
> gcc4.2 in my amigaos4 and aros toolchains

> Note: Newer gcc (e.g. gcc-9) do warn but with `-Wstrict-aliasing=1` or 2.

OK, good to know. I remember getting spammed out with these a lot in another project for similar `(void **)` casts. Will verify and get back soon.

sezero:
> edit: GCC 11 with `-Wstrict-aliasing` (either) does not reproduce the warning—I wonder if they revised it again at some point?

It may change back with gcc12 or 13 in future, so no guarantees.

AliceLR:
> edit: GCC 11 with `-Wstrict-aliasing` (either) does not reproduce the warning—I wonder if they revised it again at some point?

`CFLAGS="-fstrict-aliasing -Wstrict-aliasing=2"` did it. `-Wstrict-aliasing[=3]` will NOT do it and for whatever reason (probably flag ordering in the Makefile) it didn't enable without explicitly adding `-fstrict-aliasing`.

AliceLR (approved):
This fixes the issue for me, merging. I've also updated the upstream patch.

AliceLR:
Would it be worth it to add `-Wstrict-aliasing` (either 1 or 2) to autoconf to make it easier to spot this in the future?

sezero:
> Would it be worth it to add `-Wstrict-aliasing` (either 1 or 2) to autoconf to make it easier to spot this in the future?

Not sure

AliceLR:
After experimenting with this warning a little bit elsewhere and reading the [manual](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html):

> Level 1: Most aggressive, quick, least accurate. Possibly useful when higher levels do not warn but -fstrict-aliasing still breaks the code, as it has very few false negatives. However, it has many false positives. Warns for all pointer conversions between possibly incompatible types, even if never dereferenced. Runs in the front end only.

> Level 2: Aggressive, quick, not too precise. May still have many false positives (not as many as level 1 though), and few false negatives (but possibly more than level 1). Unlike level 1, it only warns when an address is taken. Warns about incomplete types. Runs in the front end only.

> Level 3 (default for -Wstrict-aliasing): Should have very few false positives and few false negatives. Slightly slower than levels 1 or 2 when optimization is enabled. Takes care of the common pun+dereference pattern in the front end: *(int*)&some_float. If optimization is enabled, it also runs in the back end, where it deals with multiple statement cases using flow-sensitive points-to information. Only warns when the converted pointer is dereferenced. Does not warn about incomplete types.

If this gets enabled we would *probably* want level 2—it seems like 3 might have a performance impact. It also doesn't seem to catch cases like this where the dereference occurs away from the address, but that might not be very common in the use cases it's trying to prevent. Level 1 seems pretty much useless and it warns for very common (and valid C89) subtyping conversions that the others ignore:

```C
#include <stdio.h>

struct type
{
  int a;
};

struct subtype
{
  struct type t;
  int b;
};

static __attribute__((noinline)) int getb(struct type *t)
{
  struct subtype *s = (struct subtype *)t;
  return s->b;
};

int main()
{
  struct subtype _s = { 1, 2 };
  struct subtype *s = &_s;

  printf("%d\n", getb((struct type *)s));
  return 0;
};
```
```
$ gcc -fstrict-aliasing -Wstrict-aliasing=1 getb.c
getb.c: In function 'getb':
getb.c:16:31: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
   16 |   struct subtype *s = (struct subtype *)t;
      |                               ^~~~~~~
getb.c: In function 'main':
getb.c:25:31: warning: dereferencing type-punned pointer might break strict-aliasing rules [-Wstrict-aliasing]
   25 |   printf("%d\n", getb((struct type *)s));
      |                               ^~~~
```

As an aside, it seems like the "cleanest" way to hide the intermediate `void *` conversion is to just use a `union` like you're supposed to with the first field as a `void *`, but that wouldn't really help here.

AliceLR:
Actually, I missed this, but it seems -Wstrict-aliasing has defaulted to -Wstrict-aliasing=3 since GCC 4.3 and is already enabled by -Wall. There's probably no reason to change it.

sezero:
> Actually, I missed this, but it seems -Wstrict-aliasing has defaulted to -Wstrict-aliasing=3 since GCC 4.3 and is already enabled by -Wall. There's probably no reason to change it.

Yep.

PR #535, opened 2022-01-02
miniz_zip: fix mz_zip_reader_extract_to_heap to read correct sizes

sezero:
Original libxmp report: https://github.com/libxmp/libxmp/pull/527
See upstream issue: https://github.com/richgel999/miniz/issues/218


sezero:
If this is good, we can propose the solution to upstream.

sezero:
Patch now submitted to upstream as https://github.com/richgel999/miniz/pull/220

AliceLR (approved):
I think this is probably the most reasonable fix for this there is, and ponylips.64.zip does work as expected now. Merging.

PR #536, opened 2022-01-02
Add Zip64 regression test.

AliceLR:
Draft until I'm sure it covers remotely anything.

edit: it covers the central directory portions of Zip64 support but not the bulk of the stat portions, will update.


AliceLR:
I hex edited the ZIP archive to force miniz to handle Zip64 compressed sizes and offsets. Codecov won't display the coverage change on their website but it *should* be okay. The rest of the missing coverage in miniz_zip.c seems to be largely missing coverage of error paths.

sezero:
> I hex edited the ZIP archive to force miniz to handle Zip64 compressed sizes and offsets.

I don't understand this part: file is a zip64 archive, so what is changed to test what?

AliceLR:
> I don't understand this part: file is a zip64 archive, so what is changed to test what?

Not sure what you're asking for specifically so here's an overview of Zip64, what Info-ZIP `zip` output, and what I had to change:

* To get the initial Zip64 file I used `zip -fz ponylips.64.zip ponylips.mod`

* Zip64 Extended Information Extra Field: Zip64 adds this to the extra data area of both local and central directory file headers to contain 64-bit values for the uncompressed size, compressed size, and local header offsets for a file. In usual cases these will only be stored if their values don't fit in the original ZIP record (in which case the dummy `0xFFFFFFFF` is stored), but it's valid to use this extension even if they would fit, which this test relies on.

  `zip` wrote both 64-bit sizes to the local header when I forced it to use Zip64 but only wrote the 64-bit uncompressed size to the central directory file header, so I had to hex edit the archive to add the compressed size and local file header offset to the central directory file header's extra data area. This forces miniz to read these from the extra data area when it performs its stat routine, which it wasn't doing before, and this change bumped the coverage to >80%. This increased the size of the central directory by 16 bytes.
* Zip64 EOCD record: this is stored after the central directory file headers and prior to the Zip64 EOCD locator and stores extended copies of values from the standard ZIP EOCD record. `zip` wrote this as usual, I just had to hex edit the size of central directory field here to account for the extra bytes.
* Zip64 end of central directory (EOCD) locator: this is stored immediately prior to the standard ZIP EOCD record and stores an extended value for the total number of disks, the disk number for the Zip64 EOCD record, and the offset in the disk for the Zip64 EOCD record. `zip` also wrote this, so I just had to increase the offset for the Zip64 EOCD by 16 bytes like the central directory size.
* `zip` also wrote the central directory size to the ZIP EOCD as usual so I changed that to match the Zip64 EOCD.

Then, I tested it with 7-Zip, Info-ZIP `unzip`, and libxmp to make sure I didn't break anything.

sezero:
OK then

sezero (approved):

Issue #460, opened 2021-07-28
Add Beni Tracker module support


Sembiance:
Would be awesome if libxmp supported Beni Tracker. Info, file samples and more: http://fileformats.archiveteam.org/wiki/Beni_Tracker_module

I haven't been able to locate ANY modern player. The wiki link above has a link to adplug which seems to have some sample source code for playing the files, but doesn't actually appear to work with AdPlug.

Format info:
```
The file extension .PIS is used since the previous incarnation
of Beni Tracker was called "Psychisch Tracker" (unreleased).

Offset   Bytes    Description
------   -----    -----------
0        1        Song length
1        1        N. of patterns
2        1        N. of instruments
3        ?        Pattern map (N. of patterns * 1 byte)
?        ?        Instrument map (N. of instruments * 1 byte)
?        ?        Order list (Song length * 9 bytes)
?        ?        Pattern data (N. of patterns * 192 bytes)
?        ?        Instrument data (N. of instruments * 11 bytes)
?        4        "B.J." (v1.8 identification mark)

-Pattern and instrument maps
   PIS, the intelligent format it is, saves only the
   patterns and instruments that are used. The maps
   therefor contain for each pattern (instrument) in
   the file the number that was used for it during editing.
   Eg. if You have this in Your (otherwise empty) order list:
      02
      04
      00
   The pattern map, containing the indices of used patterns
   in ascending order, will look like this:
      ptnmap[0]=0
      ptnmap[1]=2
      ptnmap[2]=4
   Ergo, the maps contain the order in that the patterns
   (instruments) are saved in the data blocks below.

-Pattern data format
   A pattern consists of 64 * 3 bytes where 3 bytes encode
   one note as:

      nnnn.... ........ ........
      ....ooo. ........ ........
      .......i iiii.... ........
      ........ ....eeee eeeeeeee

   n = note (0..12) as index into frequency table, 12 is blank
   o = octave
   i = instrument
   e = effect

-Instrument format
   An instrument consists of 11 bytes.

   Offset   Description
   ------   -----------
   0        Op.1: Multiple/EG type/Vibrato/Amp.mod.
   1        Op.2:                 "
   2        Op.1: Output level
   3        Op.2:    "     "
   4        Op.1: Attack/Decay
   5        Op.2:       "
   6        Op.1: Sustain/Release
   7        Op.2:        "
   8        Op.1: Waveform
   9        Op.2:     "
   10       Feedback/Connection type

-"B.J."
[...]

PR #537, opened 2022-01-03
Remove AdLib/OPL formats from formats.txt

mmontag:
As mentioned in a few issues (e.g. https://github.com/libxmp/libxmp/issues/460#issuecomment-898948675), fmopl.c is dead code and the AdLib formats are not currently supported.

### DIFF
--- a/docs/formats.txt
+++ b/docs/formats.txt
@@ -104,13 +104,6 @@ PSM		Protracker Studio	0.01, 1.00
 STIM		Slamtilt		-
 UMX		Epic Games Unreal/UT	IT, S3M, MOD, XM
 
-YM3812 (Adlib) formats:
-
-AMD		Amusic Adlib Tracker	-
-RAD		Reality Adlib Tracker	-
-HSC		NEO soft/HSC-Tracker	1.5
-S3M		Scream Tracker 3	3.00, 3.01+
-
 
 Formats marked with (*) are experimental and are known to have
 replay errors and unimplemented effects.


AliceLR (approved):
Thanks. I'm not sure how this bit of formats.txt stuck around for so long.

mmontag:
Thanks; by the way, do you know of any alternative to Adplug for these OPL-based tracker formats?

AliceLR:
I don't know of any, sorry. OpenMPT has started picking up *some* niche OPL formats, but presumably its support is not very mature yet. If you're looking specifically for a RAD library, there is freely available C++ playercode for it with several bugfixes [here](https://github.com/AliceLR/megazeux/tree/master/contrib/rad).

These may be added back to libxmp at some time in the future but there's so many other things that need fixing that this probably won't happen soon :(

sagamusix:
> OpenMPT has started picking up some niche OPL formats, but presumably its support is not very mature yet

My focus is mainly on "formats noone else supports yet" that can be reasonably emulated within OpenMPT's S3M-centric view of OPL (which I understand is very different from many "pure" advanced AdLib trackers). RAD may be a bit too advanced for that at the moment, generally I think AdPlug is still the best when it comes to most pure OPL formats.

mmontag:
Thanks for the insights - I might look into implementing the RAD format, but batch VGM conversion might be the way to go (there's one by @SudoMaker https://github.com/SudoMaker/adlib2vgm).

AliceLR:
Proper RAD support might be possible in libxmp using its extras system but
unless OpenMPT has anything equivalent then riffs are basically just not
going to work without a lot of hacks. That said, they'd be a really nice
feature for MPTM, and a RAD v2 tracker that doesn't crash or have broken
KSL in Windows would be even better... : )

sagamusix:
That feature would definitely be RAD to have in OpenMPT as a proper feature, not just something hidden in the replayer for imported formats. ;)

PR #540, opened 2022-01-06
Replace nomarch LGPL-licensed ARC/ArcFS depackers.

AliceLR:
Replaces the old ARC/ArcFS depackers with new depackers I wrote. Here's a summary of their differences.

* The new depackers are permissive licensed.
* The new depackers have working decoders for compression methods 3 and 4. There wasn't much preventing these from being fixed in the nomarch code, though.
* The new depackers do *not* support compression methods 5-7 in any form, which is probably okay. I've never seen these ever. These were also dummied out in the nomarch code.
* The new depackers compute the ARC/ArcFS CRC-16 checksum. The nomarch depackers didn't bother with this.
* Depending on the input file, the new depackers are slower for uncompressed and RLE-packed (method 3) files due to the new CRC-16 (10-20x slower for unpacked, 3-4x slower for RLE). This is pretty much meaningless in practice because these methods are absurdly fast otherwise and are much faster than other methods even with the CRC-16.
* Depending on the input file, the new depackers seem to finish in about 30-50% of the time that the nomarch code does ***for compressed files***, due to various flaws in it. (This includes the CRC-16.)
  * The nomarch LZW decoder outputs single bytes into an output *function pointer*. Its bitstream implementation is also slow.
  * The nomarch Huffman decoder does not use an optimization table.
* The nomarch depackers don't support some edge case method FF compressed streams for reasons I don't remember, but I remember having to explicitly fix it in my code.
* The new depackers should have *much* better support for ARC6, Spark, and ArcFS subdirectories.
* The new ARC depacker should be aware of PAK and ARC+ archives, but it doesn't really support these.
* The new depackers were both individually fuzzed a bit but aren't guaranteed to be 100% bug free. The last runs I did for each I let go to a little over 30 million execs.

I also updated some info in docs/CREDITS.

~~Draft until I get around to adding method 3/4/9/FF regression test cases and determining which fuzzer files are still useful.~~


AliceLR:
The tests are in now, but I probably need a better method 4 example (this one didn't emit codes >11 bits long).

AliceLR:
I went through the fuzzer files and ended up hex editing a lot of them to be mostly valid so they'd be useful again. A couple were for fairly nomarch-specific UMR bugs so I just deleted them.

This should be ready to review now.

src/depackers/arc_unpack.h
```diff
@@ -0,0 +1,119 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#include <stddef.h> /* size_t */
+
+/* TODO libxmp should probably detect this in autoconf. */
+#if 0
+#define ARC_RESTRICT __restrict__
```
sezero, line 39: Maybe this? (lodepng does this)
```c
#if (defined(__GNUC__) && (__GNUC__ > 3 || (__GNUC__ == 3 && __GNUC_MINOR__ >= 1))) ||\
    (defined(_MSC_VER) && (_MSC_VER >= 1400)) || \
    (defined(__WATCOMC__) && (__WATCOMC__ >= 1250) && !defined(__cplusplus))
#define ARC_RESTRICT __restrict
#else
#define ARC_RESTRICT
#endif
```

src/depackers/arcfs.c
```diff
@@ -1,212 +1,335 @@
...
+  {
+#ifdef ARCFS_DEBUG
+    debug("short read in header\n");
+#endif
+    return -1;
+  }
+
+  if(arcfs_check_magic(buffer) < 0)
+  {
+#ifdef ARCFS_DEBUG
```
sezero, line 121: The ifdefs need indenting for consistency with other arc sources

AliceLR: OK, fixed.

src/depackers/arc_unpack.h
AliceLR: I assumed it would be more annoying to reliably detect than that. In that case, would it be worth adding a `LIBXMP_RESTRICT` or similar to src/common.h? Presumably some other things in libxmp could benefit from `restrict` as well.

sezero: Oh yes.

AliceLR: OK, I added `LIBXMP_RESTRICT` for this.

sezero (approved):
Great, let's get this in.

PR #543, opened 2022-01-12
Add support for MMDC packed OctaMED modules.

### LINKED ISSUE #18: Make support MMDC
Make support MMDC Med Packer

AliceLR:
I came across some of these while looking for example LZX files. They're MMD0 with some basic block packing that was very easy to reverse engineer. Playback sounds accurate to the longplay videos I've found for the games that used this format, but further tweaks *might* be necessary. I couldn't actually find whatever the MMDC packer's name is (if any) but there are some references to "MED Packer", so I went with that.

To make this less painful I had to change the way MMD0/1 patterns are loaded a bit—they now use a shared pattern buffer instead of reading bytes one at a time. This change wasn't very difficult so I also applied it to the MMD2/3 loader.

Fixes #18.

edit: coverage decrease due to slightly more missed error paths, removal of some repetitive covered lines in the pattern loaders, and apparently a couple of effects that were covered by patterns in invalid modules no longer being covered.


src/loaders/mmd1_load.c
```diff
@@ -401,85 +425,102 @@ static int mmd1_load(struct module_data *m, HIO_HANDLE *f, const int start)
 			block.lines = hio_read8(f);
 		}
 
-		/* Sanity check--Amiga OctaMED files have an upper bound of 3200 lines per block. */
-		if (block.lines + 1 > 3200) {
-			D_(D_CRIT "invalid line count %d in block %d", block.lines + 1, i);
-			goto err_cleanup;
+		size = block.numtracks * (block.lines + 1) * (ver ? 4 : 3);
+
+		if (mmdc > 0) {
```
sezero, line 430: is this not a boolean?

AliceLR: It is. I think I wrote it like that to match up with the `ver` check below it, but I can change it.

sezero (approved):
Let's get this in.

Issue #542, opened 2022-01-07
error returns from read24 procs


sezero:
In dataio.c, we return `0xffffff` in error case from read24b and read24l,
but from callbackio.h and mdataio.h we return `0xffffffff` (`EOF` really).
Which should we really do?

On a related note: My gcc3.4 emits the following warnings in C++ mode:
Nothing major, but wanted to note it here anyway.
```
In file included from src/hio.h:4,
                 from src/hio.c:25:
src/callbackio.h: In function `uint16 cbread16l(CBFILE*, int*)':
src/callbackio.h:34: warning: converting of negative value `-0x000000001' to `uint16'
src/callbackio.h: In function `uint16 cbread16b(CBFILE*, int*)':
src/callbackio.h:47: warning: converting of negative value `-0x000000001' to `uint16'
src/callbackio.h: In function `uint32 cbread24l(CBFILE*, int*)':
src/callbackio.h:60: warning: converting of negative value `-0x000000001' to `uint32'
src/callbackio.h: In function `uint32 cbread24b(CBFILE*, int*)':
src/callbackio.h:73: warning: converting of negative value `-0x000000001' to `uint32'
src/callbackio.h: In function `uint32 cbread32l(CBFILE*, int*)':
src/callbackio.h:86: warning: converting of negative value `-0x000000001' to `uint32'
src/callbackio.h: In function `uint32 cbread32b(CBFILE*, int*)':
src/callbackio.h:99: warning: converting of negative value `-0x000000001' to `uint32'
In file included from src/hio.c:27:
src/mdataio.h: In function `uint16 mread16l(MFILE*, int*)':
src/mdataio.h:45: warning: converting of negative value `-0x000000001' to `uint16'
src/mdataio.h: In function `uint16 mread16b(MFILE*, int*)':
src/mdataio.h:60: warning: converting of negative value `-0x000000001' to `uint16'
src/mdataio.h: In function `uint32 mread24l(MFILE*, int*)':
src/mdataio.h:75: warning: converting of negative value `-0x000000001' to `uint32'
src/mdataio.h: In function `uint32 mread24b(MFILE*, int*)':
src/mdataio.h:90: warning: converting of negative value `-0x000000001' to `uint32'
src/mdataio.h: In function `uint32 mread32l(MFILE*, int*)':
src/mdataio.h:105: warning: converting of negative value `-0x000000001' to `uint32'
src/mdataio.h: In function `uint32 mread32b(MFILE*, int*)':
[...]

PR #544, opened 2022-01-12
miscellenaous small stuff

sezero:
Closes: https://github.com/libxmp/libxmp/issues/542


AliceLR (approved):
Wish there was a more future-proof solution for `xxi->map` changes but even with casts they'd break if that field ever gets expanded (some formats support more than 255 samples and use instrument mapping tables, I believe). Too bad `typeof` is an extension and `decltype` is C++...

PR #547, opened 2022-01-16
Use const CRC32 table, use libxmp CRC32 for miniz, and move CRC16.

AliceLR:
Makes various tweaks to libxmp's CRC support:

* libxmp was using an `extern` CRC-32 table for some reason, and required a separate call to initialize it at runtime. This is pretty pointless as the CRC-32 LSB table is well known and is never going to change.  It's `static const` now, with deadcode to regenerate it if required.
* Using a loop of unrolled CRC iterations for the CRC calculation to get a minor speed boost is pretty standard. After some testing, it seems this performance boost is small, and using type punning or unrolling more iterations either doesn't improve it much or can even make it slower. I went with a loop of only 4 unrolled CRC calculations.
* Added defines to make miniz use libxmp's CRC-32 implementation instead of yet another duplicate.
* I moved the ARC CRC-16 to crc32.c because it turns out LHA uses the same CRC (CRC-16-IBM). This allowed further merging of the ARC headers.
* I tried to make this patch as painless for #539 as possible but I don't think it was possible to avoid a merge conflict. Since no relevant archive formats aside from bunzip2 use the MSB CRC-32 I didn't bother updating it.

Some possible improvements I didn't bother with here:

* ARMv8+crc and ARMv8.1 builds have hardware instructions available for the original CRC-32. I have a [proof of concept for CRC-32C](https://gist.github.com/AliceLR/8e872607036fe77edd2ea58e8c7945c4) but I didn't bother with it here.


sezero:
I'm happy with this. Minor issue:

C++ build fails, so we need the following with which it builds OK:
```diff
diff -u b/src/depackers/miniz_zip.h b/src/depackers/miniz_zip.h
--- b/src/depackers/miniz_zip.h
+++ b/src/depackers/miniz_zip.h
@@ -22,10 +22,12 @@
 
 #include "crc32.h"
 #define USE_EXTERNAL_MZCRC
+LIBXMP_BEGIN_DECLS
 static inline mz_ulong mz_crc32(mz_ulong crc, const mz_uint8 *ptr, size_t buf_len)
 {
     return libxmp_crc32_A(ptr, buf_len, crc);
 }
+LIBXMP_END_DECLS
 #endif /* LIBXMP-SPECIFIC */
 
 /* ------------------- ZIP archive reading/writing */
```

However, I found a compiler (vbcc) which fails in another way:
```
>mz_ulong mz_crc32(mz_ulong crc, const mz_uint8 *ptr, size_t buf_len);
error 28 in line 60 of "src/depackers/miniz_zip.c": redeclaration of var <mz_crc32> with new storage-class
```
I think we can molest miniz_zip.c a bit more by removing all of its
mz_crc stuff and add ours as a static procedure in it instead of an
inline in miniz_zip.h.

AliceLR:
OK, I'll do that. I kind of figured one of the older compilers and/or C++ compilers would complain about that (the force push was getting rid of some trailing commas since C++ compilers hate those too...).

Other changes that could be part of this but I didn't bother: the gzip and LHA depackers don't actually calculate their CRCs currently. Both are due to get attention soon anyway though.

AliceLR:
OK, pushed a patch mostly similar to yours, but I changed a couple of things. It *seems* fine with g++ on my end.
* Added `LIBXMP_BEGIN_DECLS` and `LIBXMP_END_DECLS` to crc32.h.
* Moved the crc32.h include up a little so `#ifdef __cplusplus` could stay where it is.
* Fixed an unrelated trailing comma in arc_unpack.h that I missed.

Speaking of C++ builds, should we get rid of all of the `register` usage in mix_all.c or at least make it `LIBXMP_REGISTER` or something? I can't imagine it's actually helping anything in modern compilers. (The other place `register` is still used is unlzx.c which is getting a PR right after this is in anyway...)

sezero (approved):
Great!

sezero:
> Speaking of C++ builds, should we get rid of all of the `register` usage in mix_all.c or at least make it `LIBXMP_REGISTER` or something? I can't imagine it's actually helping anything in modern compilers.

That's a good idea. Maybe something like https://github.com/sezero/libmodplug/commit/c0db4e2c053f5ccd962c5a3ff7033adafe58d520 or just remove?

AliceLR:
> That's a good idea. Maybe something like [sezero/libmodplug@c0db4e2](https://github.com/sezero/libmodplug/commit/c0db4e2c053f5ccd962c5a3ff7033adafe58d520) or just remove?

Sorry, missed this. After taking a closer look, there are still two usages.

1. It's used for [two pointers](https://cdn.petsmagazine.com.sg/img/articles/dreamstime_xxl_33707079_edit.jpg) in the unzoo.c deadcode that sane compilers should be assigning registers to regardless (or even optimizing the inner loop to an inline `memcpy`).
2. The aforementioned [temporary sample value](https://github.com/libxmp/libxmp/blob/master/src/mix_all.c#L152) in the software mixer functions. Same thing, I'd hope...

The usage in mix_all.c has been around for at least 21 years with no modification. Unless there are still compilers in regular use that actually need that hint I think `register` can just be removed from both places.

sezero:
>  It's used for [two pointers](https://cdn.petsmagazine.com.sg/img/articles/dreamstime_xxl_33707079_edit.jpg)

Made my day :))

>  The aforementioned [temporary sample value](https://github.com/libxmp/libxmp/blob/master/src/mix_all.c#L152) in the software mixer functions. Same thing, I'd hope...
> 
> The usage in mix_all.c has been around for at least 21 years with no modification. Unless there are still compilers in regular use that actually need that hint I think `register` can just be removed from both places.

As you wish.

PR #548, opened 2022-01-16
Add permissive licensed Amiga LZX depacker, remove XAD LZX code.

AliceLR:
Replaces the XAD LZX depacker with a custom one I wrote. Changes worth noting:

* The new depacker uses libxmp's MIT license.
* Supports stored files (test included). The XAD depacker had support for this but it looks like there was some libxmp-specific modification that disabled it. These aren't very common.
* Supports merged files (test included). The original unlzx.c allegedly supported this, so it might have been broken either by XAD or by libxmp. These are fairly common in LZX packed modules since [most examples I could find](http://wt.exotica.org.uk) are bundles of multiple modules. I also suspect LZX is capable of packing e.g. readme files into a merge with a module if they are similar enough, but I wasn't able to get it to do this.
* When both the old and new depackers are built at -O3 with modern GCC (MinGW-w64, both i686 and x86_64), the new depacker completes in 50-67% of the time the old depacker takes. Still not sure why, but some ideas:
  * Faster bit reversal for canonical Huffman lookups (unverified).
  * Better aliasing characteristics due to smaller functions and `restrict`.
  * Builds with 64-bit `size_t` get a slight boost due to reading 3 words from the stream at a time.
* The new code has been fuzzed quite a bit, but as usual is not guaranteed to be free of bugs.
* Coverage is around 82% and 89% for the file parsing code and decompressing code, respectively. Most of the missed branches are related to error checks or to cases where files/merges would be skipped. edit: also, one of the test files I chose uses a MED effect that wasn't covered before...

(Note: the "unlzx.c" referenced in some source comments is not the XAD depacker, but rather the *original* unlzx.c [available on Aminet](http://aminet.net/misc/unix/unlzx.c.gz) and featured on the [Amiga LZX website](http://xavprods.free.fr/lzx/). XAD appears to have mostly copied that file and replaced its credits with a GPL notice. I haven't found any license info for unlzx.c so I only lightly referenced it for my depacker.)


sezero (approved):

sezero:
I get `src/depackers/lzx_unpack.c(265): Warning! W201: Unreachable code`
from watcom, which is blah and we can actually live with it unless there is some
compile-time check (`defined(_LP64) || defined(_WIN64)` maybe??  Or really 
just ignore and merge.)

AliceLR:
> I get `src/depackers/lzx_unpack.c(265): Warning! W201: Unreachable code` from watcom

That's kind of the point Watcom :(

> unless there is some compile-time check (`defined(_LP64) || defined(_WIN64)` maybe?? Or really just ignore and merge.)

Not familiar enough with the `_LP64` define to know if it's adequate to detect every case (there's at least [one SO answer](https://stackoverflow.com/a/70140862) that warns against it). For MegaZeux I ended up putting this horrible thing together and it probably still doesn't cover everything.
```C
/**
 * Also try to get the platform bit width.
 * Emscripten natively supports 64-bit math when compiling to Wasm.
 */
#if defined(_WIN64) || defined(__EMSCRIPTEN__) || \
  (defined(__sparc__) && defined(__arch64__)) || \
  ((defined(__riscv) || defined(__riscv__)) && __riscv_xlen >= 64) || \
  ((defined(__mips__) || defined(__mips) || defined(__MIPS__)) && \
    defined(_MIPS_SIM) && defined(_ABI64) && _MIPS_SIM == _ABI64) || \
  (defined(__GNUC__) && \
    (defined(__x86_64__) || defined(__powerpc64__) || defined(__PPC64__) || \
     defined(__aarch64__) || defined(__alpha__)))
#define ARCHITECTURE_BITS 64
#else
#define ARCHITECTURE_BITS 32
#endif
```

sezero:
> > I get `src/depackers/lzx_unpack.c(265): Warning! W201: Unreachable code` from watcom
> 
> That's kind of the point Watcom :(

Heh :)

> > unless there is some compile-time check (`defined(_LP64) || defined(_WIN64)` maybe?? Or really just ignore and merge.)
> 
> Not familiar enough with the `_LP64` define to know if it's adequate to detect every case (there's at least [one SO answer](https://stackoverflow.com/a/70140862) that warns against it).

The LLP64 case mentioned in there is already covered by `_WIN64`
Apart from that, `LONG_MAX > 0x7fffffffL` can be good check.

>  For MegaZeux I ended up putting this horrible thing together and it probably still doesn't cover everything.

Oh boy..

> ```c
> /**
>  * Also try to get the platform bit width.
>  * Emscripten natively supports 64-bit math when compiling to Wasm.
>  ```

wasm32 ??

AliceLR:
> The LLP64 case mentioned in there is already covered by `_WIN64` Apart from that, `LONG_MAX > 0x7fffffffL` can be good check.

Looks annoying to cover all of the data model macros but since `LONG_MAX` is guaranteed to be a macro maybe a simplified version like this would work. The only special cases required are for Windows (LLP64) and x32/etc. (ILP32).
```C
#if defined(_WIN64)
typedef size_t buffertype;
#define BUFFERTYPE_IS_64
#elif (defined(_ILP32) || defined(__ILP32__)) && defined(LLONG_MAX)
typedef unsigned long long buffertype;
#define BUFFERTYPE_IS_64
#else
typedef unsigned long buffertype;
#if LONG_MAX > 0x7fffffffL
#define BUFFERTYPE_IS_64
#endif
#endif
```

If `_WIN64` guarantees `long long` is available like it *seems* like it should, this could be simplified further. (Alternatively, the ILP32 bits are untested and could be dropped.)
```C
#if defined(_WIN64) || ((defined(_ILP32) || defined(__ILP32__)) && defined(LLONG_MAX))
typedef unsigned long long buffertype;
#define BUFFERTYPE_IS_64
#else
typedef unsigned long buffertype;
#if LONG_MAX > 0x7fffffffL
#define BUFFERTYPE_IS_64
#endif
#endif
```

The second version seems to work fine with my MINGW64 setup and with MSVC/nmake when I force it to do a x64 build.

> it probably still doesn't cover everything.

realized that i forgot PA-RISC 2.0 💀

> wasm32 ??

IIRC I found it wasn't relevant for MegaZeux when I added that, at least.

sezero:
Using long long with ILP32 can bring in unnecessary 64 bit math on
32 bit arches, no?  The following should be enough:
```c
typedef size_t buffertype;
#if defined(_WIN64) || defined(_LP64) || (LONG_MAX > 0x7fffffffL)
#define BUFFERTYPE_IS_64
#endif
```

AliceLR:
> Using long long with ILP32 can bring in unnecessary 64 bit math on 32 bit arches, no? The following should be enough:

They should not be defining ILP32 macros on 32-bit architectures, but maybe they do. ILP32 is a [64-bit data model](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models) for using 32-bit pointers. Removing it simplifies things and it's an extreme edge case I haven't tested, so I'll just go ahead and drop it.

As a side note, `size_t` has a badly defined width relative to `unsigned long`—it can theoretically be 16-bit, but in practice it should never be unless you're targeting real mode or something.

`_LP64` is compiler-specific and in this situation seems redundant if `LONG_MAX` (ANSI C) indicates it is wider than 32-bit. If `LONG_MAX` does not correctly reflect the size of `long` then the environment it is defined in is broken/not ANSI C compliant.

(Will update this patch shortly, sorry.)

sezero:
>  Is it just there to guard against more annoying Watcom warnings?

Which one, `_LP64`? Not relevant to watcom. And you are right that `LONG_MAX`
check already covers `_LP64`.

AliceLR:
OK, patch pushed. I also fixed an annoying warning in callbackio.c that affected x64 MSVC builds.

sezero:
Great, let's get this in.

PR #549, opened 2022-01-16
miscellaneous clean-ups

sezero:
This stupid PR is a result after my stupid attempts at building libxmp
with vbcc, which is capable of emiting really annoying warnings. After
this patchset it still emits only a few warnings (from imported sources
i.e. miniz_zip and vorbis), and the build actually fails because I can
not find a way around the missing `alloca` support in vorbis.c (I first
attempted using C99 VLAs but failed doing so - I possibly haven't tried
hard enough.)

Anyways, here it is.

### DIFF
--- a/src/mixer.c
+++ b/src/mixer.c
@@ -941,7 +941,7 @@ int libxmp_mixer_on(struct context_data *ctx, int rate, int format, int c4rate)
 	s->format = format;
 	s->amplify = DEFAULT_AMPLIFY;
 	s->mix = DEFAULT_MIX;
-	/* s->pbase = C4_PERIOD * c4rate / s->freq; */(void) c4rate;
+	/* s->pbase = C4_PERIOD * c4rate / s->freq; */
 	s->interp = XMP_INTERP_LINEAR;	/* default interpolation type */
 	s->dsp = XMP_DSP_LOWPASS;	/* enable filters by default */
 	/* s->numvoc = SMIX_NUMVOC; */
--- a/lite/watcom.mif.in
+++ b/lite/watcom.mif.in
@@ -65,6 +65,7 @@ clean: .symbolic
 	rm -f $(TEST_OBJS)
 
 distclean: clean .symbolic
+	rm -f *.err
 	rm -f $(DLLNAME) $(EXPNAME) $(MAPNAME) $(LIBNAME) $(LIBSTATIC) test/$(DLLNAME) test/$(TESTNAME)
 
 !ifdef __UNIX__
--- a/src/depackers/miniz_zip.c
+++ b/src/depackers/miniz_zip.c
@@ -35,12 +35,10 @@ extern "C" {
 
 #ifndef MINIZ_DISABLE_ZIP_READER_CRC32_CHECKS
 #define MZ_CRC32_INIT (0)
-/* libxmp hack */
 static mz_ulong mz_crc32(mz_ulong crc, const mz_uint8 *ptr, size_t buf_len)
-{
+{/* libxmp hack: */
     return libxmp_crc32_A(ptr, buf_len, crc);
 }
-/* end libxmp hack */
 #endif /* MINIZ_DISABLE_ZIP_READER_CRC32_CHECKS */
 
 static void *miniz_def_alloc_func(void *opaque, size_t items, size_t size)
--- a/src/depackers/xz.h
+++ b/src/depackers/xz.h
@@ -22,7 +22,7 @@ extern "C" {
 
 /* In Linux, this is used to make extern functions static when needed. */
 #ifndef XZ_EXTERN
-#	define XZ_EXTERN extern
+#	define XZ_EXTERN /*extern*/
 #endif
 
 /**
--- a/src/loaders/digi_load.c
+++ b/src/loaders/digi_load.c
@@ -73,7 +73,7 @@ struct digi_header {
     uint8 ver;			/* Version hi-nibble.lo-nibble */
     uint8 chn;			/* Number of channels */
     uint8 pack;			/* PackEnable */
-    uint8 unknown[19];		/* ??!? */
+    uint8 unknown[19];		/* ?! */
     uint8 pat;			/* Number of patterns */
     uint8 len;			/* Song length */
     uint8 ord[128];		/* Orders */
--- a/src/loaders/imf_load.c
+++ b/src/loaders/imf_load.c
@@ -338,7 +338,7 @@ static int imf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 
 	pat_len = hio_read16l(f) - 4;
 
-        rows = hio_read16l(f);
+	rows = hio_read16l(f);
 
 	/* Sanity check */
 	if (rows > 256) {
@@ -370,7 +370,7 @@ static int imf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		n = hio_read8(f);
 		switch (n) {
 		case 255:
-		case 160:	/* ??!? */
+		case 160:	/* ?! */
 		    n = XMP_KEY_OFF;
 		    break;	/* Key off */
 		default:
@@ -437,7 +437,7 @@ static int imf_load(struct module_data *m, HIO_HANDLE *f, const int start)
 
 	xxi->nsm = ii.nsm;
 
-        if (xxi->nsm > 0) {
+	if (xxi->nsm > 0) {
 	    if (libxmp_alloc_subinstrument(mod, i, xxi->nsm) < 0)
 		return -1;
 	}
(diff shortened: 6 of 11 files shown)


sezero:
Noticed a curiosity: the os/2 dll build with this patch is about 296 bytes smaller
than the one built without. The affected source is vorbis.c where the dummy `(void)0`
statements are replaced with dummy `do {} while(0)` statements. Huh... The gcc
outputs are the same though. Some quirk with watcom, I guess. Anyone has any
possible insight?

sezero:
> Noticed a curiosity: the os/2 dll build with this patch is about 296 bytes smaller than the one built without. The affected source is vorbis.c where the dummy `(void)0` statements are replaced with dummy `do {} while(0)` statements. Huh... The gcc outputs are the same though. Some quirk with watcom, I guess. Anyone has any possible insight?

Reported this to openwatcom: https://github.com/open-watcom/open-watcom-v2/issues/799
However, even though the generated vorbis code isn't identical, the dll seems to be working
(plays jerry-boleti.oxm from test-dev on my VirtualBox'ed OS/2.)

AliceLR:
> avoid stupid trigraph warnings from some stupid and silly compilers

😅

> Noticed a curiosity: the os/2 dll build with this patch is about 296 bytes smaller than the one built without. The affected source is vorbis.c where the dummy `(void)0` statements are replaced with dummy `do {} while(0)` statements. Huh... The gcc outputs are the same though. Some quirk with watcom, I guess. Anyone has any possible insight?

I'm not familiar with Watcom, but since `(void)statement;` is a fairly standard way to shut up certain compiler warnings, maybe Watcom emits code there intentionally (to somehow suppress the warning) or by mistake (it doesn't realize it's effectively deadcode unless the parameter is a function call or a `volatile` variable).

src/mixer.c
```diff
@@ -941,7 +941,7 @@ int libxmp_mixer_on(struct context_data *ctx, int rate, int format, int c4rate)
 	s->format = format;
 	s->amplify = DEFAULT_AMPLIFY;
 	s->mix = DEFAULT_MIX;
-	/* s->pbase = C4_PERIOD * c4rate / s->freq; */(void) c4rate;
+	/* s->pbase = C4_PERIOD * c4rate / s->freq; */
```
AliceLR, line 944: Likely to have been to suppress `-Wunused-parameter`? Which isn't a *totally* useless warning, but libxmp rightly turns it off now.

AliceLR (approved):

src/mixer.c
sezero: indeed

sezero:
Can you see a way of using C99 VLA instead of alloca in stb_vorbis?
I found it hard the first time I tried because it relies on alloca semantics.

AliceLR:
> Can you see a way of using C99 VLA instead of alloca in stb_vorbis? I found it hard the first time I tried because it relies on alloca semantics.

I don't think any solution to this is going to be particularly clean. It looks like the intended opt-out is to use `stb_vorbis_alloc`, but this also affects the setup buffer allocation. IMO both VLAs and `alloca` are kind of bad, particularly for this sort of usage:

* `inverse_mdct`: this could be a single allocation reused for the entire decode, and there's no reason to put it on the stack. The buffer here is sized `(n >> 1) * sizeof(float)`. From the half a dozen other stb_vorbis patches, we know `n` is famously bounded by `blocksize_1`, which can never be more than 8192 and is known during setup. So IMO, just allocate the <=32k array at startup and don't make embedded systems users wonder why their audio thread crashed loading an OXM.
* `decode_residue`: there are two here, both of which are effectively `f->channels * part_read * sizeof(void *)`. `part_read` is derived from `n_read` should be bounded by `blocksize_1` (8192), `f->channels` is bounded by `STB_VORBIS_MAX_CHANNELS` (16), so up to 1MB each. There's no excuse for 2 arrays this big to be on the stack, and the (likely much smaller) bound can once again be calculated during setup. Even better, one of the required work buffers can be shared with `inverse_mdct` if needed.

I should note both functions are mandatory steps of the decode.

sezero:
> I don't think any solution to this is going to be particularly clean. It looks like the intended opt-out is to use `stb_vorbis_alloc`, but this also affects the setup buffer allocation. IMO both VLAs and `alloca` are kind of bad, particularly for this sort of usage:
> 
>     * `inverse_mdct`: this could be a single allocation reused for the entire decode. The buffer here is sized `(n >> 1) * sizeof(float)`. From the half a dozen other stb_vorbis patches, we know `n` is famously bounded by `blocksize_1`, which can never be more than 8192 and is known during setup. So, just allocate the <=16k array at startup.
> 
>     * `decode_residue`: there are two here, both of which are effectively `f->channels * part_read * sizeof(void *)`. `part_read` is derived from `n_read` should be bounded by `blocksize_1` (8192), `f->channels` is bounded by `STB_VORBIS_MAX_CHANNELS` (16), so up to 1MB each. There's no excuse for 2 arrays this big to be on the stack, and the (likely much smaller) bound can once again be calculated during setup. Even better, one of the required work buffers can be shared with `inverse_mdct` if needed.

I hear you.

In the meantime, the following VLA solution does work: (and it was easy, I
probably was half asleep during my first failed attempt.)
```diff
diff --git a/src/loaders/vorbis.c b/src/loaders/vorbis.c
index b7fe69b..29ee6e8 100644
--- a/src/loaders/vorbis.c
+++ b/src/loaders/vorbis.c
@@ -962,12 +962,17 @@ static int error(vorb *f, enum STBVorbisError e)
 
 #define array_size_required(count,size)  (count*(sizeof(void *)+(size)))
 
-#define temp_alloc(f,size)              (f->alloc.alloc_buffer ? setup_temp_malloc(f,size) : alloca(size))
+#ifdef HAVE_ALLOCA
+#define temp_alloc(f,size,stakbuf)      (f->alloc.alloc_buffer ? setup_temp_malloc(f,size) : alloca(size))
+#else
+#define temp_alloc(f,size,stakbuf)      (f->alloc.alloc_buffer ? setup_temp_malloc(f,size) : stakbuf)
+#endif
 #define temp_free(f,p)                  do {} while (0)
 #define temp_alloc_save(f)              ((f)->temp_offset)
 #define temp_alloc_restore(f,p)         ((f)->temp_offset = (p))
 
-#define temp_block_array(f,count,size)  make_block_array(temp_alloc(f,array_size_required(count,size)), count, size)
+#define temp_block_array(f,count,size,stakbuf) \
+        make_block_array(temp_alloc(f,array_size_required(count,size),stakbuf), count, size)
 
 // given a sufficiently large block of memory, make an array of pointers to subblocks of it
 static void *make_block_array(void *mem, int count, int size)
@@ -2160,11 +2165,30 @@ static void decode_residue(vorb *f, float *residue_buffers[], int ch, int n, int
    int n_read = limit_r_end - limit_r_begin;
    int part_read = n_read / r->part_size;
    int temp_alloc_point = temp_alloc_save(f);
+
    #ifndef STB_VORBIS_DIVIDES_IN_RESIDUE
-   uint8 ***part_classdata = (uint8 ***) temp_block_array(f,f->channels, part_read * sizeof(**part_classdata));
+   uint8 ***part_classdata;
+   #ifndef HAVE_ALLOCA
+   const int sz = f->alloc.alloc_buffer ? 1 : part_read * sizeof(**part_classdata);
+   uint8 **stk_array[sz];
    #else
-   int **classifications = (int **) temp_block_array(f,f->channels, part_read * sizeof(**classifications));
+   const int sz = part_read * sizeof(**part_classdata);
+   #define stk_array NULL
    #endif
+   part_classdata = (uint8 ***) temp_block_array(f,f->channels, sz, stk_array);
+
+   #else
+   int **classifications;
+   #ifndef HAVE_ALLOCA
+   const int sz = f->alloc.alloc_buffer ? 1 : part_read * sizeof(**classifications);
+   int *stk_array[sz];
+   #else
+   const int sz = part_read * sizeof(**classifications);
+   #define stk_array NULL
+   #endif
+   classifications = (int **) temp_block_array(f,f->channels, sz, stk_array);
+   #endif
+   #undef stk_array
 
    CHECK(f);
 
@@ -2678,11 +2702,21 @@ static void inverse_mdct(float *buffer, int n, vorb *f, int blocktype)
    int ld;
    // @OPTIMIZE: reduce register pressure by using fewer variables?
    int save_point = temp_alloc_save(f);
-   float *buf2 = (float *) temp_alloc(f, n2 * sizeof(*buf2));
+   float *buf2;
    float *u=NULL,*v=NULL;
    // twiddle factors
    float *A = f->A[blocktype];
 
+   #ifndef HAVE_ALLOCA
+   const int sz = f->alloc.alloc_buffer ? 1 : (n2 * sizeof(*buf2));
+   float stk_var[sz];
+   #else
+   #define stk_var NULL
+   const int sz = (n2 * sizeof(*buf2));
+   #endif
+   buf2 = (float *) temp_alloc(f, sz, stk_var);
+   #undef stk_var
+
    // IMDCT algorithm from "The use of multirate filter banks for coding of high quality digital audio"
    // See notes about bugs in that paper in less-optimal implementation 'inverse_mdct_old' after this function.
 
```
It of course requires `HAVE_ALLOCA` detection but whatever.

PR #552, opened 2022-01-18
Replace alloca() usage in stb_vorbis with setup_malloc().

AliceLR:
Alternative patch to #551: instead of using `alloca` or a VLA, this patch instead uses a temporary memory bound that stb_vorbis *already calculated* to `setup_malloc` a correctly-sized work buffer for `inverse_mdct` and `decode_residue`.

I'm going to fuzz it a little bit just to be sure the bound isn't missing anything.

### DIFF
--- a/cmake/libxmp-checks.cmake
+++ b/cmake/libxmp-checks.cmake
@@ -135,7 +135,6 @@ endif()
 xmp_check_function(powf "math.h" HAVE_POWF)
 cmake_pop_check_state()
 
-xmp_check_include(alloca.h HAVE_ALLOCA_H)
 xmp_check_function(popen "stdio.h" HAVE_POPEN)
 xmp_check_function(fnmatch "fnmatch.h" HAVE_FNMATCH)
 xmp_check_function(umask "sys/stat.h" HAVE_UMASK)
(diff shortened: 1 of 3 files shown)


sezero:
It was this easy? Man.. And I don't think there would be noticeable performance
degradation because of moving from stack to heap.

Just remove the alloca checks from configure.ac and cmake, and we're good
to go: [patch.txt](https://github.com/libxmp/libxmp/files/7889375/patch.txt)

sezero (approved):

AliceLR:
> It was this easy? Man.. And I don't think there would be noticeable performance degradation because of moving from stack to heap.

Yeah, I wasn't expecting to find that stb_vorbis is *already* doing the bound, which would have been the hard/annoying part to get right...

> Just remove the alloca checks from configure.ac and cmake, and we're good to go: [patch.txt](https://github.com/libxmp/libxmp/files/7889375/patch.txt)

Will do in a minute. The very small amount of fuzzing so far is going fine (aside from the known slow load bugs) but I forgot I deleted my stb_vorbis corpus to shrink the VM image when I updated Fedora... oh well.

sezero:
Let's get this in.

sezero:
Now pushed this to SDL_sound, too.

PR #553, opened 2022-01-18
miscellaneous stuff

sezero:
- remove now-unnecessary HAVE_ALLOCA_H defines from watcom makefiles.
- make LIBXMP_CORE_PLAYER imply LIBXMP_NO_PROWIZARD && LIBXMP_NO_DEPACKERS
  lite/jni/Android.mk would have failed in src/load.c before.
  (This was probably discussed before, but can't remember when/where.)


AliceLR:
>   (This was probably discussed before, but can't remember when/where.)

I think it existed before when I implemented those flags and then it got removed... : )

src/load.c
```diff
@@ -112,7 +112,7 @@ static int test_module(struct xmp_test_info *info, HIO_HANDLE *h)
 		if (format_loaders[i]->test(h, buf, 0) == 0) {
 			int is_prowizard = 0;
 
-#if !defined(LIBXMP_CORE_PLAYER) && !defined(LIBXMP_NO_PROWIZARD)
+#if !defined(LIBXMP_NO_PROWIZARD)
```
AliceLR, line 115: This could be changed back to an `#ifndef` for consistency with the other checks in this file. (I *think* it was `#ifndef` before, anyway...)

sezero: Done.

AliceLR (approved):
This seems OK.

Issue #477, opened 2021-08-04
cmake: documentation not generated


sezero:
As title says. We should do this (possibly similar to libmikmod, i.e.
disabled silently when the tools aren't found.)

@Wohlstand can you do this, or should I make the effort?

Wohlstand:
I'll try, however, first off I need to check how it's doing. Also, when the case tool is missing, I'll pop up the warning on the fact the important tool is absent and documentation will NOT be generated.

Wohlstand:
I Will try something in the evening (UTC+3), right now I am busy with my main job.

Wohlstand:
Okay, I'll try some tomorrow, this day sucks... Tomorrow I should have more time and energy for the work

sezero:
Any progress?

Wohlstand:
Recently I tried to implement some, and I almost finished my work, I will need to polish a bit until I will send an update, I was busy on real-life tasks and other works, thanks for reminding me, I'll try to resume this work as soon as possible.

PR #555, opened 2022-01-19
Make cmake build the documentation:

sezero:
Also update docs distfiles so that style.css goes into the tarball.
Changed the existing docs build, where we maintain two headers i.e.
manpage.rst and pdfdoc.rst which include libxmp.rst, so the tricky
sed machinery is no longer needed.

Closes: https://github.com/libxmp/libxmp/issues/477

Please test and review. CC: @Wohlstand

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -154,6 +154,10 @@ if(EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/examples/CMakeLists.txt")
 endif()
 
 
+# === Documentation ====
+add_subdirectory(docs)
+
+
 # === Install ====
 include(GNUInstallDirs)
 
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -1,0 +1,55 @@
+option(LIBXMP_DOCS "Build the library documentation" ON)
+
+set(libxmp_man_rst ${CMAKE_CURRENT_SOURCE_DIR}/manpage.rst)
+set(libxmp_html_rst ${CMAKE_CURRENT_SOURCE_DIR}/pdfdoc.rst)
+set(libxmp_man ${CMAKE_CURRENT_BINARY_DIR}/libxmp.3)
+set(libxmp_html ${CMAKE_CURRENT_BINARY_DIR}/libxmp.html)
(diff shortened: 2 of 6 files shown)


Wohlstand:
I tested this at me, and seems fine! :+1:

The code looks good. I see you had to move/modify doc files theme selves. I tried to make the work without touching them, however, I basically lost all the changes I made because of a stupid accident, and I lost interest to redo the work again, but I wanted to redo it as soon as possible. And, I really glad you did that work by yourself, thanks for the help! :+1:

sezero:
Should we add restrictions to manpage installation, e.g. not for MSVC, etc? If so, which targets?

Wohlstand:
> Should we add restrictions to manpage installation, e.g. not for MSVC, etc? If so, which targets?

I guess, yeah, basically, unnecessary for MSVC-based projects at all

sezero:
OK, maybe the following, then? (Indentation intentionally not changed to show
the actual difference):
```diff
diff --git a/docs/CMakeLists.txt b/docs/CMakeLists.txt
index cebcc60..d8d3b70 100644
--- a/docs/CMakeLists.txt
+++ b/docs/CMakeLists.txt
@@ -22,10 +22,12 @@ if(LIBXMP_DOCS)
       VERBATIM)
     add_custom_target(libxmp_gen_man ALL DEPENDS ${libxmp_man})
 
+    if(UNIX OR MINGW)
     include(GNUInstallDirs)
     install(FILES "${libxmp_man}"
             DESTINATION ${CMAKE_INSTALL_MANDIR}/man3
     )
+    endif()
   endif()
 
   if(NOT RST2HTML_EXE)
```

Wohlstand:
In theory, would be better to disabling the whole block against man file, not just installing

sezero:
> In theory, would be better to disable the whole block against the man file, not just install.

Well, they won't have the tools to build it any way. Pushed 712b954e9cf2fd5044950452632112b96c4a7459

sezero:
>  Also, what about the Cygwin environment?

Done: 0ad0ac5ae6855db49db9cfe7f15b1de5c2784dc8

sezero:
Squashed and rebased to master. Any changes needed or can we get this in?

AliceLR (approved):
The Makefile and rst changes seem fine to me.

sezero:
OK, merging.

PR #557, opened 2022-01-24
Fix reverse sustain loop release bug, add IT effect S9F support.

AliceLR:
When a sample sustain loop playing in reverse is released, the reverse state should be canceled unless the forward loop is bidirectional. This was the last issue preventing OpenMPT's SusAfterLoop.it from working in libxmp. Found while checking the unused OpenMPT IT tests for #145.

Prior to this fix, the voice would instead stop playing and generate large negative voice position values due a separate bug, which was the lack of reverse handling for non-bidirectional samples. This would eventually cause undefined behavior due to invalid float to integer conversion and potentially other much more serious bugs. The easiest way to verify I (probably?) fixed this second bug was to just implement the Modplug `S9F` Play Reverse effect ~~(but not the equivalent XM effect `X9F`, currently. If it sees significant usage I can add it to this patch)~~.

In addition to adding a test for SusAfterLoop.it, I added a separate regression test for `S9F` and its interactions with sample loops.

Comment/feedback from @sagamusix is definitely welcome.


AliceLR:
MINGW32 test failure may be x87 (yet again).

sagamusix:
Reverse play support could also be added to DBM (command E3x), DMF (though IIRC that loader is currently unused) and MED (command 20).

src/loaders/it_load.c
```diff
@@ -137,9 +137,16 @@ static void xlat_fx(int c, struct xmp_event *e, uint8 *last_fxp, int new_fx)
 			e->fxt = FX_SETPAN;
 			e->fxp = l << 4;
 			break;
-		case 0x9:	/* 0x91 = set surround */
-			e->fxt = FX_SURROUND;
-			e->fxp = l;
+		case 0x9:
+			if (l == 0 || l == 1) {
```
sagamusix, line 141: `l == 0` turns off surround in MPT (it could be imported as `X80`), in IT itself it doesn't do anything.

sagamusix:
Works well, just some small comments remaining.

AliceLR:
Thanks for looking at this patch and testing.

> `l == 0` turns off surround in MPT (it could be imported as `X80`), in IT itself it doesn't do anything.

> DBM (command E3x)

> MED (command 0x20)

Still trying to decide what a good way to fix `compare_mixer_data` for x87 is, so yeah, I might as well add XM and these to the patch (and Ultra Tracker, assuming its effect even works...).

The `l == 0` bit was effectively already there—before it was passing the low nibble directly to `FX_SURROUND`. libxmp puts the parameter for that effect into `xc->pan.surround`, which is used as a toggle rather than centering when surround is disabled. I don't know what the reasoning is there or if it was intentional...

> DMF

I don't remember but if libxmp even supports its effects, I'll just add it and worry about whether it works whenever I have to fix that loader... 😅

sagamusix:
> The l == 0 bit was effectively already there—before it was passing the low nibble directly to FX_SURROUND. libxmp puts the parameter for that effect into xc->pan.surround, which is used as a toggle rather than centering when surround is disabled. I don't know what the reasoning is there or if it was intentional...

Oh, I just now realized that it puts `l` into the parameter, not `1`. Stupid font. Nothing to fix there. :D

AliceLR:
> Oh, I just now realized that it puts `l` into the parameter, not `1`. Stupid font. Nothing to fix there. :D

Yeah, not a great variable name choice. The only reason I noticed the difference originally is syntax highlighting in VS Code, but they look very close with dark theme GitHub.

> DBM (command E3x)

>     E3x  Play sample backwards
>         x=0 "plays sample from backward" (sic)
>         x=1 "plays sample from backward, next from front loops" (sic)

🤨

> MED (command 0x20)

`20 00` should be fine locally, but I need to test before I push. I don't think the other parameter values can be trivially supported currently.

> (Ultra Tracker, assuming its effect even works...)

libxmp's effect translation is kind of garbage here and this patch really isn't the place to give it a full overhaul, skipping.

> I don't remember but if libxmp even supports [X-Tracker] effects

It doesn't, I'm guessing Claudio (or whoever) realized pretty quickly it's kind of futile with only 2 effects slots.

AliceLR:
> DBM (command E3x)

Looks like DigiBooster Pro, like MED, requires the effect to be on the same row as the note. Instead of looping in reverse, it plays in reverse once and then plays forward (seemingly regardless of the parameter). The two separate effects are DIGI Booster only and that loader also seems to need separate attention, so I'll skip it.

sezero:
Seems like devcheck is fixed for mingw as of f282fb3, which also passes on my i686-linux.

sagamusix:
Ah, seems like DBPro's reverse sample command is the same as in DigiBooster then (makes sense, doesn't it?) except for the strange interaction between software-mixed channels. Just importing it as S9F is probably still a better approximation than not supporting it at all.

sagamusix:
>    x=0 "plays sample from backward" (sic)
>    x=1 "plays sample from backward, next from front loops" (sic)

I know this is the behaviour from Digi Booster, but I cannot find it being mentioned in the Digi Booster Pro manual. Playing around a bit in the tracker, any E3x parameter causes the whole sample to play in reverse (if there's a note next to it), and if the sample has a loop, it will continue playing that loop (in the correct direction) once it hits the start of the sample.

AliceLR:
> > x=0 "plays sample from backward" (sic)
> > x=1 "plays sample from backward, next from front loops" (sic)
> 
> I know this is the behaviour from Digi Booster, but I cannot find it being mentioned in the Digi Booster Pro manual. Playing around a bit in the tracker, any E3x parameter causes the whole sample to play in reverse (if there's a note next to it), and if the sample has a loop, it will continue playing that loop (in the correct direction) once it hits the start of the sample.

I found this specifically in the DIGI Booster format documentation in [docs/formats/](https://github.com/libxmp/libxmp/blob/master/docs/formats/digi-format.txt#L94). I only brought it up because the format document libxmp has for DigiBooster Pro is [even less helpful](https://github.com/libxmp/libxmp/blob/master/docs/formats/DigiBoosterPro.txt#L182).

AliceLR:
Still not sure what to do about the other loaders with this effect. It seems like there needs to be a better way of doing the loader tests so small effects conversions can get covered independently without blowing up the number of tests...

AliceLR:
@sagamusix dumb question, but do `X9E` and `X9F` even work in OpenMPT currently? It seems like they do nothing when I make a test module that uses them, whereas Modplug Tracker plays it back roughly the same as `S9E`/`S9F`. Not sure if this is because I converted from IT or if they're just broken. might have to switch back to MPT 1.16 because of this 😎

sagamusix:
Ah, true, there is a [compatibility setting](https://wiki.openmpt.org/Manual:_Compatible_Playback#XM_compatibility_settings), "Do not allow ModPlug extensions to X command" (on by default), which disallows any X commands except for X1 and X2. For modules made with older OpenMPT / MPT the commands thus keep working, but for newly created modules they don't by default to discourage people from using these extensions.

AliceLR:
> Ah, true, there is a [compatibility setting](https://wiki.openmpt.org/Manual:_Compatible_Playback#XM_compatibility_settings), "Do not allow ModPlug extensions to X command" (on by default), which disallows any X commands except for X1 and X2. For modules made with older OpenMPT / MPT the commands thus keep working, but for newly created modules they don't by default to discourage people from using these extensions.

Oh, okay, that explains it. I didn't notice it in the list because I was looking for something unchecked... whoops. With that option on everything seems to be the same as with IT, i.e. the forward loop sample plays a backward loop, and the bidirectional loop sample starts from the end instead of the start.

> Edit: I do remember that S9F + looped samples can indeed be finicky, I will look into it tonight. But since OpenMPT essentially introduced that effect (in IT / S3M / XM, anyway), I never considered changing the behaviour.

I actually meant MPT has the unusual (buggy?) behavior of ignoring it with loops. Looks like it does this with the IT module, too!

sagamusix:
> I actually meant MPT has the unusual (buggy?) behavior of ignoring it 
> with loops. Looks like it does this with the IT module, too!

Still waking up... x) Yes, I meant that MPT introduced the buggy 
behaviour and that's why I didn't bother fixing it in OpenMPT, too. From 
what I remember it doesn't really ignore it but it's an issue that the 
loop end check is done immediately after setting the position to the end 
of the sample, so it starts playing from the start instead. I will look 
into the code again later today to confirm this.

sagamusix:
> and that's why I didn't bother fixing it in OpenMPT

Apparently I did change my mind! In OpenMPT the sample is always played from the end of the loop.

sezero (approved):
I guess @sagamusix is OK with this - so am I.

PR #559, opened 2022-01-27
Fix IT tone portamento + offset.

AliceLR:
When the IT tone portamento and offset effects are used together on a note, the new note should slide from the previous note and also have its offset changed. libxmp was ignoring the portamento in this case. Yet another bug from #145.


sezero (approved):

AliceLR:
[Found why this was implemented incorrectly](https://wiki.openmpt.org/Development:_Test_Cases/IT#porta-offset.it). The description for this test seems to be totally false: there is no apparent portamento because the test module in question does a `g9` volume column portamento from C-5 to... C-5. With a lower tone portamento rate and two *different* notes, it blatantly does do the portamento.

sagamusix:
porta-offset.it was created months before I created the wiki page, and was just a complementary test to porta-offset.xm, so some details may have gotten lost when writing it up. Thanks, I will fix the descriptions.

sagamusix:
I removed the last sentence from the Off-Porta description. I think that should fix, as the rest was pretty much correct? I will change the portamentos just in case to demonstrate that an actual portamento is happening.

AliceLR:
> porta-offset.it was created months before I created the wiki page, and was just a complementary test to porta-offset.xm, so some details may have gotten lost when writing it up. Thanks, I will fix the descriptions.

Oh, interesting. I wasn't aware XM tone portamento blocks the offset effect altogether in this situation. That makes sense how the test ended up like that, then.

> I removed the last sentence from the Off-Porta description. I think that should fix, as the rest was pretty much correct? I will change the portamentos just in case to demonstrate that an actual portamento is happening.

*If* I understand this test correctly, it's intended to check that the envelope release and fade aren't changed by a portamento except in Compatible Gxx mode, which I think the wiki accurately describes. It was just that last sentence that stuck out, I think. (Should 6 be added to the list of rows in the Off-Porta.it description?)

sagamusix:
Hm yeah, that list doesn't make much sense anymore now that the last sentence is gone, I can remove those row numbers entirely.

PR #561, opened 2022-02-01
Fix multiple IT playback bugs affecting Atomic Playboy.

### LINKED ISSUE #145: Sagamusix "atomic_playboy.it" doesn't play correctly

AliceLR:
Fixes a few flaws in libxmp's IT playback that broke Atomic Playboy and adds two new tests:

* Envelope release and sample release should NOT be tracked with the same flag for IT. I've added a separate sample release flag. For most replayers where there isn't a distinction, I kept `NOTE_RELEASE` as a macro referencing both flags so they should continue to work and allow things like Digital Symphony to use sample sustain loops.
* Sample sustain release persists through tone portamento, even with sample changes (portamento_sustain.it).
* Sample sustain release should not be affected by whatever instrument value (or lack of one) is provided (portamento_sustain.it).
* The sample default volume flag should always be set if a valid sample/instrument is provided (default_volume.it).
* Fadeout should not be restarted on portamento+offset (already tested by Off-Porta.it, but libxmp missed it).

Also updates the OpenMPT tests Off-Porta.it, Off-Porta-CompatGxx.it, and porta_offset.it due to some recent changes discussed in #559. The new version of Off-Porta.it still tests the fadeout bug mentioned above, but to get a better look at what changed in that test before the update, look at the second commit in this PR instead of the overall changes.

This probably needs testing with a variety of modules to make sure I didn't introduce any regressions or reveal additional issues that were masked by these issues. Atomic Playboy seems to play correctly now, at least. Fixes #145.

edit: Codecov is lying, the -4% to DBM was a different patch :|


sezero:
@sagamusix - please review?

sagamusix:
Hmm, I don't think that you should need a separate release flag. Both IT itself and OpenMPT only use a single release flag that is set under the same circumstances, but they also directly store loop points in their channel state, I guess the lack of such data in `struct channel_data` is the reason for having the extra flag?

AliceLR:
> Hmm, I don't think that you should need a separate release flag. Both IT itself and OpenMPT only use a single release flag that is set under the same circumstances, but they also directly store loop points in their channel state, I guess the lack of such data in `struct channel_data` is the reason for having the extra flag?

I think so, yes. In libxmp, the software mixer chooses which sample endpoints to use using sample flags and its own release flag, and it seemed like a corresponding player sample release flag was the best way to reduce complexity (and did significantly with a couple extreme edge cases e.g. carrying sample release between samples). I think there are probably other libxmp player flags that don't correspond to anything remotely resembling the original player.

sezero:
@sagamusix: verdict?

sagamusix:
If it passes the tests it's fine by me 😅

sezero:
OK then, getting this in.

sezero (approved):

knight-ryu12:
congratz for merge!

PR #562, opened 2022-02-03
Fix playercode and depacker bugs found by libFuzzer.

AliceLR:
A few of these have been sitting around for a while and I didn't get around to finishing this patch. The recent changes have resulted in new bugs though and at least one of them will be messy to fix right, so I wanted to get this in first.

* XM volume envelopes could contain junk values, eventually resulting in poorly defined behavior.
* Certain combinations of pitch effects could result in very small period values, which behave badly, particularly when converted to int.
* The stb_vorbis clamps when converting samples to 16-bit integers relied on signed overflow.
* The tinfl-based depackers didn't have allocation limits. These are nice to have for the sanitizers, and also because the Linux kernel thinks it's OK to return pointers to nonexistent RAM instead of NULL and then goes "oopsie woopsie uwu" and kills your program when you try to use them
* The gzip depacker was missing a bounds check and also didn't check the footer CRC.


AliceLR:
Coverage dropped, going to push a fix in a sec...

src/depackers/miniz.h
```diff
@@ -303,6 +304,10 @@ typedef int mz_bool;
 #define MZ_FALSE (0)
 #define MZ_TRUE (1)
 
+/* tinfl doesn't have a nicer way of limiting the output buffer size. */
```
sezero, line 307: We can move these either down after the commented-out originals or above among the libxmp-specific config section.

AliceLR: I misread and thought it was in the libxmp hacks, sorry. I moved it next to the old `MZ_REALLOC`/etc defines.

sezero (approved):

sezero:
While browsing through this, I noticed that we have an `arc_unpack` procedure and a `struct arc_unpack`, looks funny..

AliceLR:
> While browsing through this, I noticed that we have an `arc_unpack` procedure and a `struct arc_unpack`, looks funny..

Yeah, I don't know why I did that, I used `struct lzx_data` for LZX. I *think* this would just get redefined to `struct libxmp_arc_unpack` and be fine but I'll update it, thanks.

Issue #497, opened 2021-11-03
Broken OpenMPT test modules.


AliceLR:
Several test modules from the OpenMPT regression tests are not played correctly:

* `mod/PortaSmpChange.mod`: the new sample is never applied.
* `it/extreme-filter-1.it`: the output clips at the end of the envelope for the first two filter notes.
* `it/extreme-filter-2.it`: note timbre sounds weak/empty. This one currently isn't tested, but I checked it due to the prior being wrong.
* `it/s7xinsnum.it`: instrument NNA note cut cancels all three notes playing in channel 1, but it should only cut the most recent/current note.

I am sure this list is *in no way complete*, these are just the ones I noticed while fixing tests for #496.

AliceLR:
More breakage found while trying to implement IT MIDI macros:

* filter-reset.it "works" but for the wrong reasons. libxmp always resets the channel filter cutoff to 127 on a new note, which is incorrect. Instead, the cutoff and resonance are channel settings that should persist between notes. When cutoff is not 127 or resonance is not 0, the filter should always be applied according to the channel settings. When cutoff is 127 and resonance is 0, the filter values calculated off of the previous settings should be used until cutoff or resonance is changed or until a new note without portamento is encountered. *Only in this situation* will a new note disable the filter.
   
  There are other hacks littered around libxmp working around this fundamental misunderstanding. The test description for filter-reset.it is lying about the circumstances that cause the filter to turn off, which is probably why this is so broken. Fixed in this branch, but [this commit](https://github.com/libxmp/libxmp/commit/70d505ae4926e9e6d8277e04dd43e82d3ac2480a) sabotages it and had to be partially disabled: https://github.com/libxmp/libxmp/compare/master...AliceLR:fix-filter-cutoff-newnote
* gxsmp2.it relies on the above commit to play correctly, which suggests whatever is going on here is implemented incorrectly. "Fixing" anything by completely clearing the current mixer channel outside of a note cut seems very suspicious, anyway. The test descriptions for this set of modules are mostly useless so it's hard to tell what's going on here.

AliceLR:
> how portamento and NNAs are supposed to interact

It disables NNAs entirely for that note as far as I can tell.

AliceLR:
Not OpenMPT test bugs, but found while fixing the NNA/DCA portamento issue. These specifically affect an instrument with NNA continue/duplicate note check/DCA note cut with different samples on different keys assigned to the same note, I haven't tested any other cases.

[...]

PR #565, opened 2022-02-08
Fix NNA and tone portamento interaction with sample changes.

AliceLR:
When tone portamento is used on an instrument with an NNA and the old note and the new note have different samples, the tone portamento should cancel the NNA for that note even though the tone portamento isn't actually played. libxmp was playing this correctly before, but it was clearing the entire mixer voice instead of just the NNA.

I added an extra module to test this slightly more thoroughly than gxsmp2.it does, Both modules played correctly prior to this change. The old behavior is worse though: a separate fix I'm working on for IT filters needs the filter parameters to be able to persist in the mixer voice during tone portamento, and clearing the entire voice struct was creating an edge case where they did not.

Related to #497.


AliceLR:
Starting to think *just maybe* the filter shouldn't be used in tests where it has no business being used

AliceLR:
This is failing test.c because fixing this revealed another bug, which is that duplicate checks are nonsensically applied to the origin channel as well as the virtual channels they're meant for. (I've encountered the same issue with past note action effects so I'm not surprised.)

sezero (approved):

PR #568, opened 2022-02-12
Fix mixer crashes caused by bad IT sustain loop bounding.

AliceLR:
This fixes a regression caused by the bidirectional loops/sustain loop overhaul that broke `libxmp_load_sample` bounding for sustain loops. Prior, the regular loop and the sustain loop would enter separate calls of `libxmp_load_sample`. This call does some final bounding on the sample loops to prevent mixer crashes. This only gets called once now, but the sustain loop fields are in the `extra_sample_data` struct now, which is inaccessible to `libxmp_load_sample`.

The real fix would be to move these fields to the `xmp_sample` struct, but since it's part of the API that isn't really doable right now. I worked around this by adding a bounding loop in `libxmp_load_epilogue` because all of the other alternatives I trialed were crap:

* Pass a sample index to `libxmp_load_sample`. I was originally going to do this one but that function mostly restricts itself to operating on the provided `xmp_sample`, and some places (the tests mainly) rely on that not being part of the module data structures. This also requires changing >60 loaders, some of which are messy due to I *guess* a phobia of temporary variables?
* Pass `extra_sample_data` to `libxmp_load_sample`, which has similar issues to the former.
* Add `libxmp_load_sample_extra` which takes an extra parameter and redefine `libxmp_load_sample` to it—now it's hidden behind an ugly hack and you have to remember to use the correct function for every loader that needs sustain loops.
* Bounds check sustain loops manually in loaders (same problem as above).
* Derive an `extra_sample_data` pointer from the `xmp_sample` pointer with pointer arithmetic if it is in `mod->xxs` (ugly hack, and also asking for trouble).

(This entirely fixed all of the issues I was seeing with my corpus related to sustain loops, but now my corpus is generating segmentation faults for memory writes in bunzip2. I guess those are most likely all AddressSanitizer's fault since upstream doesn't use it, so we shouldn't report them!!1)


sezero (approved):

AliceLR:
> segmentation faults for memory writes in bunzip2

This is fortunately libxmp's fault, `setjmp` needs to be called earlier to prevent invalid stack unwinds.

PR #569, opened 2022-02-12
Move bunzip2 setjmp to write_bunzip_data to fix crashes.

AliceLR:
Calling `setjmp` in `read_block_header` (called from `read_bunzip_data`) can cause invalid stack unwinds if `longjmp` is used later in `read_huffman_data` (also called from `read_bunzip_data`). I moved this call to `write_bunzip_data`, which should fix this with no further issues.

This change revealed a minor warning in `read_block_header` for an uninitialized read that, as far as I can tell, should never actually occur.


sezero (approved):

sezero:
I guess I missed these (don't know how I missed):

With my old gcc-4.0, 4.2 and 4.4, I'm getting:
```
src/depackers/bunzip2.c: In function 'write_bunzip_data':
src/depackers/bunzip2.c:528: warning: argument 'len' might be clobbered by 'longjmp' or 'vfork'

```
Older gcc has an additional warning: (seems blah..)
```
src/depackers/bunzip2.c:531: warning: variable 'gotcount' might be clobbered by `longjmp' or `vfork'
```

AliceLR:
Those are both false positives, the `longjmp` shouldn't clobber anything because that function exits immediately.

sezero:
> Those are both false positives, the `longjmp` shouldn't clobber anything because that function exits immediately.

Yeah, I'll ignore them for now and maybe look for another solution later.

PR #570, opened 2022-02-15
Add missing IT filter output clamp and fix misc. sanitizer bugs.

AliceLR:
The main bugfix in this patch is the clamps added to the filter mixer loops. The lack of this was causing undefined behavior in the filter math in certain edge cases, but more importantly it was causing the filter to play much louder than Impulse Tracker in certain cases, often to the point of clipping. This issue could be found by playing one of several OpenMPT test files, but the ones I used to compare output while fixing this were env-max-filter.it, filter-reset.it, and the extreme-filter-test#.it files.

There are some other less interesting fixes that should finally get libxmp to the point where I can run the fuzzers on my corpus again:

* The Galaxy Music System 4 loader wasn't checking `hio_read` return values when reading the envelopes, which was revealed by the envelope clamps added in a different fuzz patch.
* I revised `IS_VALID_SAMPLE` to reject all samples that failed to reach or complete `libxmp_load_sample` instead of just the ones with bad lengths. Some of the samples getting through to the mixer would cause undefined behavior in the loop/position calculations (but were otherwise safe).
* The BZip2 decompressor could index the Huffman code base array past its end due to the bounds check on the index value following the indexing. The least intrusive fix was to just extend its length to match the limits array. This theoretically could be offered upstream.
* The BZip2 decompressor was expanding its output buffer linearly, which caused some extremely slow loads with certain files. I've revised it to expand by powers of 2 instead. It seems to be just slightly slower than the old one now.
* tinfl also had a slow load/OoM bug that seems to be related to underspecified trees (I haven't verified this). I reported this one upstream already: https://github.com/richgel999/miniz/issues/229


sezero (approved):
I'm OK with this, although the patch you proposed at https://github.com/richgel999/miniz/issues/229 is different from the one here?

AliceLR:
> I'm OK with this, although the patch you proposed at [richgel999/miniz#229](https://github.com/richgel999/miniz/issues/229) is different from the one here?

I just removed one of the checks (since the loop is unrolled once), and moved the remaining one to a slightly different position. It should work the same as before.

sezero:
> I just removed one of the checks (since the loop is unrolled once), and moved the remaining one to a slightly different position. It should work the same as before.

`code_len` is computed as `sym2 >> 9` twice and your original patch handles that --
I can't follow how the new one does that? (Is the second case somehow impossible to
be 0 in the second case?)

AliceLR:
> > I just removed one of the checks (since the loop is unrolled once), and moved the remaining one to a slightly different position. It should work the same as before.
> 
> `code_len` is computed as `sym2 >> 9` twice and your original patch handles that -- I can't follow how the new one does that? (Is the second case somehow impossible to be 0 in the second case?)

It just lets the second one handle the "code" of length 0 normally since it will loop back to the first one with the check. It's kind of lazy but it only needs to work until upstream fixes it... 😅

sezero:
OK then, let's get this in.

PR #571, opened 2022-02-16
Fix for IT filter cutoff 127 on new note behavior.

AliceLR:
The IT filter should be disabled if the channel filter cutoff is currently 127 and the channel filter resonance is currently 0. This combination of settings should only be applied when a new note is encountered that does not use portamento; otherwise, the previous settings should be used. libxmp was instead always resetting the cutoff on a new note. (Note: assuming no filter envelope.)

The changes to the OpenMPT test data files are mostly caused by resonance values being added to the output, plus 254 (IT filter cutoff 127*2) being used in some places for the channel cutoff value instead of 255 (the value libxmp was setting the cutoff to on a new note). The diff highlighting here is actually useful for a change!


sezero:
@sagamusix: can you have a look?

test-dev/compare_mixer_data.c
```diff
@@ -80,6 +80,9 @@ static void _compare_mixer_data(const char *mod, const char *data, int loops, in
 			}
 			fail_unless(abs(vi->pos0 - pos0) <= 1, "position mismatch");
 			if (num >= 11) {
+				if (cutoff >= 254 && vi->filter.cutoff >= 254) {
+					cutoff = vi->filter.cutoff;
```
sezero, line 84: How do we hit this?

AliceLR: Cutoff values of 254 and 255 are interchangeable due to libxmp doubling channel cutoff values. The maximum channel filter cutoff value is 127 in IT, which libxmp doubles to 254, but libxmp also uses 255 elsewhere. Most of the mixer tests contain cutoff values of 255 due to this, but this patch causes some of them to change to 254, since the cutoff isn't being reset to 255 anymore. This check is to suppress a bunch of mixer tests that broke from this minor and mostly irrelevant change (I think it was a couple dozen?).

The doubling is presumably to appease IMAGO Orpheus modules, which have a cutoff effect that *maybe* uses its full parameter range. I haven't figured out how to get the filter to work in IMAGO Orpheus with DOSBox yet so I can't reasonably verify this. This doubling behavior should probably die (and the player should not divide the cutoff value by the filter envelope range), but I can't do much about it without being able to test that format. After that gets fixed the mixer data tests can probably just normalize the new cutoff values to 255 and this check can probably be removed.

sezero: OK.

sezero (approved):
I'm OK with this if @sagamusix is too.

test-dev/compare_mixer_data.c
sagamusix: > The doubling is presumably to appease IMAGO Orpheus modules, which have a cutoff effect that maybe uses its full parameter range

I think it's more for the sake of increased precision after multiplying the filter envelope with the cutoff. True, the cutoff itself is limited to 0...127, but by using the filter envelope you could get fractional values. IT itself uses even higher precision here in fact (127*255 instead of just 127*2 possible values).

sagamusix:
Fine by me if the tests pass :)

sezero:
> Fine by me if the tests pass :)

OK, the patch is in.

PR #573, opened 2022-02-23
Add IT MIDI macro filter effects support.

AliceLR:
Impulse Tracker has a built-in MIDI macros system for sending external commands that libxmp could ignore, except that it also controls the channel filter cutoff/resonance/mode. Due to the way the macro parameters are implemented, these need to be expanded in the playercode with a text parser, so that's what this patch adds. This took a while because I also added 6 OpenMPT tests libxmp was previously ignoring/missing and also ended up simplifying parts of the parser after consulting the MIDI 1.0 spec.

Known limitations:
* Smooth MIDI macros (OpenMPT extension) probably aren't quite correct.
* No handling for `F0 F0 02 xx`, which allows enabling/disabling the high pass filter (not currently supported by libxmp).
* Several macro parameters are probably implemented wrong due to their documentation: `u` `v` `x` `y` rely on IT-specific intermediate volume and panning calculations and/or calculations from the previous tick. The specifics aren't really clear to me and I haven't extensively tested them with IT's WAV writer. I *think* `v` should be pretty close to correct, though.
* Several macro parameters are unsupported and emit 0: `a` `b` `c` `p` require somewhere to place the IT MIDI instrument fields and don't seem that useful outside of MIDI output; `s` (OpenMPT extension) is irrelevant because libxmp doesn't output MIDI.

Diff size is mostly thanks to the new regression tests...


src/effects.c
```diff
@@ -835,6 +837,25 @@ void libxmp_process_fx(struct context_data *ctx, struct channel_data *xc, int ch
...
+	case FX_MACRO:
+		SET(MIDI_MACRO);
+		xc->macro.val = fxp;
+		xc->macro.slide = 0;
+		break;
+	case FX_MACROSMOOTH:
+		if (ctx->p.speed && xc->macro.val < 0x80) {
+			SET(MIDI_MACRO);
+			xc->macro.target = fxp;
+			xc->macro.slide = round((fxp - xc->macro.val) / (double)ctx->p.speed);
```
sezero, line 852: We should make `libxmp_round` in period.c public to the rest of the source and use it here.

AliceLR: I'm not sure this `round` is actually required at all... this was one of the earlier parts I implemented a few weeks ago I and forgot why I did that. I can test further but I'm going to try to finish up a fuzz bugfix patch and the Digital Symphony patch first.

sagamusix: Since values here are never negative you could just as well go with +0.5 and truncation.

sagamusix:
Code looks good, didn't have time to test it so far though. I guess one next step could be supporting MIDI macros in MPT-made XMs (:

sezero (approved):
@sagamusix is OK with this, so I'm good too. Needs rebasing and direct use of `round` should be handled: OK with those changes.

AliceLR:
> MIDI macros in MPT-made XMs (:

Eww... I don't think libxmp even allows the filter with XMs at all, currently. How many XM modules actually use these :(

> Needs rebasing and direct use of `round` should be handled: OK with those changes.

Alright, I'll figure that out soonish. I didn't pay that close attention for licensing reasons but I think OpenMPT uses float math for macro slides, so the "fix" to get rid of `round` might just be to add extra precision (float or not).

AliceLR:
> Not _that_ many but more than zero. :D

Old folder that I haven't updated yet, but after a lazy check I'm getting 301 XMs that at the very least contain EFx, Zxx, or \xx... vs. 5 ITs that use \xx. :| Most of those are probably messages written in pattern data which don't actually use the filter.

> the "fix" to get rid of `round` might just be to add extra precision (float or not).

Single precision floats ended up being the cleanest way of addressing this without increasing the size of `channel_data` further and without introducing more x87 rounding bugs (so far).

sezero:
OK, the patch is in.

PR #581, opened 2022-03-16
Fix bugs found by libFuzzer/ASan and MSan.

AliceLR:
I've been running libFuzzer/MSan quite a bit on libxmp (hence going quiet for a while) but I haven't found a whole lot.

* Fixed a mixer crash caused by yet another assumption that all samples play forward that I failed to notice before.
* ArcFS: `is_directory` can be read uninitialized in some edge cases. This wouldn't have actually caused any bugs.
* LHA: another hang caused by reading unsigned 32-bit fields into `int`...
* MED4 and OctaMED: the synth tables could contain uninitialized junk due to missing HIO error checks.
* UMX: missing HIO error check allowed `memcmp` to compare against uninitialized junk. Another case that wouldn't cause any significant bugs.

I skipped adding tests to some of the UMRs because they're extremely minor bugs, so expect a marginal coverage drop.


src/loaders/umx_load.c
```diff
@@ -129,7 +129,8 @@ static int get_objtype (HIO_HANDLE *f, int32 ofs, int type)
 	char sig[16];
 _retry:
 	hio_seek(f, ofs, SEEK_SET);
-	hio_read(sig, 16, 1, f);
+	if (hio_read(sig, 16, 1, f) < 16)
```
sezero, line 132: This is broken: Valid umr files are rejected, because:
- the result should be `1`, not `16`
- do the lines 162 and 174 not need similar checks?
- can we work this around by zeroing the `sig` array?

AliceLR: * You're right, missed that part because I'm used to libxmp passing the size into `count` everywhere... I'll revise this.
* 142, 144, 162, 174 probably wouldn't have this issue since the first read is guaranteed to initialize the array.
* That would probably also work and is also guaranteed to initialize the array.

sezero: >     * 142, 144, 162, 174 probably wouldn't have this issue since the first read is guaranteed to initialize the array.
>     * That would probably also work and is also guaranteed to initialize the array.

You are right, initialization is what only matters - sorry.

sezero (approved):

PR #582, opened 2022-03-30
Fix issues found by libFuzzer/UBSan and other issues.

AliceLR:
UBSan found a bunch of pedantry as usual, but also a couple of out of array indexing bugs.

* stb_vorbis: the `Mapping` submap arrays could be indexed out of bounds.
* stb_vorbis: codeword lengths of 32-bits could cause signed overflows.
* stb_vorbis: hack to suppress warnings about stb_vorbis not handling the case of codebooks where `entries`*`dimensions` does not fit into an `int`.
* Junk data in envelopes could cause volume to overflow.
* Extreme edge case Schism Tracker timestamp values could overflow the date calculation.
* Formats using markers in the order list (S3M, IT, etc) could cause libxmp to attempt to index beyond the end of the order list to find an end marker.
* Repeating from the end of the module or from a module end marker into a marker at the start of the module would cause a junk global volume value to be set.
* Improved Heatseeker module data requests so some theoretical modules shouldn't be as slow anymore. I have several example fuzz files but didn't bother with a test; they're all ~100kb and compressing them results in memory loading, which isn't nearly as slow. This doesn't especially help but it's better than before...
* Another LHA negative length bug in the level 1 header handler would cause unsigned underflows but not really anything else of note.


sezero (approved):

sezero:
Great, this is in.

Issue #506, opened 2021-12-22
muse.c should be moved out of depackers


sezero:
As the title says - without that, gal4/gal5 loaders are useless.

sezero:
Related issue: https://github.com/libxmp/libxmp/issues/387

AliceLR:
Even though I proposed this I'm having a hard time seeing what the clean implementation of it is—maybe a muse_load.c and move tinfl to core code? Would it be better to have a "light" depackers mode where we keep the MOD-specific depackers (similar to how libmodplug and MikMod handle depackers) and get rid of the external/general/copyleft depackers?

sezero:
General idea sounds good to me.

> edit: I mean specifically these ones:
> 
>  MMCMP (PD? MIT???)

_Should_ be PD (all implementations around are lifted from libmodplug)

> Muse (trivial to rewrite as permissive with tinfl.c).
> PowerPacker (PD)
> StoneCracker (PD)
> XPK-SQSH (MIT)

Yes

> MAYBE ARC/ArcFS after I rewrite them since support for these seems uncommon at best, but a TON of modules from the RISC OS scene are packed in them.

Sound nice.

The ones to be left out are unzip, gunzip, bunzip, unlha, unlzx, unxz (the
ones that we actually build): My general idea of them is they should be handled
by the player -- i.e. libxmp5 maybe?

Also, maybe we can disable depackers by default, and add a big fat notice
to configury that with depackers enabled the library is (L)GPL'ed ??

AliceLR:
To clarify, I didn't mean disabling a bunch of depackers as a *default* option, but as an option to strip the library down a little bit without getting rid of required functionality like MUSE (if needed).

I don't think the LGPL licensed code will be much of an "issue" in the long term, since the worst of it has already been replaced (KZip) or is queued up for replacement (nomarch). That leaves what, bunzip2 and LHA (and LZX which is Amiga only)?

AliceLR:
Ugh, thinking about this more, it seems like there's four different states for the depackers we'd want to target in our current situation (emphasis on the relevant case for this ticket):

* All depackers, including LGPL and external. (`--enable-depackers-lgpl` or something)
* All depackers except LPGL and external. This seems like the most sane default state, particularly after more depackers get updated. (`--enable-depackers`)
[...]

PR #586, opened 2022-04-11
move MUSE container out of depackers.

sezero:
Fixes: https://github.com/libxmp/libxmp/issues/506.


AliceLR:
I didn't explain myself very well in #506—this is creating an interdependency on depacker and loader components, similar to what previously needed to be untangled with Digital Symphony. The depacker architecture is *not incorrect* here: it *should* be handling MUSE the way it currently is. The actual problem is the inclusion of archive and general compression formats in the depacker system and there's no easy, non-API breaking fix for that.

No one is forcing anyone to use `--disable-depackers`. Just leave it as-is for now IMO.

This also makes libxmp report a non-existent module format to users (MUSE is just a container).

sezero:
I'd argue that muse depends on miniz_tinfl (which I moved to src/), depackers->miniz_zip happens to depend on that too. Is it that bad as an interdependency?

> This also makes libxmp report a non-existent module format to users (MUSE is just a container).

Same with umx: I guess we need to add a container interface

sezero:
Pushed the NUM_FORMATS patch to master which doesn't have anything to do with this.

src/loaders/muse_load.c
```diff
@@ -0,0 +1,126 @@
...
+	uint32 r;
+
+	if (hio_read(in, 1, 8, f) != 8) {
+		return -1;
+	}
+	if (memcmp(in, "MUSE", 4) != 0) {
+		return -1;
+	}
+	r = readmem32b(in + 4);
+	/* MOD2J2B uses 0xdeadbabe */
```
sagamusix, line 56: there's two versions of mod2j2b: One for the new format and one for the old format. 0xDEADBEAF indicates that the content is an RIFF AM file, 0xDEADBABE indicates that it's a RIFF AMFF file.

sezero: gal4_load.c and gal5_load.c handle them correctly, no?

sagamusix: What I mean is that this comment doesn't make any sense. modj2b writes *both* magic byte variants, it just depends on which version you use. I'd either remove the comment entirely or update it to be correct.

sezero: Oh, I see - just removed the comment. Thanks!

src/loaders/muse_load.c
```diff
@@ -0,0 +1,125 @@
...
+#include "../miniz.h"
+#include "../md5.h"
+
+static void set_md5sum(unsigned char *src, unsigned char *md5, size_t sz)
+{
+	MD5_CTX ctx;
+	MD5Init(&ctx);
+	MD5Update(&ctx, src, sz);
+	MD5Final(md5, &ctx);
+}
```
AliceLR, line 33: These MD5 hacks in load.c and muse_load.c could probably get lost if there was an `hio_reopen_mem`—what do you think? The depackers could also use this to get rid of the `HIO_HANDLE **`.

sezero: > if there was an `hio_reopen_mem`

Do you have a draft, or some pseudo code?

AliceLR: I'll follow up with a general patch PR + a patch for this branch when I get a chance, sorry. I'm trying to close out #607 but I found a module that potentially complicates it quite a bit, so now I have ~100 of the things to sift through.

AliceLR: I implemented this in a branch here: https://github.com/AliceLR/libxmp/commit/b71b779b5f673656a647b4a4d877368ab8d1f689

I'm not sure it ended up being as big of an improvement for the depackers as I hoped, but it should still solve this problem. If it looks acceptable I'll go ahead and open a pull request for it.

sezero: Looks fair to me at first glance.

sezero: Rebased and re-adjusted after PR #611

AliceLR:
I'm still not a fan of this solution, but:

> I'd argue that muse depends on miniz_tinfl (which I moved to src/), depackers->miniz_zip happens to depend on that too. Is it that bad as an interdependency?

Splitting up miniz just kind of annoys me, I think. Sooner or later some cursed format that internally relies on DEFLATE is going to show up, so this is not that big of a deal.

> Same with umx: I guess we need to add a container interface

This is a fair point. The depackers are already the container interface and I feel like MUSE is better suited to them, but I don't think it matters very much. UMX wouldn't work as a depacker the same way MUSE does because the UMX loader cheats and provides a start offset to the secondary loader : ) But yes, this is a fair point.

I already reviewed most of the patch before bringing up the MD5 thing, I'll finish up soon so this can just be merged already.

AliceLR:
I'm still not a fan of this solution, but:

> I'd argue that muse depends on miniz_tinfl (which I moved to src/), depackers->miniz_zip happens to depend on that too. Is it that bad as an interdependency?

Splitting up miniz just kind of annoys me, I think. Sooner or later some cursed format that internally relies on DEFLATE is going to show up, so this is not that big of a deal.

> Same with umx: I guess we need to add a container interface

This is a fair point. The depackers are already the container interface and I feel like MUSE is better suited to them, but I don't think it matters very much. UMX wouldn't work as a depacker the same way MUSE does because the UMX loader cheats and provides a start offset to the secondary loader : ) But yes, this is a fair point.

I already reviewed most of the patch before bringing up the MD5 thing, I'll finish up soon so this can just be merged already.

src/loaders/muse_load.c
```diff
@@ -0,0 +1,111 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include "loader.h"
+#include "../miniz.h"
+#include "../md5.h"
```
AliceLR, line 25: Can this include be dropped too, or does something else need it?

AliceLR (approved):
This looks OK to me. (Sorry if this sends a duplicate message, I think Github is having problems right now.)

src/loaders/muse_load.c
sezero: Ah yes, missed that. Removed it, force-pushed.

Will merge when CI succeeds.

Issue #614, opened 2022-09-15
use of amiga/windows types in med.h and rtm_load.c


sezero:
Is there a point to using amiga / windows types in med.h and rtm_load.c?
https://github.com/libxmp/libxmp/blob/master/src/loaders/med.h#L13
https://github.com/libxmp/libxmp/blob/master/src/loaders/rtm_load.c#L26

Can't we convert to using libxmp types instead? [patch.txt](https://github.com/libxmp/libxmp/files/9579260/patch.txt)

AliceLR:
In both cases I believe this is because the format documentation also uses them. I don't really see the need in replicating them down to the type names (especially given that it's easier to mistake WORD and DWORD than it is uint16 and uint32).

sezero:
Should we patch, then?

AliceLR:
I think it would be OK to get rid of those types, yes.

PR #615, opened 2022-09-15
remove amiga/windows types use in med.h and rtm_load.c.

sezero:
Closes https://github.com/libxmp/libxmp/issues/614


sezero:
I think I was careful, but this needs at least another pair of eyes.

AliceLR (approved):
It looks like everything has been replaced correctly to me.

PR #618, opened 2022-10-29
VORBIS.C: Fix typo

raziel-:
Just a tiny typo fixed

Sorry for the noise


sezero:
This should go to stb mainstream.

raziel-:
Sorry, did i target the wrong file...or repository?

sezero:
The file is correct, but we inherit it from https://github.com/nothings/stb (with modifications, of course, but all submitted mainstream.) You should submit your patch to https://github.com/nothings/stb (against https://github.com/nothings/stb/blob/master/stb_vorbis.c) first. We may then inherit it. Besides, the typo fix is in the changelog, not even in a code comment, that's why I didn't consider its inclusion immediately here.

raziel-:
Understood, didn't know that.

Sorry for the noise

PR #625, opened 2023-02-22
Detect S3M files made with Graoumf Tracker, and use full OpenMPT version

sagamusix:
For a moment I thought about pulling the extended Schism Tracker version information into this PR as well (which is required for identifying Schism Tracker versions built after 2020-10-28), but then my brain froze looking at that code... Maybe later.

### DIFF
--- a/src/loaders/s3m_load.c
+++ b/src/loaders/s3m_load.c
@@ -381,8 +381,15 @@ static int s3m_load(struct module_data *m, HIO_HANDLE * f, const int start)
 		}
 		break;
 	case 5:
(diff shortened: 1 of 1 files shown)


sezero (approved):
@AliceLR: This is OK by me. Awaiting youd review.

AliceLR (approved):
This looks okay to me.

Regarding the Schism extended version info: please correct me if I am wrong, but I think I already added support for this in the function `libxmp_schism_tracker_string` in src/loaders/common.c?

sagamusix:
Oh, you're right, no logic needs to be rewritten, it just needs to be used for S3Ms as well. :)

PR #627, opened 2023-02-27
Add command line test utility based on the libFuzzer test function.

AliceLR:
This adds the option to build the libFuzzer test function as a standalone program with some minimal options and status output. The standalone version of the test function is essentially the same test program I've been using to test files by loading them into memory for a few years now.

Mainly this means I don't have to copy around that utility everywhere I set up libxmp, and I don't need to update its version of the fuzzing test to match the libxmp test when I change it. However, this might also be really useful for AFL++ or other fuzzers that execute a command line e.g. `./xmpchk -q -- [filename]`. (@debrouxl)

The program is located at `test-dev/xmpchk` and can be built with the test-dev target `make utilities` after libxmp has been built normally.

Examples:
```shell
# Test a file with 8 frames of playback instead of the default (to minimize regression test execution time).
./xmpchk -f8 file.mod

# Get output info for a directory tree that won't fit into argv.
find data/f/ | sort -V | ./xmpchk -

# Filter crash files out of directory.
for f in CORPUS/*; do ./xmpchk "$f" || mv "$f" ARTIFACTS; done
```

### DIFF
--- a/test-dev/libxmp_fuzz.c
+++ b/test-dev/libxmp_fuzz.c
@@ -1,18 +1,72 @@
+/* Extended Module Player
+ * Copyright (C) 2021-2023 Lachesis
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a
+ * copy of this software and associated documentation files (the "Software"),
+ * to deal in the Software without restriction, including without limitation
+ * the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the
+ * Software is furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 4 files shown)


test-dev/libxmp_fuzz.c
```diff
@@ -38,21 +81,206 @@ int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
...
+#ifdef __cplusplus
+extern "C"
+#endif
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+	xmp_context opaque = xmp_create_context();
+	FILE *f = NULL;
+
+#ifdef HAVE_FMEMOPEN
+	f = fmemopen((void *)data, size, "rb");
```
sezero, line 135: What happens if fmemopen isn't available, i.e. f == NULL, here? (Similar down below.)

AliceLR: `libxmp_test_function` guards against `NULL` from this and then it doesn't get closed. The only reason this is passed to `libxmp_test_function` in the first place is that I was considering having the command line version pass the original `FILE` directly, and then I changed my mind... I'll just move this into `libxmp_test_function`.

AliceLR: Okay, I did this, and also moved a few things around so all the includes and static variables are at the top.

sezero: OK, will merge as soon as CI completes.

sezero (approved):

PR #630, opened 2023-03-28
Implement S3M and IT mix volume.

AliceLR:
S3M and IT both contain a mix volume parameter, which ranges from 16 to 128 (S3M), 0 to 128 (IT), or 0 to 255 (OpenMPT), and typically has a default value of 48. The way this value is applied varies but a common approach is to use (mix volume)/256 to reduce the level of a -0dB sample. Since libxmp effectively quarters the level of samples by default through other means, this implementation just treats 48 as the "base" and allows higher values to apply gain for the time being.

Other changes made by this patch:
* Removed a sanity check for IT mix volume--OpenMPT can use the full range of this field.
* Revised S3M channel default panning calculation, which seems to have not been especially grounded in reality.

Prerequisite for #629.

edit: treating 48 as a base also means numerous tests that already rely on this output level don't need to be changed :) The only test that relies on PCM data from an S3M or IT with a non-48 mix volume is `test_api_play_buffer`, which I had to regenerate the data file for.


sezero (approved):

src/loaders/s3m_load.c
```diff
@@ -284,14 +284,15 @@ static int s3m_load(struct module_data *m, HIO_HANDLE * f, const int start)
...
 			continue;
 
 		mod->chn = i + 1;
 
-		if (sfh.mv & 0x80) {	/* stereo */
-			int x = sfh.chset[i] & S3M_CH_PAN;
-			mod->xxc[i].pan = (x & 0x0f) < 8 ? 0x30 : 0xc0;
+		x = sfh.chset[i] & S3M_CH_NUMBER;
+		if ((sfh.mv & 0x80) && x < S3M_CH_ADLIB) {	/* stereo */
+			mod->xxc[i].pan = x < S3M_CH_RIGHT ? 0x30 : 0xc0;
```
AliceLR, line 295: Probably the part I'm least sure about—it seems correct for ST3 and OpenMPT, but MPT 1.16 will foolishly emit AdLib channels for channels past 16. It also emits default panning, so I don't think this affects much...

sagamusix:
I don't want the rabbit hole to go even deeper 🙈 but for S3Ms saved with GUS in ST3, the master volume should be assumed to be 48 (see #357). This is mostly relevant to Adlib+PCM S3Ms of course, but also for PCM-only it can make sense.

src/loaders/s3m_load.c
```diff
@@ -622,6 +620,16 @@ static int s3m_load(struct module_data *m, HIO_HANDLE * f, const int start)
 
 	m->quirk |= QUIRKS_ST3 | QUIRK_ARPMEM;
 	m->read_event_type = READ_EVENT_ST3;
+	m->mvolbase = 48;
+	m->mvol = sfh.mv & 0x7f;
```
sagamusix, line 624: Note that in earlier ST3 versions, master volume used a range from 0 to 7 instead, with 0x10 being the stereo flag. Later ST3 versions try to convert this to the newer interpretation, even when they shouldn't... Feel free to take a peek at OpenMPT's S3M loader for the gory details.

sagamusix:
Other formats using master volume (not all are supported by libxmp): X-Tracker DMF, DSIK DSM, Imago Orpheus IMF, Disorder Tracker PLM, Epic MegaGames MASI PSM16, OctaMED MED, MadTracker 2 MT2, ProTracker 3.6 wrapper for MOD files (of all things!).

src/loaders/s3m_load.c
sagamusix: Yes, MPT always writes a pan table, so in practice this doesn't matter.

AliceLR:
I will patch the old S3M issue soon. I think the rest can wait for their own follow-up or until their respective issues/formats get addressed.

> Imago Orpheus IMF, Epic MegaGames MASI PSM16, OctaMED MED, ProTracker 3.6 wrapper for MOD files (of all things!).

Will get to these ASAP after this is merged.

sezero:
Is this ready for merging?

src/loaders/s3m_load.c
AliceLR: Pushed what *ought* to be a fix for this.

sagamusix (approved):

AliceLR:
> Is this ready for merging?

I think so.

Also, it looks like the Ubuntu 18.04 runners are dead now, will update that soonish too.

sezero:
OK, this is in, then.

AliceLR:
There's an incoming sequel to this specifically for MPT <=1.17rc2 modules whenever I can get it cleared by @sagamusix 😰

PR #632, opened 2023-03-28
avoid calling libxmp_note_to_period() twice from within a CLAMP() macro

sezero:


AliceLR (approved):
This looks OK, nice catch!

PR #635, opened 2023-03-29
Add compatibility for Modplug Tracker preamp scaling.

AliceLR:
Modplug Tracker and early versions of OpenMPT have a pre-amp routine that, among other things, attenuates the mix volume based on the current module's channel count. This channel scaling needs to be implemented in libxmp to fix the output levels of some especially loud MPT modules.

Modplug Tracker will scale the pre-amp level "logarithmically" from 100% (<=5 channels) to 50% (30+ channels) at the default setting 128. (This changes for higher settings, but not lower settings.)

Currently this is implemented in the loaders instead of the mixer, and only the default setting of 128 is supported. ITs should work fine now, but XMs need better MPT detection to be added.

—

This fixes probably quite a few loud modules from MPT 1.16, including [fb-pandora.it](https://modarchive.org/index.php?request=view_by_moduleid&query=41557), which I have mentioned in other issues as having volume problems. It has 62 channels and 48 mix volume, and the pre-amp scaling reduces its level by 50% due to its channel count.

I didn't try to bother with S3Ms here, as they don't have MPT detection at all (just OpenMPT).


src/loaders/common.c
```diff
@@ -535,6 +535,46 @@ void libxmp_schism_tracker_string(char *buf, size_t size, int s_ver, int l_ver)
 		snprintf(buf, size, "Schism Tracker 0.%x", s_ver);
 	}
 }
+
+/* Old MPT versions (MPT <=1.16, OpenMPT 1.17rc1 and rc2) rely on a pre-amp
```
sagamusix, line 539: If you would like to have a historically more accurate comment here: For OpenMPT, this applies to any modules using "classic" (MPT), RC1 or RC2 mixmode, which can be accessed also in later versions when opening a module made with OpenMPT 1.17 or MPT that is using those legacy mix modes. Once a module uses mixmode 1.17 RC3 and "Compatible" there's no way to go back to those older mix modes and the dynamic pre-amp is not used. The number of modules created with later versions (1.18+) and using the older mix modes is hopefully close to 0.

AliceLR: OK, I will revise this comment, sorry. I think I got that from a comment somewhere in OpenMPT and didn't check further into it.

sagamusix: It's very much possible that some older comments aren't too accurate on the topic, if you can find it again I will double-check if it needs to be updated.

sagamusix: (it really doesn't help that mixmodes 1.17 RC1 and RC2 are named after the specific versions in which they were released, but there are several other versions inbetween that could also have used it, and 1.17 RC3 was not really a version that existed as such...)

AliceLR: > It's very much possible that some older comments aren't too accurate on the topic, if you can find it again I will double-check if it needs to be updated.

The best I can figure out looking through OpenMPT is that I assumed all three were versions from this source file: https://github.com/OpenMPT/openmpt/blob/master/soundlib/SoundFilePlayConfig.cpp

It doesn't really claim it's a version, but it clearly can be misread that way (if you're me...).

sagamusix: I think the only missing bit of information now is that which mix mode is actually used is in encoded in the extended song properties saved in OpenMPT-made files, but it is safe enough to approximate this behaviour by assuming that one of those mixmodes is used when the 8.88 file version (or other file versions identifying early 1.17 versions, such as cwtv/cmwt == 0x0300) is present. In theory that module could still use the RC3 mix mode (which is a bit louder).

sezero (approved):
OK by me when you and @sagamusix finalize

src/loaders/common.c
AliceLR: OK, tweaked it to mention the extended properties, and pushed.

sagamusix (approved):

PR #636, opened 2023-03-31
Make MMD0 pattern loader honor the play transpose setting

sagamusix:
CC @AliceLR


sagamusix:
Eh... there's a testcase depending on the broken behaviour 😩

sagamusix:
That looks better.

AliceLR (approved):
OK, merging.

PR #637, opened 2023-03-31
XMF: tighten sample header validation

sagamusix:
CC @AliceLR

These are the additional sanity checks I use in OpenMPT. I guess the middle-C frequency check may be a bit questionable as it is not *technically* wrong, but given the low number of valid modules that actually exist, it can help catching some sparse binary files disguising as XMFs.


AliceLR (approved):
This looks OK to me.

PR #639, opened 2023-04-03
Fix typo in CMake config

dbarkar:
CMAKE_CURRENT_LIST_DIR is the proper name, reference: https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_DIR.html


sezero:
CC: @madebr

sezero (approved):

sezero:
Thanks.

Issue #640, opened 2023-04-03
Static builds are not working with MSVC


dbarkar:
According to this code from `xmp.h` 
```
#if defined(_WIN32) && !defined(__CYGWIN__)
# if defined(BUILDING_STATIC)
#  define LIBXMP_EXPORT
# elif defined(BUILDING_DLL)
#  define LIBXMP_EXPORT __declspec(dllexport)
# else
#  define LIBXMP_EXPORT __declspec(dllimport)
# endif
...
#endif
```
It's required for an end user to define `BUILDING_STATIC` before the inclusion of `xmp.h`.
Is it expected behavior?

It's a good practice to export such defines with CMake targets. I see that you add this define to the target `XMP_IF` but it isn't exported. 
I suggest:
1) Export `BUILDING_STATIC` with `xmp_static` target
2) Export `BUILDING_DLL` with `xmp_shared` target
3) Use more specific name to avoid name collisions, like `LIBXMP_STATIC` or so.

I'm happy to help with the issue if we are on the same page.

sezero:
> It's required for an end user to define `BUILDING_STATIC` before the inclusion of `xmp.h`. Is it expected behavior?

Yes

> It's a good practice to export such defines with CMake targets. I see that you add this define to the target `XMP_IF` but it isn't exported. I suggest:
> 
>     1. Export `BUILDING_STATIC` with `xmp_static` target
> 
>     2. Export `BUILDING_DLL` with `xmp_shared` target
> 
>     3. Use more specific name to avoid name collisions, like `LIBXMP_STATIC` or so.
> 
> I'm happy to help with the issue if we are on the same page.

Patches are welcome. CC: @madebr

madebr:
I think `BUILDING_DLL` is an artifact from autotools.
I don't know how to make `libtool` pass different macro's to shared and static libraries.

For CMake, adding these macro's with `INTERFACE` visibility would make the choice between shared/static libxmp library more transparent.

sezero:
> I think `BUILDING_DLL` is an artifact from autotools.

It is: to trigger the dllexport in xmp.h

> I don't know how to make `libtool` pass different macro's to shared and static libraries.

libxmp doesn't use libtool at all

PR #641, opened 2023-04-03
Export DLIBXMP_STATIC (former BUILDING_STATIC) with CMake's target

dbarkar:
Fixes https://github.com/libxmp/libxmp/issues/640


madebr:
> OK by me. I wonder though, should we bump from 4.5.1 to 4.6.0 for this because the change is public in xmp.h?

I think so. The abi remains the same, but the api changed in a way that can cause breaking builds.

@dbarkar 
Perhaps we can add `-DLIBXP_STATIC` in the pc's too?
https://github.com/libxmp/libxmp/blob/785a0cbd49f1a9415cb40871de15e97464e66ae1/libxmp.pc.in#L13
and
https://github.com/libxmp/libxmp/blob/785a0cbd49f1a9415cb40871de15e97464e66ae1/lite/libxmp-lite.pc.in#L13

sezero:
> @dbarkar Perhaps we can add `-DLIBXP_STATIC` in the pc's too?

Sounds like a good idea to me.

dbarkar:
I don't know much about pkg-config, can you help me with that?

So, CMake can generate both static and dynamic libraries at once, but there is only one .pc file. 
Is it possible to add cflags for static build only?

sezero:
> I don't know much about pkg-config, can you help me with that?

I said that it's a good idea, but `-DLIBXMP_STATIC` needs adding to
`Cflags.private` field which is a [pkgconf](https://github.com/pkgconf/pkgconf) extension not supported by
mainstream pkg-config. So I guess it's not worth the trouble?

sezero:
Noticed that we also have a problem in non-windows static builds where libxmp symbols are actually exported. We need something like the following, possibly after this is finalized (not sure emscripten would be correct after such a change.)
```diff
diff --git a/include/xmp.h b/include/xmp.h
index 686d324..47f8b45 100644
--- a/include/xmp.h
+++ b/include/xmp.h
@@ -15,15 +15,15 @@ extern "C" {
 #define XMP_VER_MINOR 5
 #define XMP_VER_RELEASE 1
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
-# if defined(LIBXMP_STATIC)
-#  define LIBXMP_EXPORT
-# elif defined(BUILDING_DLL)
+#if defined(LIBXMP_STATIC)
+# define LIBXMP_EXPORT
+#elif defined(_WIN32) && !defined(__CYGWIN__)
+# if defined(BUILDING_DLL)
 #  define LIBXMP_EXPORT __declspec(dllexport)
 # else
 #  define LIBXMP_EXPORT __declspec(dllimport)
 # endif
-#elif defined(__OS2__) && defined(__WATCOMC__) && defined(__SW_BD)
+#elif defined(__OS2__) && defined(__WATCOMC__)
 #  define LIBXMP_EXPORT __declspec(dllexport)
 #elif (defined(__GNUC__) || defined(__clang__) || defined(__HP_cc)) && defined(XMP_SYM_VISIBILITY)
 # define LIBXMP_EXPORT __attribute__((visibility ("default")))
```

dbarkar:
Fixed.
There is probably a more beautiful way, but the code looks straightforward and it's working on all platforms.

include/xmp.h
```diff
@@ -15,20 +15,26 @@ extern "C" {
 #define XMP_VER_MINOR 5
 #define XMP_VER_RELEASE 1
 
-#if defined(_WIN32) && !defined(__CYGWIN__)
+#if (defined(_WIN32) && !defined(__CYGWIN__)) || (defined(__OS2__) && defined(__WATCOMC__) && defined(__SW_BD))
```
sezero, line 18: Please don't merge os/2 watcom case in here. It is wrong because dllimport isn't supported there and results in link failures.

dbarkar: Oops, sorry about that. Fixed.

sezero:
@AliceLR, @madebr: Comments?

AliceLR (approved):
This looks OK to me.

> I wonder though, should we bump from 4.5.1 to 4.6.0 for this because the change is public in xmp.h?

OK, but I'm not adding anything else to my TODO for the next release except minimal API changes (stereo sample flag...).

sezero:
Pushed 1b4338284b09ae028059d50934952267a9b8e86d before merging to tweak the watcom + os/2 case.

Question: Does anyone know why we are excluding Cygwin from dllexport/dllimport stuff?

dbarkar:
It's coming from this commit: https://github.com/libxmp/libxmp/commit/aa60a45a675ae99a4a74cb2522dad6b65ddf936d

sezero:
> It's coming from this commit: [aa60a45](https://github.com/libxmp/libxmp/commit/aa60a45a675ae99a4a74cb2522dad6b65ddf936d)

OK then.

Issue #642, opened 2023-04-12
UWP support


dbarkar:
Currently, the library can't be built for UWP:
```
FAILED: CMakeFiles/xmp_shared.dir/src/depackers/ptpopen.c.obj 
C:\PROGRA~1\MICROS~3\2022\PROFES~1\VC\Tools\MSVC\1434~1.319\bin\Hostx64\x64\cl.exe   -DBUILDING_DLL -DHAVE_POWF=1 -D_CRT_SECURE_NO_WARNINGS -D_USE_MATH_DEFINES -Dxmp_shared_EXPORTS -IC:\workspace\vcpkg\buildtrees\libxmp\src\7464e66ae1-c4ed86357c.clean\include /DWIN32 /D_WINDOWS /D_UNICODE /DUNICODE /DWINAPI_FAMILY=WINAPI_FAMILY_APP /D__WRL_NO_DEFAULT_LIB__ /nologo /Z7 /MP /GS /Gd /Gm- /W4 /WX- /Zc:wchar_t /Zc:inline /Zc:forScope /fp:precise /Oy- /EHsc  /utf-8  /wd4244 /wd4018 /wd4996 /wd4048 /wd4267 /wd4127 /D_DEBUG /MDd /Od /RTC1  -DDEBUG -D_DEBUG /showIncludes /FoCMakeFiles\xmp_shared.dir\src\depackers\ptpopen.c.obj /FdCMakeFiles\xmp_shared.dir\ /FS -c C:\workspace\vcpkg\buildtrees\libxmp\src\7464e66ae1-c4ed86357c.clean\src\depackers\ptpopen.c
C:\workspace\vcpkg\buildtrees\libxmp\src\7464e66ae1-c4ed86357c.clean\src\depackers\ptpopen.c(146): error C2065: 'STARTF_USESTDHANDLES': undeclared identifier
ninja: build stopped: subcommand failed.
```

I guess external helpers should be disabled for UWP.

PR #643, opened 2023-04-12
Fix compilation for UWP platform, disable external helpers

dbarkar:
Fixes https://github.com/libxmp/libxmp/issues/642
This fix will work for all build systems.
CMake sets `-DWINAPI_FAMILY=WINAPI_FAMILY_APP` when building for UWP so there is a simpler fix but only if you don't care about other build systems.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -60,6 +60,21 @@
 #define LIBXMP_AMIGA	1
 #endif
 
+#if defined(_MSC_VER) && defined(__has_include)
+#if __has_include(<winapifamily.h>)
+#define HAVE_WINAPIFAMILY_H 1
+#else
+#define HAVE_WINAPIFAMILY_H 0
+#endif
+#endif
+
+#if HAVE_WINAPIFAMILY_H
+#include <winapifamily.h>
+#define LIBXMP_UWP (!WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP))
+#else
+#define LIBXMP_UWP 0
+#endif
+
 #ifdef HAVE_EXTERNAL_VISIBILITY
 #define LIBXMP_EXPORT_VERSIONED __attribute__((visibility("default"),externally_visible))
 #else
--- a/src/depackers/depacker.c
+++ b/src/depackers/depacker.c
@@ -28,7 +28,7 @@
 #include "../tempfile.h"
 #include "xfnmatch.h"
 
-#ifdef _WIN32
+#if defined(_WIN32 ) && !LIBXMP_UWP
 /* Note: The _popen function returns an invalid file opaque, if
  * used in a Windows program, that will cause the program to hang
  * indefinitely. _popen works properly in a Console application.
--- a/src/depackers/ptpopen.c
+++ b/src/depackers/ptpopen.c
@@ -12,7 +12,9 @@
  * useful. -- Kurt Keller, Aug 2013
  */
 
-#ifdef _WIN32
+#include "../common.h"
+
+#if defined(_WIN32 ) && !LIBXMP_UWP
 
 #include "ptpopen.h"
 
--- a/src/mkstemp.c
+++ b/src/mkstemp.c
@@ -44,18 +44,21 @@
 #include <fcntl.h>
 #include <errno.h>
 
-#ifdef _MSC_VER
-#include <process.h>
-#define getpid _getpid
-#define open _open
+#ifdef _WIN32
+#ifndef WIN32_LEAN_AND_MEAN
+#define WIN32_LEAN_AND_MEAN
 #endif
-
+#include <windows.h>
+#endif
 #if defined(_MSC_VER) || defined(__WATCOMC__)
 #include <io.h>
 #else
 #include <unistd.h>
 #endif
-
+#ifdef _MSC_VER
+#include <process.h>
+#define open _open
+#endif
 #ifndef O_BINARY
 #define O_BINARY 0
 #endif
@@ -64,12 +67,11 @@ int mkstemp(char *pattern)
 {
 	int start, i;
 #ifdef _WIN32
-	int   val;
+	int val = GetCurrentProcessId();
 #else
-	pid_t val;
+	pid_t val = getpid();
 #endif
 
-	val = getpid();
 	start = strlen(pattern) - 1;
 
 	while (pattern[start] == 'X') {


sezero (approved):
Looks good to me at first glance. @AliceLR ?

sezero:
@madebr: Maybe we can add a github CI runner for uwp?

sezero:
> @madebr: Maybe we can add a github CI runner for uwp?

I naively tried the following (adapted from SDL, and also with
Debug replaced with Release),
```diff
diff --git a/.github/workflows/regression.yaml b/.github/workflows/regression.yaml
index 2fd5bb1..f01501f 100644
--- a/.github/workflows/regression.yaml
+++ b/.github/workflows/regression.yaml
@@ -230,6 +230,27 @@ jobs:
           cd test-dev
           nmake -f Makefile.vc
 
+  windows-uwp:
+    runs-on: windows-latest
+    strategy:
+      matrix:
+        include: [
+          {targetplatform: "amd64", cmakeflags: "-A x64" },
+        ]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: CMake configure
+        run: |
+          cmake . -DCMAKE_SYSTEM_NAME=WindowsStore -DCMAKE_SYSTEM_VERSION="10.0" -DCMAKE_BUILD_TYPE=Debug -DWITH_UNIT_TESTS=ON -B build ${{ matrix.cmakeflags }}
+      - name: CMake build
+        run: |
+          cmake --build build --config Debug --target ALL_BUILD
+      - name: CMake checks
+        run: |
+          cd build
+          ctest -C Debug --output-on-failure
+
   windows-msys2:
     runs-on: windows-latest
     strategy:
```

Got the following warnings from the CI runner:
```
D:\a\libxmp\libxmp\src\mkstemp.c(72,8): warning C4013: '_getpid' undefined; assuming extern returning int
D:\a\libxmp\libxmp\test-dev\main.c(191,7): warning C4013: '_spawnl' undefined; assuming extern returning int
```

And the tests are failing:
```
 Test project D:/a/libxmp/libxmp/build
    Start 1: libxmp-test
1/2 Test #1: libxmp-test ......................Exit code 0xc0000135
***Exception:   0.01 sec

    Start 2: libxmp-tests-dev
2/2 Test #2: libxmp-tests-dev .................Exit code 0xc0000135
***Exception:   0.01 sec
```

madebr:
That's how I would have done it. I don't have experience with UWP, so don't know what options are possible.

I think `getpid` should be replaced by `GetCurrentProcessId`, and `spawnl` by `CreateProcess`.

madebr:
I'm not 100% sure, but I think it's failing because the test executable can't find the libxmp dll.
In SDL, we set the `PATH` environment to that of the SDL DLL.
For cygwin, that does not work because these need `cygwin1.dll`, so in my SDL cygwin PR, I change the working directory instead.
Perhaps, we need to prepend to the PATH environment variable: `"$<TARGET_DIR:libxmp>;$ENV{PATH}"`.

sezero:
> I'm not 100% sure, but I think it's failing because the test executable can't find the libxmp dll.

If that's the case, how does ctest of the `windows-vc` run succeed?

madebr:
What if you build with release configuration?
Perhaps it cannot find the debug runtime with uwp?

sezero:
> What if you build with release configuration? Perhaps it cannot find the debug runtime with uwp?

No difference in my test branch: https://github.com/sezero/libxmp/tree/uwp2
https://github.com/sezero/libxmp/actions/runs/4721414602/jobs/8374658632

madebr:
I installed the uwp SDK on my Windows system, and I can reproduce the error.
A dialog appears with:
```
The code execution cannot proceed because VCRUNTIME140_APP.dll was not found. Reinstalling the program may fix this problem.
```
[The suggestion in this StackOverflow question did not help](https://superuser.com/a/1547342)

However, when I switch from the `Visual Studio 16 2019` generator to `Ninja`, the test executables start without the error above. There are other errors however.

[Output of `ctest -V`](https://github.com/libxmp/libxmp/files/11255368/output.log)
I've got no clue why the 2nd test failed, but shows no error.

sezero:
Thanks, that's good progress

> However, when I switch from the `Visual Studio 16 2019` generator to `Ninja`, the test executables start without the error above.

Maybe the ninja dependencies include that dll ?

> [Output of `ctest -V`](https://github.com/libxmp/libxmp/files/11255368/output.log) I've got no clue why the 2nd test failed, but shows no error.

I have no clue about that either -- that it'd be because of unsupported `spawnl` use is unlikely?

What is the op's own experience / ideas anyway: @dbarkar ?

dbarkar:
I'm still figuring it out...
Did it build DLL? Locally I have to link with `runtimeobject.lib`, strange that it's building on CI successfully. 
Is it possible to run tests without fork/spawn? It's not easy to spawn UWP process, not sure if it's possible for this case.

sezero:
> I'm still figuring it out... Did it build DLL?

It did, according to the log linked above:
```
      Creating library D:/a/libxmp/libxmp/build/Release/libxmp.lib and object D:/a/libxmp/libxmp/build/Release/libxmp.exp
  xmp_shared.vcxproj -> D:\a\libxmp\libxmp\build\Release\libxmp.dll
```

> Locally I have to link with `runtimeobject.lib`, strange that it's building on CI successfully. Is it possible to run tests without fork/spawn? It's not easy to spawn UWP process, not sure if it's possible for this case.

Don't know much about uwp myself, therefore I can't help with it much.

AliceLR:
> Is it possible to run tests without fork/spawn?

If it was limited explicitly to UWP builds, maybe, but it would take some work. The runner relies on the tests being able to terminate separate from the parent process. Changing this would make getting useful regression info slower and more difficult in some cases (crashes, sanitizers, etc).

Re: replacing `_spawnl` with `CreateProcess`, there would probably be no issue here. I switched it to `_spawnl` because the former forking method for Windows (`system`) was horribly slow. If you can't get `_spawnl` to work, selectively reenabling `system` *only* for UWP builds is an option. I'd really rather not have to go back to MSYS2 debugging with `system`, it's already slow enough :)

Worst case, the regression tests are probably skippable if the regular check runs. Our CI already has both MSYS2 and MSVC regression test runs.

madebr:
I installed some extra redist packages for my VS2019 installation, and now it starts building `.msix` packages:
```
  xmp_shared.vcxproj -> C:\Users\maarten\source\repos\libxmp\build\Release\libxmp.dll

  libxmp-test.vcxproj -> C:\Users\maarten\source\repos\libxmp\build\Release\libxmp-test.exe
  libxmp-test -> C:\Users\maarten\source\repos\libxmp\build\test\AppPackages\libxmp-test\libxmp-test_1.0.0.0_x64_Test\libxmp-test_1.0.0.0_x64.msix
  C:/Users/maarten/source/repos/libxmp/build/Release/libxmp.dll
  libxmp-tests-dev.vcxproj -> C:\Users\maarten\source\repos\libxmp\build\Release\libxmp-tests-dev.exe
  libxmp-tests-dev -> C:\Users\maarten\source\repos\libxmp\build\test-dev\AppPackages\libxmp-tests-dev\libxmp-tests-dev_1.0.0.0_x64_Test\libxmp-tests-dev_1.0.0.0_x64.msix
  xmp_static.vcxproj -> C:\Users\maarten\source\repos\libxmp\build\Release\libxmp-static.lib
```
Executing the `.msix` failed with the following message:
cannot-open-libxp-certificate-issue

When I tried to open the project in Visual Studio (using `cmake --open .`), I was greeted with a request to install yet another runtime, even though I enabled all UWP related packages in the VS2019 installer.

msvc_dl_extra_things

Next hurdle I needed to take was my Windows 10 system not having enabled Developer mode. Seriously.

image

And now, I can start debugging the application.

`libxmp-test.exe` fails because it cannot load `test.itz`.
I think the permissions of UWP apps need to be configured (see [msdn](https://learn.microsoft.com/en-us/windows/uwp/files/file-access-permissions)).
I hard coded an absolute path to the itz file and `GetFileAttributesA` keeps returning -1 (=> file does not exist).

sezero:
> `libxmp-test.exe` fails because it cannot load `test.itz`. I think the permissions of UWP apps need to be configured (see [msdn](https://learn.microsoft.com/en-us/windows/uwp/files/file-access-permissions)). I hard coded an absolute path to the itz file and `GetFileAttributesA` keeps returning -1 (=> file does not exist).

Is this due to difference in compiler or something? Log from your previous attempt showed success.

madebr:
The dll + exe build fine.
SDL is in the same situation as libxmp: the CMake script builds a fine default UWP executable with minimal capabilities.

The issue is me not knowing how to build multiple UWP executables with a custom manifest (see https://github.com/libsdl-org/SDL/pull/7633)

sezero:
OK, pushed the getpid patch to this branch.

Should we merge this patch as is for now and create a new ticket for further uwp issues?

madebr:
That sounds fine.
The issue is the test scscutables don't have sufficient rights or it doesn't add enough files to the package.

sezero:
OK, this is in. Someone with enough knowledge should create a ticket for remaining uwp issues.

PR #645, opened 2023-05-03
Add tracker detection for ModPlug Tracker XMs, pre-alpha ITs.

AliceLR:
Adds a small handful of MPT module detection improvements:

* ModPlug Tracker XMs are now detected in many (but not all) cases. This is currently used only to apply MPT preamp to XMs that expect it. Example: `test-dev/data/m/xyce-dans_la_rue.xm`.
* ModPlug Tracker XM comments are now loaded.
* IT modules *probably* saved using early ModPlug Tracker pre-alpha releases are now detected. These have tracker version `0202h`, format version `0200h`, an a non-zero first pattern parapointer that is less than the first sample parapointer. There are only a very small number of these on Modland, and this fingerprinting was only recently discovered by Saga Musix.

### DIFF
--- a/src/loaders/xm_load.c
+++ b/src/loaders/xm_load.c
@@ -773,7 +773,10 @@ static int xm_load(struct module_data *m, HIO_HANDLE * f, const int start)
 	int i, j;
 	struct xm_file_header xfh;
 	char tracker_name[21];
+#ifndef LIBXMP_CORE_PLAYER
+	int claims_ft2 = 0;
 	int is_mpt_116 = 0;
+#endif
 	int len;
 	uint8 buf[80];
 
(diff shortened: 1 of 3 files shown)


sezero (approved):

src/loaders/xm_load.c
```diff
@@ -944,6 +945,73 @@ static int xm_load(struct module_data *m, HIO_HANDLE * f, const int start)
...
+				known = 1;
+
+			if (sz) hio_seek(f, sz, SEEK_CUR);
+			break;
+		}
+
+		if(known && claims_ft2)
+			is_mpt_116 = 1;
+
+		if (ext == MAGIC4('X','T','P','M'))
```
sagamusix, line 1001: Note that it would be good if this also exited the loop when encountering `STPM` in addition to `XTPM`... right now if no instrument extensions are used, OpenMPT still writes an `XTPM` tag with `STPM` following immediately afterwards (because I was expecting this kind of code which relies on both being always present), but I might want to change this in the future.

AliceLR: OK, noted!

PR #649, opened 2023-05-05
Fix misc. bugs in the MASI 16 loader.

### LINKED ISSUE #647: Issues in MASI16 loader
Just from a quick skim, there are several issues with the MASI16 loader:

- Channel panning is not imported
- Notes play at the wrong pitch
- No effect translation whatsoever, so any effects encountered in pattern data will do a completely different thing compared to what was intended (as a consequence there are lots of random pattern breaks)
- Probably more issues

As a consequence, most of these files play completely incorrectly: https://www.exotica.org.uk/mediawiki/index.php?title=Special%3AModland&md=search&mod=old&aut=&for=epic+megagames+masi&col=&siz=&ins=

In its current state, it would make more sense to disable this loader temporarily until those problems are addressed.

AliceLR:
This fixes several bugs in the MASI 16 loader. This probably should *not* close [issue link removed because Github is dumb] right away as there's no test for the effects, probably still a few issues. This patch is mainly so @sagamusix doesn't try to disable this loader :)

* Fixed sample mapping bugs affecting Epic Pinball's Enigma and shareware info tunes.
* Loaded note values were being incorrectly transposed up by 1.
* The upper 4 bits of sample finetune are now treated as transpose instead of as buggy finetune values.
* The initial panning table is now loaded.
* Initial implementations for all documented MASI 16 effects.

Issue the above text is referring to: #647 

edit: codecov is lying again. Diff coverage is very low and coverage of masi16_load.c drops to 42.7% because no effects test. Everything new that matters still gets covered, though.


src/loaders/masi16_load.c
```diff
@@ -46,7 +46,154 @@ static int masi16_test(HIO_HANDLE *f, char *t, const int start)
...
+	case 30:				/* Tremolo */
+		event->fxt = FX_TREMOLO;
+		event->fxp = param;
+		break;
+	case 31:				/* Tremolo waveform */
+		event->fxt = FX_EXTENDED;
+		event->fxp = (EX_TREMOLO_WF << 4) | (param & 0x0f);
+		break;
+
+	case 40:				/* Sample Offset */
```
sagamusix, line 136: It's not used in any "official" PSMs but this effect is just as cursed as in the newer PSM format: The parameter is three bytes long, and the middle byte is the one that is equivalent to MOD/S3M/etc. offset. You can verify this by taking e.g. `song0.psm` from Epic Pinball, replace the byte at offset `0xED` (original value: `0x3C`, set speed) with `0x28` and you will notice that the menu music will skip a beat and then hang at the end of the pattern in-game, just like it will when throwing the modified file at OpenMPT.

AliceLR: oh no :(

sagamusix:
This looks fine now as far as the original issue is concerned; I think it can safely close the issue (and I'll open a new one if anything else turns up later).

sagamusix (approved):

PR #650, opened 2023-05-06
Rename XMF loader to 'Astroidea XMF', add support for 0x04 variant.

AliceLR:
Through the magic of asking nicely, @sagamusix was able to acquire playercode and some more modules in the XMF format. These extra modules use a `0x04` for the first byte instead of `0x03`, and are mixed at normal volume levels (rather than for Imperium Galactica's extremely loud Sound Blaster driver).

A test `0x04` variant is available in [this demo](https://www.pouet.net/prod.php?which=2234) (filename PROSTATE.MUS).

edit: codecov is almost right. Diff coverage is 56%, file loses 1.16% coverage, mainly because of the new effect.

### DIFF
--- a/src/loaders/xmf_load.c
+++ b/src/loaders/xmf_load.c
@@ -20,7 +20,7 @@
  * THE SOFTWARE.
  */
 
-/* Imperium Galactica XMF loader.
+/* Loader for Astroidea XMF, used by Imperium Galactica and some other modules.
  * This format is completely unrelated to the MIDI XMF format.
  */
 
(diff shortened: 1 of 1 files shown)


sezero (approved):

src/loaders/xmf_load.c
```diff
@@ -242,6 +246,11 @@ static void xmf_translate_effect(struct xmp_event *event, uint8 effect, uint8 pa
 		param |= (param << 4);
 		xmf_insert_effect(event, FX_SETPAN, param, chn);
 		break;
+
+	case 0x11:			/* Ultra Tracker retrigger */
+		/* TODO: should support the full param range, needs testing. */
```
sagamusix, line 251: Probably not necessary since the UltraTracker retrigger command is `E9` with a single nibble (like in MOD but supoosedly with some differences in behavior)

Issue #619, opened 2022-11-21
MOD file plays in slow motion?


HunterZ:
UltraX.mod seems to play in slow motion, with a listed duration of almost 40 minutes: https://modarchive.org/index.php?request=view_by_moduleid&query=58591

This is definitely incorrect behavior.

Edit: Sorry, probably should've filed this against libxmp? Let me know if I need to re-file it over there.

sezero:
@AliceLR: Any ideas?

AliceLR:
This is unfortunately a duplicate of #505. Thanks for the reminder, because this really needs to be fixed.

sagamusix:
I will close this duplicate issue. Feel free to subscribe to the related issue if you want to be notified of when this is fixed.

PR #653, opened 2023-05-20
Add new Protracker MOD CIA and VBlank timing detection routine.

### LINKED ISSUE #144: File "WCTIT.MOD" uses Fxx for BPM but it is detected as a Noisetracker mod
A game with this MOD was submitted to us today and libxmp plays it much slower than expected. From the first row of the first pattern it's pretty clearly trying to use F for both speed and BPM, but libxmp detects it as a Noisetracker MOD and enables QUIRK_NOBPM, so the BPM value is read as a speed instead. I've tested it with MPT/libmodplug, OpenMPT, and Protracker clone so far and it works correctly in all three as far as I can tell.

```
Loading WCTIT.MOD (1 of 1)
Module name  : Wall Crusher Theme
Module type  : Noisetracker M.K.
Module length: 8 patterns
Patterns     : 6
Instruments  : 31
Samples      : 31
Channels     : 4 [ 4 b b 4 ]
Duration     : 14min34s (main sequence)
               4min36s (sequence 1 at position 01)
```

[WCTIT.zip](https://github.com/cmatsuoka/libxmp/files/3408014/WCTIT.zip)

### LINKED ISSUE #505: Mod plays extremely slow.
When playing this mod, it plays extremely slow after the first pattern. 

Works fine in Milky and OpenMPT. 

I did play around with the file and removing the Set Tempo in the second pattern makes the module play fine based on a quick test. 

It also seems setting another Tempo anywhere causes it to play slow too. 

[CHAVEZ_-_Flowerpower.zip](https://github.com/libxmp/libxmp/files/7759022/CHAVEZ_-_Flowerpower.zip)

AliceLR:
Replaces the old broken VBlank detection routine with a new, more accurate routine.

Definitions:
* "Low Fxx" is a speed effect with a parameter of 0x1F or lower.
* "High Fxx" is a speed effect with a parameter of 0x20 or higher.

1) Timing detection is performed on 31-instrument MODs with the signature M.K. or M!K!, excluding WOW, OpenMPT MODs, and MODs that don't contain high Fxx. All known Noisetracker and Soundtracker signatures (currently M&K!, FEST) are considered VBlank with no further testing ("fixes" muppenkorva.mod?).
2) Timing is assumed to be CIA by default (fixes ~345 MODs).
3) If a low Fxx and a high Fxx are found on the same row, perform no further checks (MOD is considered CIA). This is a preventative check so the scan compare breaks fewer modules.
4) If a MOD is 8 or more positions long and only contains high Fxx in its last two positions, and the last occurring high Fxx is not F7D, then change to VBlank and perform no further checks (fixes a few VBlank MODs but also lengthens end silence for quite a few others). This check needs future tuning and considerations for MPT MODs.
5) The MOD is marked for scan compare and continues loading as normal. After the scan, if the MOD is determined to be 8 minutes or longer in length, the scan is performed again with VBlank timing. If the MOD is shorter with VBlank timing than with CIA timing, libxmp selects VBlank timing (fixes 56 VBlank modules that weren't detected by the old routine, plus 12 that were; fails to detect 17 VBlank modules that were previously "detected" to be VBlank, and mildly breaks one CIA module).
6) 17 VBlank modules have been added to the module quirks list, as the new routine can't determine their timing. A single VBlank module (mod.siedler ii) has been removed as its timing is correctly detected and needs no further correction.

Fixes #144. Fixes #505. Also see #619.

Data collection spreadsheet for this patch: https://docs.google.com/spreadsheets/d/1a8O9WIjPe8FS3SgyudxFhYlglb7zJb0HpH7FnS0UDCM/edit#gid=0
Extra code for data collection: https://github.com/libxmp/libxmp/compare/master...AliceLR:mod-compare-vblank-diagnostic
All libxmp module quirks list modules I could find, including the 17 new entries: [xmpquirks.zip](https://github.com/libxmp/libxmp/files/11521866/xmpquirks.zip)

~~This may need some regression tests added, marking as a draft for now. edit: yep, needs a couple, will deal with it tomorrow.~~


AliceLR:
Alright, added a few regression tests:

* Chavez - Flowerpower, the buggy module from #505.
* GDM - Space Traveller ][, a CIA module long enough to cause a scan comparison that doesn't use low `Fxx` at all. libxmp scans as CIA, scans as VBlank, and then reverts to CIA.
* Bruno - listen is a VBlank module that uses high `Fxx` near the end only, and is picked up as VBlank only because of the end `Fxx` checks (4 in the procedure described above). Verified that VBlank timing is intended using the music disk it was released on.

sezero (approved):
Looking at the spreadsheet, this certainly is a big improvement, and, as I understand it, there is no fool-proof way of magically fixing everything.

OK by me. Waiting @cmatsuoka and/or @sagamusix to review too.

src/scan.c
```diff
@@ -586,6 +588,56 @@ static int scan_module(struct context_data *ctx, int ep, int chain)
...
+	 * and pick the more likely (i.e. shorter) one. The same logic
+	 * works regardless of the initial mode selected--either way,
+	 * the wrong timing mode usually makes modules MUCH longer. */
+	struct player_data *p = &ctx->p;
+	struct module_data *m = &ctx->m;
+	struct ord_data *info_backup;
+	struct scan_data scan_backup;
+	unsigned char ctrl_backup[256];
+
+	if ((info_backup = malloc(sizeof(m->xxo_info))) != NULL) {
```
sezero, line 613: A `(struct ord_data *) ` cast here

AliceLR: OK, done. I'll combine any further tweaks into the new commit so this doesn't need to be squashed...

sagamusix:
Looks sensible but I didn't do any testing or deeper review.

sezero:
@AliceLR: Should we merge or do you want to wait more and/or have something more to add?

AliceLR:
> @AliceLR: Should we merge or do you want to wait more and/or have something more to add?

I don't have any more to add to this patch.

sezero:
Ok then, merging.

PR #654, opened 2023-05-21
Fixed Watcom builds polluting src directory with artifacts

samizzo:
I thought this was a bit cleaner. It now builds obj files into `obj/<system>` where `<system>` is for example `nt`, `dos4g`, `dos32a` etc and lib/dll files into `lib/<system>`.

### DIFF
--- a/watcom.mif
+++ b/watcom.mif
@@ -14,13 +14,14 @@ STATICFLAGS=-DLIBXMP_STATIC
 LIBROOTDIR=lib
 OBJROOTDIR=obj
 
-# We use the shell's rmdir so we need a backslashes here instead of forward slashes.
+# We use the shell's mkdir and rmdir so we need a backslashes here instead of forward slashes.
 LIBDIR=$(LIBROOTDIR)\$(SYSTEM)
 OBJDIR=$(OBJROOTDIR)\$(SYSTEM)
 
-LOADERSOBJDIR=$(OBJDIR)/loaders
-PROWIZARDOBJDIR=$(OBJDIR)/prowizard
-DEPACKERSOBJDIR=$(OBJDIR)/depackers
+LOADERSOBJDIR=$(OBJDIR)\loaders
+PROWIZARDOBJDIR=$(OBJDIR)\loaders\prowizard
+DEPACKERSOBJDIR=$(OBJDIR)\depackers
+ALL_DIRS=$(LIBDIR) $(OBJDIR) $(LOADERSOBJDIR) $(PROWIZARDOBJDIR) $(DEPACKERSOBJDIR)
 
 DLLNAME=$(LIBDIR)/libxmp.dll
 EXPNAME=$(LIBDIR)/libxmp.exp
@@ -211,13 +212,15 @@ ALL_OBJS+= $(DEPACKER_OBJS)
 !endif
 TEST_OBJS=test/md5.obj test/test.obj
 
-all: $(BLD_TARGET)
+all: $(ALL_DIRS) $(BLD_TARGET)
 
 #.SUFFIXES: .obj .c
 
+$(ALL_DIRS):
+	@!for %i in ($(ALL_DIRS)) do @if not exist %i echo mkdir %i && mkdir %i
+
 .c: src;src/depackers;src/loaders;src/loaders/prowizard;test
 .c.obj:
-	@!if not exist $^: mkdir $^:
 	$(CC) $(LIBFLAGS) -fo=$^@ $<
 
 test/md5.obj: src/md5.c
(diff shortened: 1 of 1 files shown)


AliceLR (changes requested):
`watcom.mif` is generated from `watcom.mif.in` using the GNU Makefile target `watcom-prepare`, so you should apply these changes to `watcom.mif.in` too. The changes themselves look OK to me (not a Watcom user though...).

samizzo:
Oh woops! Thanks!

sezero:
This fails under e.g. os2 console, possibly under windows cmd.exe too, at least with ow1.9:
```
SYS0082: The directory or file cannot be created.
Error occurred while processing: obj\os2v2\.
Error(E42): Last command making (obj\os2v2\virtual.obj) returned a bad status
Error(E02): Make execution terminated
```

`mkdir -p` is the issue. I couldn't find an equivalent for it for non-unix.

sezero:
I'd say that we keep things as is. @AliceLR: Close?

samizzo:
Ah yeah the ow2 docs say -p is supported. This is probably a difference with 1.9. I’ll see if I can make this work with both. On 21 May 2023, at 1:48 pm, Ozkan Sezer ***@***.***> wrote:﻿
This fails under e.g. os2 console, possibly under windows cmd.exe too, at least with ow1.9:
SYS0082: The directory or file cannot be created.
Error occurred while processing: obj\os2v2\.
Error(E42): Last command making (obj\os2v2\virtual.obj) returned a bad status
Error(E02): Make execution terminated

mkdir -p is the issue. I couldn't find an equivalent for it for non-unix.

samizzo:
Can I ask while I have you here - I find the build process a bit confusing. You support autoconf and cmake,  but it seems I have to use autoconf to generate the makefile that generates the watcom makefile, is that right?On 21 May 2023, at 2:10 pm, Ozkan Sezer ***@***.***> wrote:﻿
I'd say that we keep things as is. @AliceLR: Close?

sezero:
> Can I ask while I have you here - I find the build process a bit confusing. You support autoconf and cmake,  but it seems I have to use autoconf to generate the makefile that generates the watcom makefile, is that right?

Yes, that part is a mess..

samizzo:
> I'd say that we keep things as is. @AliceLR: Close?

The problem is at the moment you can't build with Watcom for both MS-DOS and Win32 as the obj and lib files all write to the same destination directories.

sezero:
> > I'd say that we keep things as is. @AliceLR: Close?
> 
> The problem is at the moment you can't build with Watcom for both MS-DOS and Win32 as the obj and lib files all write to the same destination directories.

Yes. You do `wmake -f Makefile.xxx distclean` before building for another target.

AliceLR:
> Can I ask while I have you here - I find the build process a bit confusing. You support autoconf and cmake,  but it seems I have to use autoconf to generate the makefile that generates the watcom makefile, is that right?

The GNU Makefile generates the CMake sources list and the MSVC/Watcom Makefiles so that we don't have to manually manage 4+ different source file lists in separate places. Most people who are just building from the Makefiles aren't affected by this. It is a little inconvenient in cases like this, though.

> I'd say that we keep things as is. @AliceLR: Close?

I don't know enough about this build system to have a good answer either way.

samizzo:
> Yes. You do `wmake -f Makefile.xxx distclean` before building for another target.

Right, but I can't build one and then build the other without the libs and obj files being removed first. I'm sure I can make this work under OW 1.9 and 2.0 etc. I don't personally think it's very clean to put build artifacts in the source directory and the build outputs in the root directory. But if you don't want this change that's fine.

samizzo:
> The GNU Makefile generates the CMake sources list and the MSVC/Watcom Makefiles so that we don't have to manually manage 4+ different source file lists in separate places. Most people who are just building from the Makefiles aren't affected by this. It is a little inconvenient in cases like this, though.

Gotcha, that makes sense. I was thrown by those generated files being checked into the repo which was why I got confused by the process and modified the wrong files. I'm used to other projects where they wouldn't even check generated files in; you'd have to generate everything from scratch. I appreciate having the makefiles right there - that's very handy!

sezero:
> > Yes. You do `wmake -f Makefile.xxx distclean` before building for another target.
> 
> Right, but I can't build one and then build the other without the libs and obj files being removed first. I'm sure I can make this work under OW 1.9 and 2.0 etc. I don't personally think it's very clean to put build artifacts in the source directory and the build outputs in the root directory. But if you don't want this change that's fine.

Yes, the idea is good. But please make it work with ow1.9. If not, we should keep things as is.

samizzo:
Woops sorry I didn't mean to push all those other changes into this PR! I'm going to close this and open a new PR when I've done this properly (I have it working now on OW 1.9, 2.0 and also under OS/2).

sezero:
Make sure that it still works under unix (linux) too: with those last commits, it fails.
As I suspected, it may not be worth the effort but feel free to try.

PR #655, opened 2023-05-21
Fixed obj files polluting source tree under Watcom.

samizzo:
Ok, this should work now. I tested under OS/2 Warp 4.5 with Open Watcom 1.9, Fedora Linux with Open Watcom 1.9, and Windows command prompt with Open Watcom 1.9 and 2.0. It doesn't work under dosbox-x (and presumably real MS-DOS) due to the rmdir and mkdir spawning a shell with a command-line which is too long. I have a local fix for that which is a bit ugly, but even with that fix it doesn't build under MS-DOS due to long filenames in some of the source files. So I figured it wasn't worth it.

### DIFF
--- a/watcom.mif
+++ b/watcom.mif
@@ -11,12 +11,49 @@ CFLAGS += -Iinclude
 DLLFLAGS=-bd -DBUILDING_DLL
 STATICFLAGS=-DLIBXMP_STATIC
 
-DLLNAME=libxmp.dll
-EXPNAME=libxmp.exp
+LIBROOTDIR=lib
+OBJROOTDIR=obj
+!ifeq __LINUX__
+PATHSEP=/
+!else
+# Under MS-DOS, Windows, and OS/2 we use the shell's mkdir and
+# rmdir, so we need a backslashes instead of forward slashes.
+PATHSEP=\
+!endif
+
+LIBDIR=$(LIBROOTDIR)$(PATHSEP)$(SYSTEM)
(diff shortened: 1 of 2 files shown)


sezero:
Well, my eyes are burning with conditionals. @AliceLR: Do you think that we should take this?

A few notes, if this is to be applied: 

The `lite` side should be patched accordingly.

Didn't know that del command supported arguments: I hope that it's supported in earliest versions of windows too.

`__LINUX__` checks can (should) be replaced with `__UNIX__`

AliceLR:
> Well, my eyes are burning with conditionals. @AliceLR: Do you think that we should take this?

This seems fine to me at a glance, with the same caveat I mentioned last time: I don't have Watcom set up right now.

samizzo:
> Well, my eyes are burning with conditionals. @AliceLR: Do you think that we should take this?
Haha yeah, that's the one downside! Every platform handles things differently.

> The `lite` side should be patched accordingly.
Ah yes, will do, thanks!

> Didn't know that del command supported arguments: I hope that it's supported in earliest versions of windows too.
`del` has supported a couple of args since Windows 2000 and later so hopefully that's adequate.

> `__LINUX__` checks can (should) be replaced with `__UNIX__`
I don't think that makes sense though unless you want to support QNX. The wmake docs say `__UNIX__` is defined in Linux and QNX environments. http://openwatcom.org/ftp/install provides packages for OS/2, MS-DOS, Win32, and Linux but no other Unixes or QNX. If you're concerned about running the Linux build under emulation like say on a BSD I can test that on FreeBSD.

PR #660, opened 2023-06-13
bump version to 4.6.0 and update the changelogs

sezero:


sezero:
Alice: You probably hand-pick your versioned symbol changes from https://github.com/libxmp/libxmp/pull/651 into here.

AliceLR:
> Alice: You probably hand-pick your versioned symbol changes from #651 into here.

I don't think #651 will be ready in time, partially or otherwise—even the bare minimum MD5 time reduction, filtering on length before performing the MD5, is blocked because I can't find two of the quirks list entries. The symbol versions in that patch are kind of just a placeholder right now :(

I'll review this in a bit and double check the changelog. (I think after this release I'll start properly tracking my changes in the changelog, sorry...)

docs/Changelog
```diff
@@ -27,6 +32,8 @@ Stable versions
 	- Fix support for most Farandole Composer effects.
 	- Implement Ultra Tracker tone portamento.
 	- Multiple OctaMED fixes and support for MMDC packed modules.
+	- Fix MMD0 pattern loader to honor play transpose setting.
```
AliceLR, line 35: This MMD0 pattern loading bugfix was @sagamusix.

docs/Changelog
```diff
@@ -37,6 +44,7 @@ Stable versions
 	- Allow xmp_smix_play functions to play key off, cut and fade events.
 	- Integrate the OXM depacker with the XM loader.
```
AliceLR, line 45: @ccawley2011 merged the OXM depacker into the XM loader.

AliceLR (commented):
I didn't notice anything especially important missing, but a couple of things are misattributed to me.

I'll go prepare a separate patch to update file copyright years since we let those slide again..

AliceLR:
`README` also has a version number that needs to be updated.

docs/Changelog
sezero: Fixed

sezero:
Force-pushed to fix the wrong/missing items noted.

AliceLR (approved):

Issue #663, opened 2023-07-22
unlha using lhasa


sezero:
Here is initial work on replacing our LGPL'ed unlha with code using
[lhasa](https://github.com/fragglet/lhasa): Passes devcheck for me.

The test function is untouched: It is limited and liblhasa supports
more compression methods. Can be extended if so wanted.

For simplicity and for being proof-of-concept, this currently links
to liblhasa. If it is good, I can extract necessary-only parts from
lhasa into our tree.

Related issue: #387 

CC: @AliceLR 

```c
#include "../common.h"
#include "depacker.h"
#include "lha_reader.h"

static int lhasa_read(void *handle, void *buf, size_t buf_len) {
    size_t r = hio_read(buf, 1, buf_len, (HIO_HANDLE*)handle);
    if (!r && ((HIO_HANDLE*)handle)->error) return -1;
    return (int)r;
}

static int lhasa_skip(void *handle, size_t bytes) {
    return !(hio_seek((HIO_HANDLE*)handle, (long)bytes, SEEK_CUR) < 0);
}

static void lhasa_close(void *handle) {
    (void) handle;      /* do nothing. */
}

static const LHAInputStreamType io_callbacks = {
    lhasa_read,
    lhasa_skip,
    lhasa_close
};

static int test_lha(unsigned char *b) {
	return b[2] == '-' && b[3] == 'l' && b[4] == 'h' && b[6] == '-' &&
		b[20] <= 3;
}

static int decrunch_lha(HIO_HANDLE *in, void **out, long *outlen)
{
    LHAInputStream *stream;
    LHAReader *reader;
    LHAFileHeader *header;
    unsigned char *outbuf;
    int error = -1;

    stream = lha_input_stream_new(&io_callbacks, in);
    if (!stream) {
        return -1;
    }

    reader = lha_reader_new(stream);
    if (!reader) {
        lha_input_stream_free(stream);
        return -1;
    }

    for (;;) {
        header = lha_reader_next_file(reader);
        if (!header) {
            break;
        }
        if (!strcmp(header->compress_method, LHA_COMPRESS_TYPE_DIR)) {
            continue;   /* directory or symlink */
        }
        if (!libxmp_exclude_match(header->filename)) {
            break;
        }
    }

    if (!header || (long)header->length <= 0) {
        goto fail;
    }

    outbuf = (unsigned char *) malloc(header->length);
    if (!outbuf) {
        goto fail;
    }
[...]

Issue #387, opened 2021-06-23
make the library MIT-licensed?


sezero:
> The majority of libxmp files is MIT, but the library itself is LGPL because there are a few third party components (unzip, etc) that are LGPL. The "lite" library version uses only MIT-licensed sources so the entire library can be released under the MIT license, which is more permissive (but the "lite" version doesn't include prowizard-pc files). ...

For what it's worth, I went through the libxmp files to figure out what is actualy LGPL vs. MIT, and in the case of `--disable-depackers` only *one* active loader actually pulls in LGPL code (Digital Symphony pulls in the LGPL LZW depacker, but since this loader needs attention anyway #347, I've thought about just writing it a new sample depacker). So if the depackers do get separated to a second library, it's very possible the entirety of libxmp could be placed under the MIT license. The remaining LGPL code:

* `fmopl.c`— currently deadcode. If OPL support gets readded for e.g. S3Ms, a bugfixed Opal might be a suitable replacement. Opal (public domain) is the RAD v2 OPL3 emulator, and is used by OpenMPT and other projects.
* `fcm_load.c`, `ftm_load.c`, `hvl_load.c`, `polly_load.c`, `ssmt_load.c`, `stc_load.c` are all deadcode right now.

Also, thanks for clarifying ProWizard's licensing. :)

_Originally posted by @AliceLR in https://github.com/libxmp/libxmp/issues/214#issuecomment-856268075_

sezero:
I guess this requires ditching the depackers (in some way) and the particular dead
code listed above, and adopting a suitably licensed lzw depacker for `sym_load.c`.

sezero:
About  lzw depacker for `sym_load.c`: [liblzw](https://github.com/vapier/liblzw) (under unlicense) maybe?

AliceLR:
A more thorough breakdown:

* The affected loaders might just as well be relicensed (or replaced if it comes to it).
  * `fcm_load.c` predates the Git repository. ProWizard supports this format so this can probably be junked.
  * `ftm_load.c` predates the Git repository.
  * `hvl_load.c` was [added in 2010](https://github.com/libxmp/libxmp/commit/37ff3c88e79475dd3b78070f10fe46d2fcdaff5a).
  * `polly_load.c` was [added in 2010](https://github.com/libxmp/libxmp/commit/58a467a6b82c543e5a8f92a4618c7f9d943dc305).
[...]

PR #664, opened 2023-07-24
Change lha depacker to use a stripped down / modified version of lhasa.

sezero:
With this, the last bit of LGPL code is gone.

The ubsan failures reported to mainstream as https://github.com/fragglet/lhasa/issues/49

Reference issue: https://github.com/libxmp/libxmp/issues/663
Reference issue: https://github.com/libxmp/libxmp/issues/387


sezero:
Pushed bd8ba6e to fix UBSan errors peek_bits() until mainstream fixes it
in a better way.

Please review, and feel free to push to my branch if necessary .

AliceLR:
Okay, that was shorter than I thought and it *seems* like it should be able to read block-wise from a stream. Dynamically growing the buffer is a problem on our end, then. Do you want to deal with it now (see unxz.c) or wait for slow load files?

I'm getting several crashes in tests locally for some reason (MSYS2/MinGW64), trying to figure out why right now...

sezero:
> WOW this is a lot. I'm going to skip reviewing the lhasa-specific files and let libFuzzer sort that out for us :)

OK :)

> Everything but unlha.c looks good, I need some time to check unlha.c though. A general concern—did you see anything that looks like lhasa dynamically resizing the output buffer, or does it try to allocate the whole output buffer in one go? If the latter, that's going to be a problem.

I'm not sure that I understand the question. We allocate the output buffer by ourselves in unlha.c and pass it to lhasa (to `lha_reader_read()`) along with its size, and it doesn't ever resize our buffer but only write to it (see `lha_decoder_read()`). Does that address the concern or am I missing anything?

AliceLR:
> I'm not sure that I understand the question. We allocate the output buffer by ourselves in unlha.c and pass it to lhasa (to `lha_reader_read()`) along with its size, and it doesn't ever resize our buffer but only write to it (see `lha_decoder_read()`). Does that address the concern or am I missing anything?

Yes, this is what I meant—I hadn't looked at unlha.c yet. This can cause slow enough loads with some depackers (bunzip2.c, unxz.c) that I've had to change them to resize the buffer by powers of 2 instead of allocating them with a single call—this is a lot faster for invalid files and thus libFuzzer.

sezero:
> I don't think it's a big enough deal to block this, I can just patch it in later.

OK, as you wish

AliceLR (approved):
OK, let's get this in already...

sezero:
Note that this adds `mktime` among the libc calls used, which we haven't
used before: It calls it when decoding timestamps. We don't really need
the timestamps and I first wanted to remove all timestamp decoding & etc
but macbinary.c depends on a timestamp comparison, so I kept them as is.
Not anything serious at all, but just wanted to note...

sezero:
>  skip reviewing the lhasa-specific files and let libFuzzer sort that out for us :)

Any news from fuzzing?

AliceLR:
> Any news from fuzzing?

So far, nothing. No ASan crashes yet, and the only slow loads and timeouts have all been caused by the stb_vorbis `qsort` (already reported upstream).

sezero:
That's great!

PR #668, opened 2023-07-24
change license to MIT

sezero:
Closes: https://github.com/libxmp/libxmp/issues/387


cmatsuoka (approved):
Looks nice, thanks!

sezero:
Thanks, this is in now.

One question remains:

Should we now completely remove the lite packaging, or should we still
keep it as an option either as a separate package as we used to, or as
part of main package but with autotools/cmake config option?

PR #669, opened 2023-07-25
depackers/lhasa: sync ubsan fix with mainstream and tweak

sezero:


AliceLR (approved):

sezero:
P.S.: This would have better got merged without squash. If you are as sick of merge commits in git history as I am, we should probably enable the rebase method in settings? (Claudio had it disabled back then, can't remember why..)

AliceLR:
Sorry, I wasn't really thinking when I squashed it, that was just what I already had selected. I think merge commits are fine for patch collections (like this or the fuzzer fixes) and squashes are fine for messy trees. I've never used the rebase merge and I don't especially trust it given my rebase experiences...

PR #671, opened 2023-07-26
minor Makefile.lite and autotools updates

sezero:
- Makefile.lite: tweak DFILES:
  Copy configure.ac and Makefile.in from the lite directory, not from
  top-level. And remove configure script from DFILES: The rest of the
  rules already run autoconf in the staging directory i.e. it will be
  generated already.  No need to generate the configure script before
  running Makefile.lite, anymore.

- autotools: allow specifying the autoconf command in AUTOCONF env. var.
  (on the make command line in case of Makefile.lite.)

### DIFF
--- a/Makefile.lite
+++ b/Makefile.lite
@@ -1,9 +1,8 @@
 
 DIST		= libxmp-lite-stagedir
-DFILES		= lite/README INSTALL install-sh configure configure.ac \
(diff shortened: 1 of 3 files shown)


AliceLR (approved):
This looks OK.

(Had to double check the autogen.sh thing is [actually POSIX](https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_06_02)... )

PR #672, opened 2023-07-26
cmake: set maximum policy to 3.5

sezero:
This fixes the following warning when configuring with CMake 3.27+:
```
CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

@madebr: This was adapted from your SDL2 patch. OK?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.2)
+cmake_minimum_required(VERSION 3.2...3.5)
 project(libxmp VERSION 4.6.0 LANGUAGES C)
 
 set(LIBXMP_DEFINES)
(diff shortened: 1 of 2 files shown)


AliceLR:
> ```
> CMake Deprecation Warning at CMakeLists.txt:1 (cmake_minimum_required):
>   Compatibility with CMake < 3.5 will be removed from a future version of
>   CMake.
> 
>   Update the VERSION argument <min> value or use a ...<max> suffix to tell
>   CMake that the project does not need compatibility with older versions.
> ```

Not relevant to your patch, but given the `...<max>` suffix wasn't introduced until [CMake 3.12](https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html) this seems like a completely unforced compatibility break on their end. CMake 3.12 is only 5 years old. This is kind of absurd.

AliceLR (approved):

PR #677, opened 2023-09-06
Fix -Wundef warning, compile with -Wundef when available.

AliceLR:
Here's another warning flag I build libxmp with downstream that wasn't part of the actual buildsystem :) The instance of it I fixed is trivial but this one can detect *actual* bugs, so I figured it was worth adding.


sezero (approved):

sezero:
Pushed a follow-up to your branch to do the same in lite side and also in cmake builds. Will squash-merge when CI runs complete.

PR #678, opened 2023-09-09
Fix XM envelope handling

neumatho:
This pull request will will improve the envelope processing for XM modules, so e.g. Ebony Owl Netsuke.xm sound correctly.

A new function has been created to be used for XM only and the old XM function has been renamed to Default instead, to make it backward compatible with other module formats.

I have also created a new QUIRK flag to indicate if the new processing should be used. I could have used FT2BUGS, but I didn't think this is a bug handling. But if you want me to change it, I'm willing to do so :-)

-Thomas


sezero:
Here is a fix to test-dev side: (along with two minor whitespace fixes):

```diff
diff --git a/test-dev/test_player_xm_envelope.c b/test-dev/test_player_xm_envelope.c
index ab16841..31feee6 100644
--- a/test-dev/test_player_xm_envelope.c
+++ b/test-dev/test_player_xm_envelope.c
@@ -17,3 +17,3 @@ TEST(test_player_xm_envelope)
 
- 	create_simple_module(ctx, 2, 2);
+	create_simple_module(ctx, 2, 2);
 
@@ -21,3 +21,3 @@ TEST(test_player_xm_envelope)
 	set_instrument_envelope(ctx, 0, 1, 16, 0);
-	set_instrument_envelope_loop(ctx, 0, 0, 0, 0);
+	set_instrument_envelope_loop(ctx, 0, 0, 0);
 
diff --git a/test-dev/test_player_xm_envpos.c b/test-dev/test_player_xm_envpos.c
index b0b0ed8..a1c2d7e 100644
--- a/test-dev/test_player_xm_envpos.c
+++ b/test-dev/test_player_xm_envpos.c
@@ -12,3 +12,2 @@ TEST(test_player_xm_envpos)
 	struct context_data *ctx;
-	int i;
 
@@ -17,3 +16,3 @@ TEST(test_player_xm_envpos)
 
- 	create_simple_module(ctx, 2, 2);
+	create_simple_module(ctx, 2, 2);
 
```

neumatho:
Thanks for the fixes.

sezero:
OK thanks.

Let's wait for approvals from @AliceLR and/or @sagamusix

AliceLR:
I will test this when I get a chance, but my initial thoughts: the new envelope function is *very* similar to the old one and it is worth testing the other formats that use this envelope handler to see if they exhibit similar behavior (or even care). If it's similar enough, I don't think there's a reason to add a new function and quirk flag. Formats that should be checked for the same behavior:

* DigiBooster Pro (dbm_load.c) - has sustain points AND a set envelope position effect `Lxx`.
* Imago Orpheus (imf_load.c) - has sustain points.
* Digitrakker MDL (mdl_load.c) - has sustain points.
* Real Tracker (rtm_load.c) - has sustain points AND a set envelope position effect `Lxx`.

Probably can be ignored:

* StarTrekker FLT4/8 (flt_load.c) - uses very simple envelopes to simulate some instrument extensions.
* J2B / Galaxy Music System 4.0 (gal4_load.c) - I think these are conversions of S3M and sample mode IT...

@sagamusix please check this info :(

sagamusix:
Sorry, don't have much capacity for anything libxmp right now, but Digitrakker and Imago Orpheus instruments are both inspired by XM instruments, so they should most likely follow XM behaviour. The same *probably* applies to DBPro.
There are no "official" J2Bs that use instrument mode at all, but there are many J2Bs from custom levels which have been converted from XM/IT that use instruments. It is worth noting that Galaxy is very XM-centered, so it will most likely behave more like XM than IT (there is no reliable way to tell if the source of the J2B file was an XM or IT).

sezero:
Progress? Verdict?

AliceLR:
> Progress? Verdict?

No updates yet, I've been busy.

AliceLR:
Real Tracker 2 relies on the "default" variant 🤦‍♀️ will update with further findings, but this at least supports having two XM-style envelope routines.

AliceLR:
I think I'd like to tidy up some comments/other mostly cosmetic tweaks, as well as expand the test cases tomorrow, but fundamentally this patch is fine. Sorry for taking so long to review it.

AliceLR:
The stuff I wanted to tweak is in now. @sezero, if this looks okay to you and I didn't break anything, this should be good to squash.

AliceLR (approved):

sezero:
>  @sezero, if this looks okay to you and I didn't break anything, this should be good to squash.

The only minor build issue I see is with xmp-lite when configured with --disable-it 
where we see this warning:
```
src/player.c: In function ‘update_envelope’:
src/player.c:275: warning: unused variable ‘m’
```

I can either silence it by addiing a `(void) m`, or I can revert the `it_env` argument
removal from `update_envelope` -- what is your preference?

AliceLR:
> I can either silence it by addiing a `(void) m`, or I can revert the `it_env` argument removal from `update_envelope` -- what is your preference?

Oh, ew. Both `HAS_QUIRK` and `IS_PLAYER_MODE_IT` rely on `m`, so reverting won't work. I think suppressing it or an `#ifdef` might be the best way to deal with it.

sezero:
OK, this is in.

AliceLR:
I should have checked it better but Imago Orpheus and Digitrakker 3.2 weren't importing the Lxx effects at all (consistent with my reply in September), which sounds pretty much identical to the loop end being clamped. This doesn't affect the patch in any way.

sezero:
Do you have a patch?

AliceLR:
There is no patch, those two formats are no longer relevant to the issue. The comment should be updated to remove them but that's about it.

PR #684, opened 2023-10-03
Fixing end-detection

neumatho:
This little fix will make sure if a sequence jumps back to before it starting position, the loop count will be incremented. This will tell a player, that the module has ended.

An example of this, is sequence 8 of "alien incident - leohou2.s3m" by Purple Motion.

### DIFF
--- a/src/player.c
+++ b/src/player.c
@@ -1634,10 +1634,12 @@ static void next_order(struct context_data *ctx)
 			if (mod->rst > mod->len ||
 			    mod->xxo[mod->rst] >= mod->pat ||
 			    p->ord < m->seq_data[p->sequence].entry_point) {
-				p->ord = m->seq_data[p->sequence].entry_point;
 				/* Increment to loop count. This will make the player to quit like other modules when
 				   playing sequence 8 of "alien incident - leohou2.s3m" by Purple Motion */
-				p->loop_count++;
+				if (p->ord < m->seq_data[p->sequence].entry_point) {
+					p->loop_count++;
+				}
+				p->ord = m->seq_data[p->sequence].entry_point;
 			} else {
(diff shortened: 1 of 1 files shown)


AliceLR:
Given that the test that broke has *working* loop detection and relies on the `if` branch this patch modifies, I suspect this is not the correct fix. I'll look at the module and get back to you.

neumatho:
The reason I need the extra if statement, is because the test module reached the new code, but it also reaches check_end_of_module() where it also increment the loop_count. The loop_count will then be incremented twice, which is wrong. That's why I added the if.

The Purple Motion module have a Bxx effect, where it jump to a position before the current sequence position. This is not detected in check_end_of_module() and therefore I added it in where the position is reset to the start position of current sequence.

If it can be placed elsewhere, e.g. in check_end_of_module(), then ofcouse it would be better :-)

AliceLR:
> If it can be placed elsewhere, e.g. in check_end_of_module(), then ofcouse it would be better :-)

Right, that'd be ideal--I suspect `loop_count` can still be doubly incremented in some cases with this but I don't have any proof. There's clearly a defect in the loop detection in any case.

AliceLR:
These were actually caused by a pair of related scan bugs that placed the loop point inside of a marker. For now I've written a patch and regression test that fixes both of them in the scan, since I don't see loop points in markers as having much use: #688

neumatho:
Ok, I guess we can close this then?

sezero:
> Ok, I guess we can close this then?

I think so. @AliceLR ?

AliceLR:
Unless there are more looping issues, yeah, I think so.

sezero:
> Unless there are more looping issues, yeah, I think so.

If there are other looping issues, they should be tracked in new tickets. Closing this one.

PR #686, opened 2023-10-07
Makefile.in: install libxmp.so with -m755 not -m644

sezero:


AliceLR (approved):

sezero:
Hmph. I seem to have missed doing the same for lite version. Will do that directly (w/o a PR) later.

sezero:
> Hmph. I seem to have missed doing the same for lite version. Will do that directly (w/o a PR) later.

Applied as https://github.com/libxmp/libxmp/commit/51facd1a00704ea4cb0d56d7cbb520ca43c727ca

PR #687, opened 2023-10-13
Initial implementation of HIO buffering

ds-sloth:
I noticed that my game engine using libxmp was hanging while loading music, and when profiling found a massive hotspot in the near-millions of single-byte(!!!) stdlib read calls. Upwards of 90% of the load time was spent in these functions. I see that AliceLR also noticed this issue and made a complete fix for it for one loader in #662, but you could at least partially fix it for all loaders by buffering reads at the HIO abstraction level.

This PR is a first pass at doing this. It creates a 4096+4 byte buffer inside the `HIO_HANDLE` struct, which is refilled with a new 4096 bytes from raw IO whenever the cursor enters the +4 byte tail of the buffer. I attempted to be paranoid about how the stdlib handles reading and seeking (but maybe wasn't paranoid enough). In my local benchmarks, these changes accelerate loading from disk via callbacks by (roughly) a factor of 8.

If accepted, this change would also allow a lot of the code duplicated across the three I/O interfaces to be removed.

I'm not familiar with the codebase or its style, so it's very possible that something here doesn't work (possibly badly). I look forward to your review!

### DIFF
--- a/src/hio.c
+++ b/src/hio.c
@@ -156,13 +156,6 @@ static int hio_eof_internal(HIO_HANDLE *h)
 	return EOF;
 }
 
-static int hio_error_internal(HIO_HANDLE *h)
-{
-	int error = h->error;
-	h->error = 0;
-	return error;
-}
-
 static void hio_fill_buffer_internal(HIO_HANDLE *h)
 {
 	if (h->buffer_end == HIO_FULL_BUFFER) {
@@ -178,6 +171,11 @@ static void hio_fill_buffer_internal(HIO_HANDLE *h)
 		/* read a full buffer from file */
 		h->buffer_end = hio_read_internal(h->buffer, 1, HIO_FULL_BUFFER, h);
 	}
+
+	/* don't set EOF error yet */
+	if(h->error == EOF) {
+		h->error = 0;
+	}
 }
 
 int8 hio_read8s(HIO_HANDLE *h)
@@ -185,6 +183,7 @@ int8 hio_read8s(HIO_HANDLE *h)
 	int8 ret;
 
 	if (HIO_BUFFER_CURSOR(h) >= h->buffer_end) {
+		h->error = EOF;
 		return ret;
 	}
 
@@ -203,6 +202,7 @@ uint8 hio_read8(HIO_HANDLE *h)
 	uint8 ret;
 
 	if (HIO_BUFFER_CURSOR(h) >= h->buffer_end) {
+		h->error = EOF;
 		return ret;
 	}
 
@@ -223,6 +223,7 @@ uint16 hio_read16l(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 1 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -246,6 +247,7 @@ uint16 hio_read16b(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 1 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -269,6 +271,7 @@ uint32 hio_read24l(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 2 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -293,6 +296,7 @@ uint32 hio_read24b(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 2 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -317,6 +321,7 @@ uint32 hio_read32l(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 3 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -342,6 +347,7 @@ uint32 hio_read32b(HIO_HANDLE *h)
 
 	if (HIO_BUFFER_CURSOR(h) + 3 >= h->buffer_end) {
 		h->cursor += h->buffer_end - HIO_BUFFER_CURSOR(h);
+		h->error = EOF;
 		return ret;
 	}
 
@@ -385,7 +391,7 @@ size_t hio_read(void *buf, size_t size, size_t num, HIO_HANDLE *h)
 	should_read -= buffer_left;
 	h->cursor += buffer_left;
 
-	/* read the rest directly */
+	/* read the rest directly, may set h->error */
 	did_read = hio_read_internal(buf, 1, should_read, h);
 	h->cursor += did_read;
 
@@ -404,6 +410,7 @@ size_t hio_read(void *buf, size_t size, size_t num, HIO_HANDLE *h)
 int hio_seek(HIO_HANDLE *h, long offset, int whence)
 {
 	int ret;
+	int seek_error;
 	ptrdiff_t ofs = offset;
 
 	/* special case for SEEK_END, will need to rewind to fill buffer */
@@ -434,8 +441,14 @@ int hio_seek(HIO_HANDLE *h, long offset, int whence)
 	ofs = (ofs / HIO_BUFFER_SIZE) * HIO_BUFFER_SIZE;
 
 	ret = hio_seek_internal(h, ofs, SEEK_SET);
+	seek_error = h->error;
 
 	h->buffer_end = hio_read_internal(h->buffer, 1, HIO_FULL_BUFFER, h);
+
+	/* don't set EOF error yet */
+	if(h->error == EOF) {
+		h->error = seek_error;
+	}
 
 	return ret;
 }
@@ -455,10 +468,9 @@ int hio_eof(HIO_HANDLE *h)
 
 int hio_error(HIO_HANDLE *h)
 {
-	if (HIO_BUFFER_CURSOR(h) < h->buffer_end)
-		return 0;
-
-	return hio_error_internal(h);
+	int error = h->error;
+	h->error = 0;
+	return error;
 }
 
 HIO_HANDLE *hio_open(const char *path, const char *mode)
--- a/src/hio.c
+++ b/src/hio.c
@@ -45,207 +45,7 @@ static long get_size(FILE *f)
 	}
 }
 
-int8 hio_read8s(HIO_HANDLE *h)
-{
-	int err;
-	int8 ret;
-
-	switch (HIO_HANDLE_TYPE(h)) {
-	case HIO_HANDLE_TYPE_FILE:
-		ret = read8s(h->handle.file, &err);
-		break;
-	case HIO_HANDLE_TYPE_MEMORY:
-		ret = mread8s(h->handle.mem, &err);
-		break;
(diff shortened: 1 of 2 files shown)


ds-sloth:
Very cool CI tests you've got! I fixed two major notes (invalid pointer arithmetic with a `void*`, and uninitialized values being returned -- I thought I saw that in the base code for failure conditions so I wanted to follow it even though I wouldn't normally do that).

sezero:
`test_effect_it_break_to_row` seems to get stuck in an endless loop

sezero:
> `test_effect_it_break_to_row` seems to get stuck in an endless loop

And there are those many test suite failures, but this one is most annoying.

AliceLR:
The workaround MegaZeux does is using `fopen` and `setvbuf` manually, but this isn't a flawless solution (it seems some of the SDKs for consoles we target have broken `setvbuf`). If it's an option for you and you haven't tried it yet, I would recommend that as a stopgap.

Do you know which loaders were causing your slow loads? They will need to be addressed (if possible) in addition to this patch.

ds-sloth:
> many test suite failures

I now see how to build the unit tests suite locally. I'll update this PR after checking what's going on with them. Thanks! (But, if you don't like the general idea of the PR, or if you like it but want to implement it yourself in a different way, please let me know, since I have limited time for projects.)

> setvbuf

I hadn't heard of that but it sounds useful. Do you know whether 3DS supports it? That's our weakest target and the one I'm noticing load latency on. (Note that I'm already loading everything out of a romfs to avoid the 3DS's terrible sdcard filesystem oscall performance.)

> They will need to be addressed (if possible) in separate patches

Yes, IT is the loader I've been using, and (strangely) even though I get a very large speedup on desktop from this patch, now that I am testing it on 3DS the speedup is more modest. I don't know about others on 3DS, I'll copy some of the libxmp test suite over and check. I'll note that libmodplug *loads* my IT tracks much faster, but *plays* them an order of magnitude slower.

AliceLR:
> 3DS

Ah, this patch makes a lot more sense now. devkitARM/libctru *does* support `setvbuf` for the 3DS. Currently MZX is preloading files to memory for this platform, but `setvbuf` with a 32k buffer seemed to work well too. `stat` and dirent also contribute to the slowness :(

The platforms/SDKs with broken `setvbuf` that I'm aware of are DS/BlocksDS (not implemented apparently) and Vita/VitaSDK (last tested several years ago but it either didn't work or was still very slow).

> Yes, IT is the loader I've been using, and (strangely) even though I get a very large speedup on desktop from this patch, now that I am testing it on 3DS the speedup is more modest. I don't know about others on 3DS, I'll copy some of the libxmp test suite over and check. I'll note that libmodplug _loads_ my IT tracks much faster, but _plays_ them an order of magnitude slower.

libmodplug preloads the entire file to memory, so it makes sense that loading is faster. I see some things that *might* be tweakable in the IT loader, but compressed samples in particular probably need this patch.

sezero:
>  I'll note that libmodplug _loads_ my IT tracks much faster, but _plays_ them an order of magnitude slower.

libmodplug accepts input as a memory buffer, so you can always use `xmp_load_module_from_memory` of libxmp to match that. If you do that, load libxmp still load slower than libmodplug?

AliceLR:
> libmodplug accepts input as a memory buffer, so you can always use `xmp_load_module_from_memory` of libxmp to match that. If you do that, does libxmp still load slower than libmodplug?

I think libmodplug's loading time is going to be hard to beat due to all the awful hackery in that library, but it'd be nice if we can get close at least...

ds-sloth:
> the speedup is more modest [on 3DS]

To update this, a 64KB buffer (which I wouldn't be comfortable proposing by default) seems to help substantially more than a 4KB one.

> I'll copy some of the libxmp test suite over and check

I did this. Out of the test cases that were recognized by our engine's file selector, MOD, STM, and XM are fast (UI seems responsive when loaded on main thread); 669, ULT, IT, MED, and S3M are moderate (~1/2s) ; STX, GDM, FAR, and AMF are slow (>=3s).

I'm not noticing an enormous difference with modplug on these (same ones are fast and slow), but the moderate ones are noticeably faster.

In my own use case, I'm using a lot of IT files generated by my spc2it fork. (It's ugly data; opening and saving with OpenMPT reduces the filesize by around half, but I don't have that automated.) I'd say around 75% of them feel "fast" with libmodplug and only 25% of them feel "fast" with libxmp (even with buffering). Without buffering, many of them take over a second to load with libxmp.

> Currently MZX is preloading files to memory for this platform
> If you do that, does libxmp still load slower than libmodplug?

That's a good idea. I'll try that out tomorrow. If that's the easier solution (and especially since my failed test cases might mean it's nontrivial to make HIO use buffering), maybe a more useful patch to libxmp would be to provide an option for it to prefer preloading files into its `MFILE` structures.

ds-sloth:
It turns out that our mixer library already had support for `xmp_load_module_from_memory`. I tested it with my use case, and it seems to have a similar effect as the buffering I'm proposing here. Both of them reduce load times from ~2.5-5s to ~.5-1s. That's still quite a bit slower than libmodplug (~.1-.5s) for my IT files, but it's a lot more reasonable.

I'm happy to test anything else out that you think might be interesting here, or I could begin profiling (I use callgrind) and thinking about hot code in the IT loader if that'd be helpful. But either of those will have to wait until tomorrow.

AliceLR:
> I did this. Out of the test cases that were recognized by our engine's file selector, MOD, STM, and XM are fast (UI seems responsive when loaded on main thread); 669, ULT, IT, MED, and S3M are moderate (~1/2s) ; STX, GDM, FAR, and AMF are slow (>=3s).

Thanks for this list. From memory and/or a quick code inspection:

* IT: needs pattern buffering. Compressed samples need block buffering.
* FAR: should be faster than this, but can be improved further.
* MED: some work can be done here, but these formats are fairly heavyweight by design.
* STM: I think this might have been due to a lucky test file—needs pattern buffering.
* 669, AMF (DSMI), GDM, S3M: pattern buffering.
* ULT: track buffering.

I'll start on a patchset for these tomorrow (hopefully). STX might be trickier to do efficiently.

> In my own use case, I'm using a lot of IT files generated by my spc2it fork. (It's ugly data; opening and saving with OpenMPT reduces the filesize by around half, but I don't have that automated.) I'd say around 75% of them feel "fast" with libmodplug and only 25% of them feel "fast" with libxmp (even with buffering). Without buffering, many of them take over a second to load with libxmp.

Sounds like you have sample compression turned on in OpenMPT—libmodplug's sample decompression routine is entirely memory based, but when it was adapted to libxmp, this was replaced by single-byte filesystem calls. This can be improved...

> (and especially since my failed test cases might mean it's nontrivial to make HIO use buffering)

I don't see any reason HIO buffering should cause problems conceptually—I'll let you know what I find when I get to reviewing this patch.

> provide an option for it to prefer preloading files into its `MFILE` structures.

This is definitely doable for the loading functions via a new `libxmp_set_player` option (and a minor version bump), but not for the standalone test functions.

sezero:
I also wonder how successful would it be if the buffering etc are implemented at library's user and the library is interacted with `xmp_load_module_from_callbacks`

ds-sloth:
> Sounds like you have sample compression turned on in OpenMPT

I don't think so; I think my auto-generated IT files are just full of garbage. I looked up and checked the setting you were referring to, and it was disabled. Enabling it reduces the file size further. (For one example file: spc2it raw - 400 KB, opened/compat saved by OpenMPT - 250 KB, with compression - 200 KB.)

> I also wonder how successful would it be if the buffering etc are implemented at library's user and the library is interacted with `xmp_load_module_from_callbacks`

That's a good question. I added better benchmarking tooling to my client application, and made the following benchmarks (all tracks are IT, each measurement is average of 3):

<details>

<summary>**Main branch libxmp (no hio buffer)**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | 6.3s  | .71s | .54s |
| Roaming  | 1.4s  | .15s | .12s |
| Mansion  | 5.9s  | .68s | .51s |
| Desert  | .49s  | .070s | .067s |

</details>

<details>

<summary>**libxmp with 4kb hio buffer**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | 3.1s  | .39s | .39s |
| Roaming  | .76s  | .095s | .094s |
| Mansion  | 2.7s  | .37s | .37s |
| Desert  | .66s  | .067s | .068s |

</details>

<details>

<summary>**libxmp with 64kb hio buffer**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | .65s  | .40s | .40s |
| Roaming  | .15s  | .099s | .099s |
| Mansion  | .59s  | .38s | .38s |
| Desert  | .31s  | .081s | .081s |

</details>

<details>

<summary>**libmodplug (from mem)**:</summary>

| Track  | mem |
| ------------- | ------------- |
| Overworld  | .12s  |
| Roaming  | .042s  |
| Mansion  | .11s   |
| Desert  | .043s   |

</details>

Conclusions:
* libmodplug seems to be faster at loading IT files in general.
* Surprisingly, 4kb HIO buffering is significantly faster than no buffering when loading from memory and memory-backed callbacks.
* 64kb HIO buffering is faster than 4kb HIO buffering from disk, but not for memory or memory-backed callbacks.
* Memory-backed callbacks are slower than memory without HIO buffering, but the same as memory with HIO buffering.

AliceLR:
Performance improvements from buffering IT patterns and compressed samples in the loader are looking good: on my PC, it cuts the time it takes to load the entirety of the Modland IT directory *from memory* by almost half. The actual improvement is going to depend a lot on the modules being loaded, though. Once I've tidied it a little I'll open the PR. https://github.com/libxmp/libxmp/compare/master...AliceLR:add-it-64k-buffer

ds-sloth:
Very exciting! I've benchmarked your changes on 3DS, with and without HIO buffering.

<details>

<summary>**Main branch libxmp (no hio buffer, same from last post)**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | 6.3s  | .71s | .54s |
| Roaming  | 1.4s  | .15s | .12s |
| Mansion  | 5.9s  | .68s | .51s |
| Desert  | .49s  | .070s | .067s |

</details>

<details>

<summary>**add-it-64k-buffer branch libxmp (no hio buffer)**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | 3.9s  | .27s | .27s |
| Roaming  | .98s  | .073s | .072s |
| Mansion  | 3.5s  | .25s | .25s |
| Desert  | .47s  | .064s | .064s |

</details>

<details>

<summary>**add-it-64k-buffer branch libxmp (with 64kb hio buffer)**:</summary>

| Track  | SDL_RWops (file) | SDL_RWops (mem) | mem |
| ------------- | ------------- | ------------- | ------------- |
| Overworld  | .65s  | .29s | .29s |
| Roaming  | .16s  | .080s | .081s |
| Mansion  | .59s  | .27s | .27s |
| Desert  | .31s  | .080s | .079s |

</details>

This benchmark tells me that global HIO buffering is not an ideal long-term solution, since optimizations targeted at the IT loader result in a better performance gain when loading from memory. So, feel free to close this PR; we can continue any discussion when you make your PR for your branch.

The benchmarks have also fully convinced me to load to memory before passing my tracks to libxmp. I guess there is simply too much random access required when loading these file formats. I think you may want to update your documentation to strongly discourage people from using file-backed I/O.

AliceLR:
> This benchmark tells me that global HIO buffering is not an ideal long-term solution, since optimizations targeted at the IT loader result in a better performance gain when loading from memory.

There are some formats where HIO buffering might be the only reasonable solution: of the formats listed in this issue, STM and some STX modules use the same type of packed patterns as S3M and IT without a length word to help with buffering.

sezero:
Should we close this?

AliceLR:
> Should we close this?

No, I just haven't got to reviewing this yet, sorry. :(

PR #691, opened 2023-10-20
scan.c: optimize for common case of no effects

ds-sloth:
This is a very simple PR. While profiling `scan_module`, I noticed that the effects slots for all events were compared to all possible FX, even though the vast majority of effects slots are empty.

The only change in the PR is to skip these checks if both effects slots for an event are empty.

In my benchmarking, this yields about a 10% load time improvement *in general*.


AliceLR (approved):
Oh, wow. Nice catch!

PR #694, opened 2023-10-23
Faster IT loading by buffering pattern, sample, and comment reads.

AliceLR:
The IT format stores 16-bit lengths for patterns and compressed sample blocks, which allows both to be loaded into a buffer to be read from memory. This patch adds a 64k temporary buffer to the IT loader to reduce load times for this format.

The average improvement in my loose measurements is about 45.5% (measured by loading every file in the Modland IT directory once with xmpchk: Windows 10 x86_64, all files buffered by the kernel, all files read into memory and loaded with `xmp_load_module_from_memory`). This is going to vary based on the structure of any given file.

|       | master (16fc5c89) | Patch      |
|-------|-------------------|------------|
| Run 1 | 3m34.269s         | 1m56.595s  |
| Run 2 | 3m34.537s         | 1m56.572s  |
| Run 3 | 3m34.021s         | 1m57.067s  |
| Avg.  | 3m34.276s         | 1m56.745s  |

Further testing in Fedora and some additional testing by ds-sloth on the 3DS verifies these performance improvements.


sezero (approved):
Should we merge or do you need additional comments from @ds-sloth ?

ds-sloth:
Thanks for the ping. I skimmed the the updated branch and there are no significant differences between this version and the version I tested locally and on 3DS (looks like mostly safety checks for the temp buffer).

If you're interested in further marginal speedups (around ~2% of IT load time each), I found a few based on local profiling. I can share them in a separate PR or multiple PRs, or quickly explain them here if the two of you would like to evaluate whether you're interested.

AliceLR:
> Should we merge or do you need additional comments from @ds-sloth ?

Yes. We can open further pull requests if there are other improvements.

> If you're interested in further marginal speedups (around ~2% of IT load time each), I found a few based on local profiling. I can share them in a separate PR or multiple PRs, or quickly explain them here if the two of you would like to evaluate whether you're interested.

Since 2% isn't very much, a quick list might be best for now.  I don't think e.g. buffering the pointer lists manually would be very substantial so I skipped them.

sezero:
> Yes. We can open further pull requests if there are other improvements.

OK then, merging.

PR #695, opened 2023-10-24
Extra optimizations for IT load

ds-sloth:
In discussion of @AliceLR's buffering for IT load I mentioned that I had noticed a few further possible optimizations. I'm sharing them here; there's no pressure to review them quickly (or at all).

On my own files, these seem to yield around a 7% total speedup when loading files and a 14% reduction in instructions spent in `it_load` (profiled with callgrind). I'm not certain what the Modland IT directory is but it would be nice to use that instead of (or in addition to) my own files for benchmarking.

A brief description of the optimizations in the commits I've shared here:
- cache channel event bases - at the top of a pattern load, get row 0's event mapping for each channel. Use this instead of the more expensive trinary `EVENT` macro.
- skip empty mask nybbles - when checking which bits are set in a mask, skip to the second nybble if the first is empty; continue to the next row if the second nybble is empty. On my files this was faster than checking and skipping over the entire byte, but this should be benchmarked with a bigger fileset.
- speed up first pass - the first pass over the IT is just checking which channels are active. It doesn't actually consume any of the information requested by masks, but it needs to advance the file pointer according to the less significant nybble of the mask. I index into a nybble-to-advancement mapping instead of using control flow to advance the file pointer. (This is less obviously interpretable to a casual reader.)

The performance improvement of each change for reduction in instructions spent in `it_load` (roughly half these for the total load time improvement):
- cache channel event bases - 4%
- skip empty mask nybbles - 7%
- speed up first pass - 3%

### DIFF
--- a/src/loaders/it_load.c
+++ b/src/loaders/it_load.c
@@ -1353,6 +1353,9 @@ static int it_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		}
 		pos = patbuf;
 
+		/* How do bits of mask[c] advance the file pointer? Bits 1, 2, and 4 advance 1 byte; bit 8 advances 2. */
+		static const uint8 adv[16] = {0, 1, 1, 2, 1, 2, 2, 3, 2, 3, 3, 4, 3, 4, 4, 5};
+
 		row = 0;
 		while (row < num_rows && --pat_len >= 0) {
 			int b = *(pos++);
@@ -1372,22 +1375,8 @@ static int it_load(struct module_data *m, HIO_HANDLE *f, const int start)
 				pat_len--;
 			}
 
-			if (mask[c] & 0x01) {
-				pos++;
-				pat_len--;
-			}
-			if (mask[c] & 0x02) {
-				pos++;
-				pat_len--;
-			}
-			if (mask[c] & 0x04) {
-				pos++;
-				pat_len--;
-			}
-			if (mask[c] & 0x08) {
-				pos += 2;
-				pat_len -= 2;
-			}
+			pos += adv[mask[c] & 0x0f];
+			pat_len -= adv[mask[c] & 0x0f];
 		}
 	}
 
--- a/src/loaders/it_load.c
+++ b/src/loaders/it_load.c
@@ -942,7 +942,7 @@ static int load_it_pattern(struct module_data *m, int i, int new_fx,
 			   uint8 *patbuf, HIO_HANDLE *f)
 {
 	struct xmp_module *mod = &m->mod;
-	struct xmp_event *event, dummy, lastevent[L_CHANNELS];
+	struct xmp_event *event, dummy, lastevent[L_CHANNELS], *event_base[L_CHANNELS];
 	uint8 mask[L_CHANNELS];
 	uint8 last_fxp[64];
 	uint8 *pos;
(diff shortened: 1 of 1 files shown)


src/loaders/it_load.c
```diff
@@ -1339,6 +1353,9 @@ static int it_load(struct module_data *m, HIO_HANDLE *f, const int start)
 		}
 		pos = patbuf;
 
+		/* How do bits of mask[c] advance the file pointer? Bits 1, 2, and 4 advance 1 byte; bit 8 advances 2. */
+		static const uint8 adv[16] = {0, 1, 1, 2, 1, 2, 2, 3, 2, 3, 3, 4, 3, 4, 4, 5};
```
sezero, line 1357: moving this to top should cure build with c90 compilers

src/loaders/it_load.c
```diff
@@ -1009,17 +1018,17 @@ static int load_it_pattern(struct module_data *m, int i, int new_fx,
...
+					b = XMP_KEY_OFF;
+					break;
+				case 0xfe:	/* cut */
+					b = XMP_KEY_CUT;
+					break;
+				default:
+					if (b > 119) {	/* fade */
+						b = XMP_KEY_FADE;
+					} else {
+						b++;	/* note */
```
sezero, line 1031: unwelcome indentation / whitespace changes

src/loaders/it_load.c
```diff
@@ -963,6 +963,11 @@ static int load_it_pattern(struct module_data *m, int i, int new_fx,
 		return -1;
 	}
 
+	/* fill event bases for this pattern */
+	for (c = 0; c < L_CHANNELS && c < m->mod.chn; c++) {
+		event_base[c] = m->mod.xxt[TRACK_NUM((i),(c))]->event;
+	}
+
```
AliceLR, line 970: This looks fragile and non-futureproof. I recommend keeping `EVENT()` as-is.

src/loaders/it_load.c
AliceLR: There isn't a style guide anywhere, but yes, in libxmp `switch` and `case` go on the same indentation level.

src/loaders/it_load.c
AliceLR: Actually, a followup here: do you see similar performance improvements by changing the function declaration to this?
```C
static int load_it_pattern(struct module_data * LIBXMP_RESTRICT m, int i, int new_fx,
			   HIO_HANDLE * LIBXMP_RESTRICT f)
```

AliceLR:
> On my own files, these seem to yield around a 7% total speedup when loading files and a 14% reduction in instructions spent in `it_load` (profiled with callgrind). I'm not certain what the Modland IT directory is but it would be nice to use that instead of (or in addition to) my own files for benchmarking.

It is [here](https://ftp.modland.com/pub/modules/Impulsetracker)—as the URL implies, the whole directory can be pulled via FTP. It is over 20GiB in size, so make sure you have that much RAM free for buffering :). Typically, I do something like this:
```sh
# Make xmpchk, the CLI frontend for libxmp's libFuzzer test routine.
(cd ~/libxmp/test-dev && make utilities)

# Make a file list for xmpchk (to bypass argument limits).
find . -iname "*.it" >filelist.txt

# First pass - force the operating system preload the directory contents to buffer.
cat filelist.txt | ~/libxmp/test-dev/xmpchk - 2>/dev/null

# Now do the above, but with bash time, /usr/bin/time, an attaching profiler, etc...
```

AliceLR:
Pushing revising this patch off to 4.6.2 since nothing is technically broken here.

AliceLR:
Spent probably far too long trying to replicate these performance improvements on x86-64 and RISC-V (ARM TBA) but these changes and some other changes I've tried are negligible or, in some cases, worse. May be worth another look later but I'm not rebasing these changes for 4.6.3.

PR #704, opened 2023-12-03
16-bit check fixed in MultiTracker loader

neumatho:
When checking for a sample is 16-bit in the MultiTracker loader, the wrong attr field was used.


sezero:
PING @AliceLR

PR #705, opened 2023-12-03
Fixed the finetune in MultiTracker loader

neumatho:
The finetune need to be shifted in the MultiTracker loader. If not, finetuned samples was played out of tune. This can especially be heard in module L'Esperanza.mtm from position 39 (the flute).

[L'Esperanza.zip](https://github.com/libxmp/libxmp/files/13539499/L.Esperanza.zip)


sezero:
PING @AliceLR

AliceLR (approved):
There's something wrong with MultiTracker's implementation of finetune that I need to look into further. Since it's (nominally) ProTracker finetune, this should be correct though.

PR #707, opened 2024-01-07
Atari Octalyser pattern loop fix

neumatho:
Atari Octalyser seems to have the pattern loop arguments as global instead of channel separated. This will fix that, so 8er-mod module will play correctly.

I have attached the modules mentioned in my comments.

[Octalyser.zip](https://github.com/libxmp/libxmp/files/13852599/Octalyser.zip)

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -266,6 +266,7 @@ int libxmp_snprintf (char *, size_t, const char *, ...) LIBXMP_ATTRIB_PRINTF(3,4
 #define QUIRK_S3MLOOP	(1 << 0)	/* S3M loop mode */
 #define QUIRK_ENVFADE	(1 << 1)	/* Fade at end of envelope */
 #define QUIRK_PROTRACK	(1 << 2)	/* Use Protracker-specific quirks */
+#define QUIRK_OCTALYSERLOOP	(1 << 3)	/* Octalyser (Atari) loop mode */
 #define QUIRK_ST3BUGS	(1 << 4)	/* Scream Tracker 3 bug compatibility */
 #define QUIRK_FINEFX	(1 << 5)	/* Enable 0xf/0xe for fine effects */
 #define QUIRK_VSALL	(1 << 6)	/* Volume slides in all frames */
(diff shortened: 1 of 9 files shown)


sezero:
PING @AliceLR

AliceLR:
I haven't had a chance to look into this yet, sorry. I need to check other formats re: the specificity of this patch and whether or not this change needs one of our very few remaining quirk bits.

sagamusix:
Scream Tracker 3 also only has one global pattern loop counter and no per-channel pattern loops, so at the very least this code could be extended to cover S3M as well.

sezero:
Any progress on this?

sezero:
Any progress ?

AliceLR:
DTM has this too, and the loop point is not reset between positions.

AliceLR:
> Dammed Illusion

Not clear why the double `E60`/`E62` works in this track, because when I do the same thing in my own test modules, Octalyser gets stuck in an infinite loop. Octalyser will also do strange things like ignoring `E60` when it follows `E61` in spite of the otherwise global loop behavior. This needs more investigation.

AliceLR:
It turns out many trackers have multiple different Pattern Loop quirks, which sometimes combine in strange ways or sometimes appear/disappear between tracker versions. From the few trackers I've tested alone I found at least 9 independent quirks that need or may need support.

Candidate replacement patch which creates a new quirks field *exclusively* for Pattern Loop (still testing): https://github.com/libxmp/libxmp/compare/master...AliceLR:add-loop-mode-quirks

AliceLR:
Update: the replacement patch is essentially implemented, but I keep finding UNRELATED things that need fixing while I am making the test modules (which I am only about half done with). In particular, one of the format loaders (Liquid Tracker NO) contains a several errors and needs an overhaul. I am going to try to split off some of these unrelated patches tonight or tomorrow.

AliceLR:
Predictably (as it was implemented off of only two modules), the other Liquid Tracker loader is broken too. It'll be a little bit longer while I fix that.

PR #708, opened 2024-01-17
Add module size in struct xmp_module_info

debrouxl:
Here's a patch I've been carrying locally a while, as the underlying changes for an xmp-cli change I'll create a PR for as well.

This is the kind of change which warrants both a version bump, so that libxmp clients can depend on the new version, and sadly, a recompilation of libxmp clients, at least on 32-bit platforms where the size of the struct changes. On 64-bit platforms, the new field could technically be stored in the padding bytes between integers and pointers in xmp_module_info.

I did think about using a `long` type for this new field, as `xmp_load_module_from_memory()` and `xmp_load_module_from_file()` take a `long` size (though the latter doesn't use its size argument). However, `src/common.h::module_data::size` has `int` type anyway.

IME on other projects, version bumps are usually a copy-paste-modify-check job, so I _can_ make a patch (attempting to check for locations which could have appeared since the previous version bump), but maintainers are usually the ones doing version bumps, so I'll let you do it, unless you want me to :)

### DIFF
--- a/include/xmp.h
+++ b/include/xmp.h
@@ -300,6 +300,7 @@ struct xmp_module_info {
 	char *comment;			/* Comment text, if any */
 	int num_sequences;		/* Number of valid sequences */
 	struct xmp_sequence *seq_data;	/* Pointer to sequence data */
+	int size;			/* Module size */
 };
 
 struct xmp_channel_info {
--- a/src/load.c
+++ b/src/load.c
@@ -394,6 +394,7 @@ int xmp_load_module(xmp_context opaque, const char *path)
 	ctx->m.filename = NULL;
 	ctx->m.dirname = NULL;
 	ctx->m.basename = NULL;
+	ctx->m.size = hio_size(h);
 #endif
 
 	ret = load_module(opaque, h);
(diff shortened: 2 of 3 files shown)


sezero:
IMO, this should wait a bit.

@AliceLR ?

sagamusix:
This addition seems dubious to me. The library caller should already have access to this information. It looks like taking the library as a shortcut for something it shouldn't be doing. If this gets added, at the very least, the property name should clearly indicate that this is the file size. This is an important distinction to make because the module size itself could be technically smaller if the file is padded with garbage (as it is sometimes the case with bad rips from demos or games). Being able to compute the actual module size and providing it to the library caller could be much more useful, but is also a lot more complex (every loader would have to keep track of the theoretical module size, or at least the highest-accessed byte position within the file would have to be tracked).

sezero:
I actually don't like this patch at all, either. @AliceLR: Close this as wont-fix?

AliceLR:
> I actually don't like this patch at all, either. @AliceLR: Close this as wont-fix?

Works for me.

Issue #710, opened 2024-02-11
Fade 2 Grey (Visage) plays incorrectly


LossyDragon:
I'm not sure when this module started to play incorrectly, but I know from libxmp 4.4.1 it sounds *mostly* fine. 

Building xmp-cli it seems that channel 14 and 15 are trying to play the notes, but goes quiet as notes become louder and higher pitched. Pattern 7 displays this well and doesnt sound nearly as well as openMPT can play it. 

Building from `master` (and 4.6.0 release) and playing through the SDL example sounds far worse with channels 14/15 becoming mute until the highest volume of those channels are playing (?)

Module in question: https://modarchive.org/index.php?request=view_by_moduleid&query=134790

sezero:
@AliceLR ?

AliceLR:
Just to double check: you built off of the *latest* HEAD? When I run `xmp fade_2_grey_visage.xm -s8` pattern 7 sounds more or less the same as OpenMPT.

I'm ~90% certain this was fixed by 9e25a5c9. When I check out its parent commit, it sounds like what you're describing.

LossyDragon:
I believe so. Before making my report I had a copy from Feb 3rd. But I did grab a fresh copy just to ensure a clean build. 

Below are some videos, I should note that sequence 3 and 4 displays the issue pretty good. Ignore the slight audio popping from the android emulator. But the issue persists on a physical device. 

The 2nd video is through WSL Ubuntu, with its own copy to build and test. 

https://github.com/libxmp/libxmp/assets/4994318/12337f4c-0f7f-4f29-9fc4-329e90425e68

https://github.com/libxmp/libxmp/assets/4994318/10147122-c851-4285-b3a2-a6383dc944f8

AS does throw a few unused warnings, just for the sake of noting it. 
```
ask :app:configureCMakeRelease[arm64-v8a]
C/C++: -- Check for working linker: C:/Users/Lossy/AppData/Local/Android/Sdk/ndk/25.1.8937393/toolchains/llvm/prebuilt/windows-x86_64/bin/ld.lld.exe

> Task :app:mergeLibDexDebug

> Task :app:buildCMakeRelease[arm64-v8a]
C/C++: ninja: Entering directory `C:\Users\Lossy\AndroidStudioProjects\xmp-android\app\.cxx\Release\4k3p1j6x\arm64-v8a'
[...]

PR #712, opened 2024-02-12
Fix XM envelope sustain points that exist on a zero-length loop.

AliceLR:
An old bugfix(?) in the XM and generic envelope handlers was disabling the sustain point when it exists on the end of the envelope loop. This doesn't really line up with reality and, in the case of XM, was definitely wrong.

[fade_2_grey_visage.xm](https://modarchive.org/index.php?request=view_by_moduleid&query=134790) contains an instrument (3) with a sustain point on the same envelope point as the loop start and loop end. In XM, loops where the start and end are on the same point are ignored, but the sustain point in this case is supposed to work. The module relies on the sustain loop working in this situation in positions 3 and 4 (channels 8 and 9).

This bug was previously masked because XM loops with the start and the end on the same point were NOT being ignored (fixed in #678).

See issue #710.

### DIFF
--- a/src/player.c
+++ b/src/player.c
@@ -137,6 +137,9 @@ static int update_envelope_generic(struct xmp_envelope *env, int x, int release)
 	 * envelope loop end and the key is released, FT2 escapes the loop
 	 * while IT runs another iteration. (See EnvLoops.xm in the OpenMPT
 	 * test cases.)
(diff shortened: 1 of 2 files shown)


sezero (approved):

AliceLR:
> Close #710 ?

I think so. I took another look at the AS warnings and they don't look serious.

PR #713, opened 2024-02-28
exclude_match: add HTML files

mistydemeo:
A few ZIPs I have from an artist include HTML READMEs alongside their .diz files, which is preventing libxmp from detecting them properly.


sezero:
I think you should include the uppercase versions too. Other than that, looks good to me. @AliceLR?

mistydemeo:
OK, added the uppercase versions!

sezero (approved):

AliceLR (approved):

PR #715, opened 2024-03-18
Fix config version variables.

NoSuck:
I am by no means an autoconf/workingset master, but...

```config
VERSION_MAJOR	= @libxmp_VERSION_MAJOR@
VERSION_MINOR	= @libxmp_VERSION_MINOR@
VERSION_RELEASE	= @libxmp_VERSION_PATCH@

VERSION		= @libxmp_VERSION@
```

This was producing library links literally named like the following:

```
libxmp.so.@libxmp_VERSION_MAJOR@
libxmp.so.@libxmp_VERSION@
```

Now:

```
libxmp.so.4
libxmp.so.4.6.0
```

```sh
$ uname -srvmo
Linux 6.7.7-100.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Fri Mar  1 16:51:49 UTC 2024 x86_64 GNU/Linux
$ autoconf --version | head -n 1
autoconf (GNU Autoconf) 2.71
```

Am I missing something?

https://www.gnu.org/software/autoconf/manual/autoconf-2.64/html_node/Setting-Output-Variables.html

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,6 +1,6 @@
-m4_define([libxmp_VERSION_MAJOR_m4], 4)
-m4_define([libxmp_VERSION_MINOR_m4], 6)
-m4_define([libxmp_VERSION_PATCH_m4], 0)
+m4_define([libxmp_VERSION_MAJOR_m4], [4])
+m4_define([libxmp_VERSION_MINOR_m4], [6])
+m4_define([libxmp_VERSION_PATCH_m4], [0])
 m4_define([libxmp_VERSION_m4], libxmp_VERSION_MAJOR_m4[.]libxmp_VERSION_MINOR_m4[.]libxmp_VERSION_PATCH_m4)
 
 AC_PREREQ(2.60)
--- a/lite/configure.ac
+++ b/lite/configure.ac
@@ -1,6 +1,6 @@
-m4_define([libxmplite_VERSION_MAJOR_m4], 4)
-m4_define([libxmplite_VERSION_MINOR_m4], 6)
-m4_define([libxmplite_VERSION_PATCH_m4], 0)
+m4_define([libxmplite_VERSION_MAJOR_m4], [4])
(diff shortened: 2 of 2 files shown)


sezero:
It has been working here just fine, tested with autoconf 2.63, 2.69, 2.71, 2.72.

However, I am not opposed to the patch provided that you apply the same to lite/configure.ac.

And, why is the fuzzer stuff is failing configure in CI?

CC: @AliceLR

AliceLR:
> And, why is the fuzzer stuff is failing configure in CI?

No clue, I'll look into it tomorrow.

AliceLR:
Using Fedora 39 and the stock Autoconf 2.71 and GNU M4 1.4.19, I can't reproduce the original bug in my environment. As far as I can understand it, [in most cases it should be the same with and without the quotes](https://ftp.gnu.org/pub/pub/pub/old-gnu/Manuals/autoconf-2.57/html_chapter/autoconf_8.html#SEC92), so I'm not sure what's breaking here or why this change fixes it.

In your `configure` built without the quotes, are `libxmp_VERSION_MAJOR` et al. even defined?

And re: the sanitizer and fuzzer builds, I have no clue what broke yet. I can still build my fuzzers just fine locally. Maybe LLVM split a package...?

AliceLR (approved):
Sanitizer and fuzzer failures are unrelated and due to incompatible Linux and LLVM versions in the 20240310.1.0 Github runner image. Everything seems to be running on the fixed image now, so I've run the tests again.

Since that was the only "issue", this patch looks good to me.

sezero:
> Since that was the only "issue", this patch looks good to me. Holding on merging re: the lite/configure.ac request though.

Yes.

@NoSuck: Please apply the same change to lite/configure.ac too, then I'm OK with this patch and we can merge.

sezero:
I amended the lite/configure.ac change myself, simplified the commit message, and merged.

PR #716, opened 2024-03-20
Github Actions sanitizer crash testing.

AliceLR:


AliceLR:
@sezero I *think* this stupid bug is fixed, but I can't say 100% for sure :(

sezero:
How is `LD=lld` helping? (`lld` by itself shouldn't actually work either, it should be `lld.ld` for e.g. to imitate GNU ld.)

AliceLR:
> How is `LD=lld` helping? (`lld` by itself shouldn't actually work either, it should be `lld.ld` for e.g. to imitate GNU ld.)

I'm not entirely sure. The MemorySanitizer executables want `-fpie`/`-pie` (harder to make work) but with the extra `LD=lld` and `-fpic` it seems to have fixed itself. AddressSanitizer executables would also periodically segfault without both. I don't really trust this fix.

sezero:
I see that `LD=lld.ld` is failing.

Does it help if you drop `LD=...` export and configure using `--without-gnu-ld` ?

AliceLR:
> Does it help if you drop `LD=...` export and configure using `--without-gnu-ld` ?

Trying now..

sezero:
> > Does it help if you drop `LD=...` export and configure using `--without-gnu-ld` ?
> 
> Trying now..

If GNU ld is detected, it adds `-DUSE_VERSIONED_SYMBOLS=1` to CFLAGS for shared builds and links with `-Wl,--version-script,libxmp.map` -- that's the only use we have for `LD` variable

AliceLR:
This seems to be 100% the fault of the runner image: https://github.com/actions/runner-images/issues/9491

I'll revert most of the patch and run the tests again...

sezero:
OK great! No more crashes.

PR #720, opened 2024-04-26
CI, macOS: run 'brew install autoconf'

sezero:
To fix `autoconf: command not found` errors.

From CI logs:
```
Run brew install autoconf cmake
Warning: Treating cmake as a formula. For the cask, use homebrew/cask/cmake or specify the `--cask` flag.
Warning: cmake 3.29.2 is already installed and up-to-date.
To reinstall 3.29.2, run:
  brew reinstall cmake
==> Downloading https://ghcr.io/v2/homebrew/core/autoconf/manifests/2.72
==> Fetching dependencies for autoconf: m4
==> Downloading https://ghcr.io/v2/homebrew/core/m4/manifests/1.4.19
==> Fetching m4
==> Downloading https://ghcr.io/v2/homebrew/core/m4/blobs/sha256:f42d89db519a07d67bcaead6c8dfb2da45e8266bebb996dd8b3f19b1ca13b8a0
==> Fetching autoconf
==> Downloading https://ghcr.io/v2/homebrew/core/autoconf/blobs/sha256:bb39057e87dd9cb940bee15ff5b5172952a0350e59b14a6f55b8006a07813186
==> Installing dependencies for autoconf: m4
==> Installing autoconf dependency: m4
==> Downloading https://ghcr.io/v2/homebrew/core/m4/manifests/1.4.19
Already downloaded: /Users/runner/Library/Caches/Homebrew/downloads/5b2a7f715487b7377e409e8ca58569040cd89f33859f691210c58d94410fd33b--m4-1.4.19.bottle_manifest.json
==> Pouring m4--1.4.19.arm64_sonoma.bottle.tar.gz
🍺  /opt/homebrew/Cellar/m4/1.4.19: 13 files, 726.2KB
==> Installing autoconf
==> Pouring autoconf--2.72.arm64_sonoma.bottle.tar.gz
🍺  /opt/homebrew/Cellar/autoconf/2.72: 71 files, 3.6MB
```

Issue #676, opened 2023-08-24
make it possible to build the lite version from git tree


sezero:
This has been noted multiple times.  The lite release package will stay
the same (see #670), but people want to build the lite version from the
git tree:  See, e.g., #596, #620, #564, #502, maybe some other tickets.

sezero:
The cmake configuration option should be easy.

Autotools side may require some work.

Standalone makefiles might require some surgery.

PR #721, opened 2024-04-27
support for building lite version from git or release dist tarball

sezero:
cmake, autotools and standalone makefiles are supported:
- cmake: enable BUILD_LITE option (-DBUILD_LITE=ON on the cmdline)
- autotools: use --enable-lite configure switch
- watcom makefiles: wmake -f Makefile.os2 lite (for e.g.)
- msvc makefile: nmake -f Makefile.vc lite
- android: ndk-build builds both libxmp and libxmp-lite

Please review and test. Things of note: (I tested these myself, but still..)
- full library builds shouldn't be affected
- distribution tarballs (from `make dist`) should be able to build lite properly
- lite builds from full tree should be identical to lite tarball builds

CC: @madebr, @ccawley2011, @cmatsuoka , @Wohlstand


sezero:
P.S.: The macOS CI failure is irrelevant will be fixed by #720

sezero:
codecov crapola is taking 6h and them timing out. sigh...

Wohlstand:
A while ago when I made my CMake build, I managed to make a full featured lite build from the same build (I.e. build both lite and normal versions from the same build). But maintainers told me to do separated builds fro lite and main as there are different tarballs released.

sezero:
> A while ago when I made my CMake build, I managed to make a full featured lite build from the same build (I.e. build both lite and normal versions from the same build). But maintainers told me to do separated builds fro lite and main as there are different tarballs released.

I know. And this is supposed to rectify that not just for cmake but for autotools and for standalone makefiles as well, as it has been asked / requested multiple times (see e.g. https://github.com/libxmp/libxmp/issues/676)

Now, are there any issues with this patch? Or does anyone have any improvements to it?

Wohlstand:
I going to take a look as soon as I finish my current main work task (today at me is a workday because of holidays, one days off for been swapped with the monday)

sezero:
Now supports disabling it format in lite builds from the full tree as of 909a622

sezero:
Now supports lite library builds from full tree in Android.mk as of 9d11f756

sezero:
> Now supports lite library builds from full tree in Android.mk as of [9d11f75](https://github.com/libxmp/libxmp/commit/9d11f756bd40c5bd946f90c25b8b5c901d0eff39)

Noticed that the Android.mk in actual lite dist package outputs the library as `libxmp` instead of `libxmp-lite`: should we change that?

sezero:
Pushed tiny fixes found during this patchset to master. Rebased this PR to current master, squashed and force-pushed.

sezero:
Rebased to current git master again and force-pushed.

AliceLR:
The diff seems fine from a glance—will test and get back to you.

Are the stub source files actually necessary? This seems like something that could also be solved by overriding `.c.o` for lite objects (not sure about nmake and Watcom though...):

```Makefile
src/.lite/%.o: src/%.c
	@CMD='$(CC) $(CPPFLAGS) $(CFLAGS_STATIC) $(CFLAGS) -o $@ $<'; \
	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo CC $@ ; fi; \
	eval $$CMD
```

sezero:
> The diff seems fine from a glance—will test and get back to you.

Thanks. Looking forward to hearing results.

> Are the stub source files actually necessary?

They really make life easier, especially with standalone makefiles

> This seems like something that could also be solved by overriding `.c.o` for lite objects (not sure about nmake and Watcom though...):
> 
> ```makefile
> src/.lite/%.o: src/%.c
> 	@CMD='$(CC) $(CPPFLAGS) $(CFLAGS_STATIC) $(CFLAGS) -o $@ $<'; \
> 	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo CC $@ ; fi; \
> 	eval $$CMD
> ```

That solution looks like GNU-make specific, and no I don't know how well our 
makefiles behave when used with e.g. bsd versions, s I'd like to have the stubs
if it isn't an issue.

As for standalone makefiles, I don't know hdw doable it is w/o the stubs.

AliceLR:
Both libraries build fine for me, with and without IT, in MINGW64, MINGW32, and UCRT64. MSYS builds fine but can't find `-lxmp` when linking libxmp-test (probably unrelated to this patch).

* Since libxmp-lite is now integrated into the main buildsystems, the libxmp-lite test should also be integrated into `make check` when lite is enabled. Originally, I *think* the lite test had an XM so that it would work with IT disabled, but that might have been lost at some point (or I am wrong). (Also, the test will need MD5s generated for both IEEE-754 and x87 floating point :(.)
* The regression tests require a full build, so no change is required for them.

> That solution looks like GNU-make specific, and no I don't know how well our
makefiles behave when used with e.g. bsd versions, so I'd like to have the stubs
if it isn't an issue.

IIRC gmake is required to build libxmp for BSDs, will fact check myself in a bit.

sezero:
> Both libraries build fine for me, with and without IT, in MINGW64, MINGW32, and UCRT64. MSYS builds fine but can't find `-lxmp` when linking libxmp-test (probably unrelated to this patch).

That's great

> Since libxmp-lite is now integrated into the main buildsystems, the libxmp-lite test should also be integrated into `make check` when lite is enabled.
> (Also, the test will need MD5s generated for both IEEE-754 and x87 floating point :(.)

Are you sure? The only difference between full and lite `make check` is one uses test.itz and the other uses test.it which is simply uncompressed test.itz: The result will always be the same, and there is no need for new md5 generation.

I cooked something quick-and-dirty for autotools for now (even for msvc and watcom makefiles but they are more ugly..) Should I really push?

> Originally, I _think_ the lite test had an XM so that it would work with IT disabled, but that might have been lost at some point (or I am wrong).

Can't find it in tags: As far as I can see 4.1.4 is the point where test and test-dev are separated. If there was something in between 4.1.3 and 4.1.4, I don't know (haven't looked.)

> The regression tests require a full build, so no change is required for them.

Indeed

> 
> > That solution looks like GNU-make specific, and no I don't know how well our
> > makefiles behave when used with e.g. bsd versions, so I'd like to have the stubs
> > if it isn't an issue.
> 
> IIRC gmake is required to build libxmp for BSDs, will fact check myself in a bit. edit: this is correct, BSD make fails entirely. Both libraries build and pass the main check in NetBSD 9.2. (One regression test fails but it's unrelated and I'm not sure why yet.)

OK.

sezero:
Force-pushed to fix missing dependency generation for lite objects
in autotools.

Questions remain answering are:

- library name for android ndk builds in lite distribution packages
  as noted above

- lite test addition really needed? (as I discussed above, and it
  will fail if --disable-it is used anyway? what do you think?

Off-topic: coverage upload is timing out. annoying...

AliceLR:
> * library name for android ndk builds in lite distribution packages
>   as noted above

IMO they should be different but I am not a user of the Android NDK builds (MegaZeux just builds its in-tree fork as part of its normal build process).

> * lite test addition really needed? (as I discussed above, and it
>   will fail if --disable-it is used anyway? what do you think?

I think it would be good to have a test that verifies the lite library works. As you've mentioned, because IT can be disabled, the lite test should be an XM. I will try to determine if I was just hallucinating that XM test...

sezero:
Feel free to push to my branch for any improvements.

> > * library name for android ndk builds in lite distribution packages
> >   as noted above
> 
> IMO they should be different but I am not a user of the Android NDK builds (MegaZeux just builds its in-tree fork as part of its normal build process for Android).

It was _probably_ an oversight. I can create a PR to change the name?

sezero:
Force-pushed again to fix android jni/* in lite distribution

sezero:
Force-pushed again to better handle the android ndk-build's
`LOCAL_MODULE_FILENAME should not contain extensions` issue

CMakeLists.txt
```diff
@@ -23,12 +24,14 @@ endif()
...
 option(BUILD_STATIC   "Build libXMP static library" ON)
 option(BUILD_SHARED   "Build libXMP shared library" ${BUILD_SHARED_DEFAULT})
+option(BUILD_LITE     "Build libXMP-lite library" OFF)
 if(NOT BUILD_STATIC AND NOT BUILD_SHARED)
     message(FATAL_ERROR "Both static and shared builds got disabled. You must enable at least one of them.")
 endif()
 
 option(LIBXMP_DISABLE_DEPACKERS     "Disable archive depackers" OFF)
 option(LIBXMP_DISABLE_PROWIZARD     "Disable ProWizard format loaders" OFF)
+option(LIBXMP_DISABLE_IT            "Disable IT format in libXMP-lite" OFF)
```
Wohlstand, line 34: Maybe add the flag that this depends on `BUILD_LITE`? I.e. it has no sense without the `BUILD_LITE=ON`

sezero: Hm, can change to cmake_dependent_option

sezero: Its description clearly states that it's for libxmp-lite, though

CMakeLists.txt
```diff
@@ -44,6 +47,14 @@ else()
     list(APPEND LIBXMP_SRC_LIST ${LIBXMP_SRC_LIST_PROWIZARD})
 endif()
 
+#lite-only defs
+list(APPEND LIBXMPLITE_DEFINES_PUBLIC -DLIBXMP_CORE_PLAYER)
+if(LIBXMP_DISABLE_IT)
+    list(APPEND LIBXMPLITE_DEFINES_PUBLIC -DLIBXMP_CORE_DISABLE_IT)
+endif()
```
madebr, line 54: These defines are exported to the cmake files, but are not present in the libxmp headers.
Shouldn't these be private instead?

sezero: I honestly don't know. I just followed the existing `LIBXMP_DEFINES_PUBLIC` in there. Maybe that's wrong too?

madebr: I think so: looks like these are implementation-only defines.
You need to rename the variable to `LIBXMP(LITE)_DEFINES_PRIVATE` and add them as `PRIVATE` compile definitions.
But make sure to keep adding `LIBXMP_STATIC` as `PUBLIC`.
Also, you don't need to prepend the defines with `-D`, when used with `target_compile_definitions`.

sezero: OK. I will force-push with the following applied: can you confirm? [cmake.patch.txt](https://github.com/libxmp/libxmp/files/15181914/cmake.patch.txt)

<details>

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 23c7bbe..1d3d7c3 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,9 +2,8 @@ cmake_minimum_required(VERSION 3.2...3.5)
 project(libxmp VERSION 4.6.0 LANGUAGES C)
 
 set(LIBXMP_DEFINES)
-set(LIBXMP_DEFINES_PUBLIC)
+set(LIBXMPLITE_DEFINES)
 set(LIBXMP_CFLAGS)
-set(LIBXMPLITE_DEFINES_PUBLIC)
 
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/libxmp-checks.cmake)
 include(GNUInstallDirs)
@@ -36,28 +35,28 @@ option(LIBXMP_DISABLE_IT            "Disable IT format in libXMP-lite" OFF)
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/libxmp-sources.cmake)
 
 if(LIBXMP_DISABLE_DEPACKERS)
-    list(APPEND LIBXMP_DEFINES_PUBLIC -DLIBXMP_NO_DEPACKERS)
+    list(APPEND LIBXMP_DEFINES LIBXMP_NO_DEPACKERS)
 else()
     list(APPEND LIBXMP_SRC_LIST ${LIBXMP_SRC_LIST_DEPACKERS})
 endif()
 
 if(LIBXMP_DISABLE_PROWIZARD)
-    list(APPEND LIBXMP_DEFINES_PUBLIC -DLIBXMP_NO_PROWIZARD)
+    list(APPEND LIBXMP_DEFINES LIBXMP_NO_PROWIZARD)
 else()
     list(APPEND LIBXMP_SRC_LIST ${LIBXMP_SRC_LIST_PROWIZARD})
 endif()
 
 #lite-only defs
-list(APPEND LIBXMPLITE_DEFINES_PUBLIC -DLIBXMP_CORE_PLAYER)
+list(APPEND LIBXMPLITE_DEFINES LIBXMP_CORE_PLAYER)
 if(LIBXMP_DISABLE_IT)
-    list(APPEND LIBXMPLITE_DEFINES_PUBLIC -DLIBXMP_CORE_DISABLE_IT)
+    list(APPEND LIBXMPLITE_DEFINES LIBXMP_CORE_DISABLE_IT)
 endif()
 #for libxmp-lite.pc
 set(libxmplite_VERSION ${libxmp_VERSION})
 
 if(MSVC)
-    set(LIBXMP_MSVC_DEFINES -D_USE_MATH_DEFINES)
-    list(APPEND LIBXMP_DEFINES ${LIBXMP_MSVC_DEFINES})
+    list(APPEND LIBXMP_DEFINES _USE_MATH_DEFINES)
+    list(APPEND LIBXMPLITE_DEFINES _USE_MATH_DEFINES)
 endif()
 
 
@@ -76,7 +75,7 @@ if(BUILD_STATIC)
 
     target_compile_options(xmp_static PRIVATE ${LIBXMP_CFLAGS})
     target_compile_definitions(xmp_static PRIVATE ${LIBXMP_DEFINES})
-    target_compile_definitions(xmp_static PUBLIC ${LIBXMP_DEFINES_PUBLIC} -DLIBXMP_STATIC)
+    target_compile_definitions(xmp_static PUBLIC LIBXMP_STATIC)
 
     if(LIBXMP_PIC)
         target_compile_options(xmp_static PRIVATE -fPIC)
@@ -101,8 +100,8 @@ if(BUILD_STATIC)
         endif()
 
         target_compile_options(xmp_lite_static PRIVATE ${LIBXMP_CFLAGS})
-        target_compile_definitions(xmp_lite_static PRIVATE ${LIBXMP_DEFINES})
-        target_compile_definitions(xmp_lite_static PUBLIC ${LIBXMPLITE_DEFINES_PUBLIC} -DLIBXMP_STATIC)
+        target_compile_definitions(xmp_lite_static PRIVATE ${LIBXMPLITE_DEFINES})
+        target_compile_definitions(xmp_lite_static PUBLIC LIBXMP_STATIC)
 
         if(LIBXMP_PIC)
             target_compile_options(xmp_lite_static PRIVATE -fPIC)
@@ -157,11 +156,10 @@ if(BUILD_SHARED)
 
     target_compile_definitions(xmp_shared PRIVATE ${LIBXMP_DEFINES})
     if(WIN32)
-        target_compile_definitions(xmp_shared PRIVATE -DBUILDING_DLL)
+        target_compile_definitions(xmp_shared PRIVATE BUILDING_DLL)
     endif()
 
     target_compile_options(xmp_shared PRIVATE ${LIBXMP_CFLAGS})
-    target_compile_definitions(xmp_shared PUBLIC ${LIBXMP_DEFINES_PUBLIC})
 
     target_include_directories(xmp_shared PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>")
     target_include_directories(xmp_shared PUBLIC "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>")
@@ -211,13 +209,12 @@ if(BUILD_SHARED)
             message(STATUS "Versioned symbols: DISABLED")
         endif()
 
-        target_compile_definitions(xmp_lite_shared PRIVATE ${LIBXMP_DEFINES})
+        target_compile_definitions(xmp_lite_shared PRIVATE ${LIBXMPLITE_DEFINES})
         if(WIN32)
-            target_compile_definitions(xmp_lite_shared PRIVATE -DBUILDING_DLL)
+            target_compile_definitions(xmp_lite_shared PRIVATE BUILDING_DLL)
         endif()
 
         target_compile_options(xmp_lite_shared PRIVATE ${LIBXMP_CFLAGS})
-        target_compile_definitions(xmp_lite_shared PUBLIC ${LIBXMPLITE_DEFINES_PUBLIC})
 
         target_include_directories(xmp_lite_shared PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>")
         target_include_directories(xmp_lite_shared PUBLIC "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/libxmp-lite>")
```

madebr: Looks good.

Before you make the change, these defines being public might be useful to users to know whether a feature is avaialble.
Perhaps that's why they were added to the public interface?

sezero: > Looks good.

OK (but waiting for the question you raised)

> Before you make the change, these defines being public might be useful to users to know whether a feature is avaialble. Perhaps that's why they were added to the public interface?

Don't know, _possibly_.. Maybe @Wohlstand can remember? They have been there since the first addition of cmake support from #415 (but they dos't look that very much useful to me...)

sezero: OK, force-pushed with cmake clean-ups.

sezero:
@AliceLR: I have done as much as I can with this. Is there anything else to take care of? (IMO, test-lite can be handled later with an xm or protracker test module addition, but I'll defer to your decision.)

AliceLR (approved):
If everything else here is done, I'm OK with it.

> (IMO, test-lite can be handled later with an xm or protracker test module addition, but I'll defer to your decision.)

That works for me.

CMakeLists.txt
madebr: How do the tests know whether a feature is enabled and to be tested?

sezero:
OK, patch is in.

Your libxmp-random patch now needs rebasing and adjusting for lite addition. Like:
```diff
diff --git a/src/lite/lite-rng.c b/src/lite/lite-rng.c
new file mode 100644
index 0000000..84e776a
--- /dev/null
+++ b/src/lite/lite-rng.c
@@ -0,0 +1,4 @@
+#ifndef LIBXMP_CORE_PLAYER
+#define LIBXMP_CORE_PLAYER
+#endif
+#include "../rng.c"
diff --git a/src/lite/Makefile b/src/lite/Makefile
index 93adaf6..4ccc1ed 100644
--- a/src/lite/Makefile
+++ b/src/lite/Makefile
@@ -3,7 +3,7 @@
 LITE		= lite-virtual.o lite-format.o lite-period.o lite-player.o lite-read_event.o \
 		  lite-misc.o lite-dataio.o lite-lfo.o lite-scan.o lite-control.o lite-filter.o \
 		  lite-effects.o lite-mixer.o lite-mix_all.o lite-load_helpers.o lite-load.o \
-		  lite-filetype.o lite-hio.o lite-smix.o lite-memio.o lite-win32.o \
+		  lite-filetype.o lite-hio.o lite-smix.o lite-memio.o lite-rng.o lite-win32.o \
 		  \
 		  lite-common.o lite-itsex.o lite-sample.o \
 		  lite-xm_load.o lite-mod_load.o lite-s3m_load.o lite-it_load.o
diff --git a/cmake/libxmp-sources.cmake b/cmake/libxmp-sources.cmake
index 875799e..20bcb08 100644
--- a/cmake/libxmp-sources.cmake
+++ b/cmake/libxmp-sources.cmake
@@ -202,6 +202,7 @@ set(LIBXMP_SRC_LIST_LITE
     src/lite/lite-hio.c
     src/lite/lite-smix.c
     src/lite/lite-memio.c
+    src/lite/lite-rng.c
     src/lite/lite-win32.c
     src/lite/lite-common.c
     src/lite/lite-itsex.c
diff --git a/Makefile.vc b/Makefile.vc
index 2f9578b..7bb5141 100644
--- a/Makefile.vc
+++ b/Makefile.vc
@@ -237,6 +237,7 @@ LITE_OBJS	= \
  src\lite\lite-hio.obj \
  src\lite\lite-smix.obj \
  src\lite\lite-memio.obj \
+ src\lite\lite-rng.obj \
  src\lite\lite-win32.obj \
  src\lite\lite-common.obj \
  src\lite\lite-itsex.obj \
diff --git a/watcom.mif b/watcom.mif
index 87e101b..b25d0e5 100644
--- a/watcom.mif
+++ b/watcom.mif
@@ -246,6 +246,7 @@ LITE_OBJS= &
  src/lite/lite-hio.obj &
  src/lite/lite-smix.obj &
  src/lite/lite-memio.obj &
+ src/lite/lite-rng.obj &
  src/lite/lite-win32.obj &
  src/lite/lite-common.obj &
  src/lite/lite-itsex.obj &
```

CMakeLists.txt
sezero: test-dev, unfortunately, expects a full-feature build of the library.

sezero:
I added test-lite thing as a new ticket, btw: https://github.com/libxmp/libxmp/issues/724

CMakeLists.txt
sezero: sigh.. test/test.c needs to know though: https://github.com/libxmp/libxmp/pull/725

madebr: An alternative solution is to create an interface library, to which you add the various compile definitions.
e.g.
```cmake
add_library(libxmp-build-options INTERFACE)
if(LIBXMP_DISABLE_DEPACKERS)
  target_compile_definitions(libxmp-build-options INTERFACE LIBXMP_NO_DEPACKERS)
endif()

target_link_libraries(libxmp_static PRIVATE $<BUILD_INTERFACE:libxmp-build-options>)
target_link_libraries(libxmp_shared PRIVATE $<BUILD_INTERFACE:libxmp-build-options>)
target_link_libraries(libxmp_lite_static PRIVATE $<BUILD_INTERFACE:libxmp-build-options>)
target_link_libraries(libxmp_lite_shared PRIVATE $<BUILD_INTERFACE:libxmp-build-options>)
target_link_libraries(libxmp-test PRIVATE $<BUILD_INTERFACE:libxmp-build-options>)
```

sezero: Will try later, thanks. (Patches are welcome, btw :)

PR #722, opened 2024-04-28
windows C4996 warning fixes and include tidy-ups :

sezero:
Defines in common.h to avoid MSVC C4996 warnings were defunct, because:
(i) dup, fileno, strnicmp, and S_ISDIR() aren't used anywhere, at least
not for windows, and:
(ii) the rest requires including io.h, and common.h used to be included
before it resulting in a mess.

Therefore, remove those defines from common.h, and define the necessary
ones only where they are needed, namely mkstemp.c and tempfile.c.

Also define `__MINGW_FEATURES__` as 0 if not already defined, in order to
avoid loud -Wundef warnings from old mingw.org headers.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -159,6 +159,7 @@ typedef int tst_uint64[2 * (8 == sizeof(uint64)) - 1];
 #define EVENT(a,c,r)	m->mod.xxt[TRACK_NUM((a),(c))]->event[r]
 
 #ifdef _MSC_VER
+
 #define D_CRIT "  Error: "
 #define D_WARN "Warning: "
 #define D_INFO "   Info: "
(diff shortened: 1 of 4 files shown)


sezero:
I forgot to quote warning messages to show how hilarious the existing
situation is w/o this patch:
```
src\mkstemp.c(85): warning C4996: '_open': The POSIX name for this item is deprecated. Instead, use the ISO C and C++ conformant name: _open. See online help for details.
```
Emphasis on `warning C4996: '_open'` which should have actually been
`warning C4996: 'open'` i.e. our common.h was causing that by redefining
the poisoned prototype..

AliceLR:
> (i) dup, fileno, strnicmp, and S_ISDIR() aren't used anywhere, at least not for windows

`fileno` and `S_ISDIR` were used in the old depacker code and the ProWizard loader. I believe `dup` was used in some very cursed `FILE` duplication code we had to remove but I don't know for sure.

Issue #723, opened 2024-04-30
storlek_20.it test fails on NetBSD 9.2.


AliceLR:
The test for storlek_20.it fails in my NetBSD 9.2 VM with the following message: `pan not left-biased`.

The pan values checked in MINGW64 are: 108 108 108 108 24 24 44 44.
The pan values checked in NetBSD 9.2 are: 0 0 128 128 80 80 0 0.

For this particular check, these values all need to be <128. I don't know if this is an issue with my environment or a real problem.

sezero:
Can this be about char being unsigned by default or something? (shooting in the dark here)

sezero:
> Can this be about char being unsigned by default or something? (shooting in the dark here)

I guess not. Built libxmp and test-dev with -funsigned-char on linux, just passes all tests

neumatho:
Can it have something to do with issue #685 that I reported a while back?

sezero:
Worth looking into.

sezero:
Yes, #685 looks directly related, as far as I can see

sezero:
Closing in favor of #685

PR #728, opened 2024-05-03
Replace rand() with a built-in reentrant alternative.

### LINKED ISSUE #685: Maybe some issue with random panning variations
Hi,

I have noticed that the Test_Storlek_20_Pan_Swing_And_Set_Pan unit test fails sometimes by randomness error, at least in my C# port of XmpLib. I have tried different random algorithms, but it keep fails occasionally.

I then tried to dive into the code on how this is calculated and found this.

In read_event.c file at line 1244, you have this block of code:

```
if (rvv) {
	CLAMP(rvv, 0, 64);
	xc->rpv = rand() % (rvv + 1) - (rvv / 2);
} else {
	xc->rpv = 0;
}
```

In the test, rvv is 64. So in cases, where _rand() % (rvv + 1)_ result in a value below 32, xc->rpv will be set to a negative number.

If we then look in player.c at line 1280, you have this block of code:

```
if (IS_PLAYER_MODE_IT()) {
	finalpan = finalpan + xc->rpv * 4;
}

CLAMP(finalpan, 0, 255);
```

finalpan is zero in the test, so it will result in a negative number by xc->rpv * 4. Then you have the CLAMP, which will set finalpan to zero. So if xc->rpv is set to different numbers, but all negative, the resulting will be zero making the unit test fail, because all values are 0.

I guess there is something wrong somewhere. Either xc->rpv should not be negative or maybe a +128 should be added somewhere. Do you have any idea what to fix or I am totally wrong here and everything is as it should be?

Thomas

AliceLR:
This patch replaces all rand() usage with a built-in implementation of xorshift32. This is not an especially cryptographically good PRNG but it is fast (4 bitshifts, 3 XORs, 1 multiply for output) and produces reasonably "random" output.

Unlike rand(), this implementation is reentrant i.e. it easily supports multiple different concurrent states and doesn't cause any race conditions with caller code outside of libxmp.

This implementation will also behave exactly the same on all platforms given a fixed seed, which allows the fuzzer/xmpchk and the storlek_20.it test to use fixed seeds. (Lack of doing this has caused failures of this test on NetBSD and in neumatho's C# port.)

Fixes #685 and #723.

### DIFF
--- a/Makefile.vc
+++ b/Makefile.vc
@@ -48,6 +48,7 @@ OBJS	= \
  src\effects.obj \
  src\mixer.obj \
  src\mix_all.obj \
+ src\rng.obj \
  src\load_helpers.obj \
  src\load.obj \
  src\hio.obj \
@@ -236,6 +237,7 @@ LITE_OBJS	= \
  src\lite\lite-hio.obj \
  src\lite\lite-smix.obj \
  src\lite\lite-memio.obj \
+ src\lite\lite-rng.obj \
  src\lite\lite-win32.obj \
  src\lite\lite-common.obj \
  src\lite\lite-itsex.obj \
--- a/cmake/libxmp-sources.cmake
+++ b/cmake/libxmp-sources.cmake
@@ -17,6 +17,7 @@ set(LIBXMP_SRC_LIST
     src/effects.c
     src/mixer.c
     src/mix_all.c
+    src/rng.c
     src/load_helpers.c
     src/load.c
     src/hio.c
@@ -201,6 +202,7 @@ set(LIBXMP_SRC_LIST_LITE
     src/lite/lite-hio.c
     src/lite/lite-smix.c
     src/lite/lite-memio.c
+    src/lite/lite-rng.c
     src/lite/lite-win32.c
     src/lite/lite-common.c
     src/lite/lite-itsex.c
--- a/lite/src/Makefile
+++ b/lite/src/Makefile
@@ -2,11 +2,12 @@
 SRC_OBJS	= virtual.o format.o period.o player.o read_event.o \
 		  misc.o dataio.o lfo.o scan.o control.o filter.o \
 		  effects.o mixer.o mix_all.o load_helpers.o load.o \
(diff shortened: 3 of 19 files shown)


sezero:
`make cmake-prepare && make vc-prepare && make watcom-prepare`
`git diff` gives the following missing parts in lite sources:

```diff
diff --git a/Makefile.vc b/Makefile.vc
index 2f9578b..7bb5141 100644
--- a/Makefile.vc
+++ b/Makefile.vc
@@ -237,6 +237,7 @@ LITE_OBJS	= \
  src\lite\lite-hio.obj \
  src\lite\lite-smix.obj \
  src\lite\lite-memio.obj \
+ src\lite\lite-rng.obj \
  src\lite\lite-win32.obj \
  src\lite\lite-common.obj \
  src\lite\lite-itsex.obj \
diff --git a/cmake/libxmp-sources.cmake b/cmake/libxmp-sources.cmake
index 875799e..20bcb08 100644
--- a/cmake/libxmp-sources.cmake
+++ b/cmake/libxmp-sources.cmake
@@ -202,6 +202,7 @@ set(LIBXMP_SRC_LIST_LITE
     src/lite/lite-hio.c
     src/lite/lite-smix.c
     src/lite/lite-memio.c
+    src/lite/lite-rng.c
     src/lite/lite-win32.c
     src/lite/lite-common.c
     src/lite/lite-itsex.c
diff --git a/watcom.mif b/watcom.mif
index 87e101b..b25d0e5 100644
--- a/watcom.mif
+++ b/watcom.mif
@@ -246,6 +246,7 @@ LITE_OBJS= &
  src/lite/lite-hio.obj &
  src/lite/lite-smix.obj &
  src/lite/lite-memio.obj &
+ src/lite/lite-rng.obj &
  src/lite/lite-win32.obj &
  src/lite/lite-common.obj &
  src/lite/lite-itsex.obj &
```

sezero (approved):
OK to merge after those missing parts are added

sezero:
Codecov thing has a grudge against me? Yours completed in 3m in both attempts, mine either timed out or took hours. Kinda jealous..

PR #731, opened 2024-05-05
fix typos

RainRat:


AliceLR (approved):

AliceLR:
@sezero I just went ahead and merged this—if you have objections to correcting e.g. old changelog entries we can revert that part, but I don't think it hurts to do so.

sezero:
OK. (I wasn't particularly fond of this one, but doesn't hurt.)

PR #732, opened 2024-05-06
use response files with msvc and watcom makefiles.

sezero:
avoids 'too long' errors on old operating systems.

P.S.: msvc makefile in test-dev is weird, I didn't touch it. If you want
I'll try and mess with it, or I can keep it as is because it works where
it matters.


sezero:
PING ?

AliceLR (approved):

Issue #171, opened 2020-06-23
Sample rate limitations?


mmontag:
`XMP_MAX_SRATE` is defined as 49170 hz: https://github.com/cmatsuoka/libxmp/blame/master/include/xmp.h#L109

Is there any reason we can't bump this up to support playback on, say, 96 kHz hosts? Or is it an arbitrary limit?

AliceLR:
@cmatsuoka @sezero @sagamusix any objections to just ignoring the API here, allowing higher frequencies (up to say 384kHz), and treating `XMP_MAX_FRAMESIZE` as the frame buffer size lower bound instead of as a fixed allocation size? I don't think this would break anything—`xmp_frame_info` lets users access the buffer (😬) but expects you to use the fields `buffer_size` and `total_size` with it rather than `XMP_MAX_FRAMESIZE`.

The define `XMP_MAX_SRATE` makes no appearance whatsoever in the documentation. The only occurrence of `XMP_MAX_FRAMESIZE` in the documentation is for `xmp_get_frame_info`, which states the buffer size will be no larger than `XMP_MAX_FRAMESIZE`.

sezero:
I don't have any objections myself. (Hopefully it doesn't break anything.)

sagamusix:
Sounds sensible. One thing to keep in mind is that in general, just rendering at a higher mix rate doesn't automatically mean that the result will be better, and some more adjustments may have to be made. For example, resonant filter characteristic will change drastically (they can be considered to be broken) if the fractional precision of filter coefficients is not big enough. Hence filter behaviour should be tested at the highest supported mix rates. In OpenMPT I had to add one or two more fractional bits for filters to keep working at such high rates.
Right now there is no OPL support in libxmp, but Opal also had a bug where its linear resampling would break once the mix rate exceeds 65536 Hz, for example, so in general the output of the mixing routines should be closely observed after bumping the limits.

AliceLR:
> For example, resonant filter characteristic will change drastically (they can be considered to be broken) if the fractional precision of filter coefficients is not big enough. Hence filter behaviour should be tested at the highest supported mix rates. In OpenMPT I had to add one or two more fractional bits for filters to keep working at such high rates.

OK, I will keep this in mind and test it.

> Right now there is no OPL support in libxmp, but Opal also had a bug where its linear resampling would break once the mix rate exceeds 65536 Hz, for example, so in general the output of the mixing routines should be closely observed after bumping the limits.

[...]

PR #740, opened 2024-05-26
Add sample preamplification to filter mixers.

AliceLR:
This is required to correct rounding errors that accumulate when using higher sample rates than libxmp currently allows, making filtered channels with low cutoffs sound distorted. Unfortunately it also messes up all of the mixer tests (again...).

I also increased `FILTER_BITS` to the maximum useful value (~~24~~ 22, since the filter parameters are calculated with single precision floats, and also x87/IEEE-754 rounding differences). This is also necessary for higher sample rates.

This is a prerequisite for fixing #171.


AliceLR:
Failure looks like cumulative x87 rounding differences due to the expanded precision of the filter calculations. Dropping the precision on the filter parameters to 22 bits or changing the filter calculations to use double precision floats fixes it. Fix incoming (just changing the precision to 22 bits...).

sezero:
is this finalized ?

AliceLR:
> is this finalized ?

??? Yes, I marked it ready for review.

sezero:
OK. @sagamusix , @cmatsuoka : Any objections?

sagamusix (approved):

sezero:
OK, this is in.

PR #741, opened 2024-05-28
Increase libxmp maximum sample rate to 384kHz (ignore API).

### LINKED ISSUE #171: Sample rate limitations?

AliceLR:
This makes the mixer initialization allow high rates that previously would have been rejected by `xmp_start_player` calls. This means the API values `XMP_MAX_SRATE` and `XMP_MAX_FRAMESIZE`, while unmodified, are effectively suggestions that will be ignored if someone requests a rate above 48kHz.

Since the mixer will never allocate less than `XMP_MAX_FRAMESIZE` samples for the buffers, and since `xmp_start_player` previously rejected these high rates, this shouldn't break any users *unless* someone was both:

1) relying on `xmp_start_player` rejecting these rates, and *also*
2) doing something very weird with the buffer pointer returned by `xmp_get_frame_info` that almost certainly would have broke before anyway.

More on `xmp_get_frame_info`, the only reason `XMP_MAX_FRAMESIZE` is in the API at all: it returns a pointer to the downmixed buffer in the field `xmp_frame_info::buffer`, and then provides two fields to bound it. The first is `xmp_frame_info::buffer_size`, which is in bytes, and the second is `xmp_frame_info::total_size`, which is in samples (frames × channels) and is *always* `XMP_MAX_FRAMESIZE`. If someone is using these fields as intended (not as documented; they barely are!) to bound accesses to this pointer, nothing should break, and if someone is relying on `XMP_MAX_FRAMESIZE` directly, at worst they won't read the whole buffer. So, I don't think this point of "compatibility" really matters.

Regardless, carefully consider before merging this, and if we need to wait for a minor version or something, that's okay. This theoretically changes the API, but the API was stupid here anyway. Fixes #171. Doesn't fix the crash bugs caused by `xmp_set_tempo_factor`, but I'm working on another patch for that.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -115,7 +115,7 @@ error. Error codes are::
 If a system error occurs, the specific error is set in ``errno``.
 
 Parameters to `xmp_start_player()`_ are the sampling
-rate (up to 48kHz) and a bitmapped integer holding one or more of the
+rate (up to 384kHz) and a bitmapped integer holding one or more of the
 following mixer flags::
 
   XMP_MIX_8BIT          /* Mix to 8-bit instead of 16 */
@@ -599,7 +599,7 @@ int xmp_start_player(xmp_context c, int rate, int format)
     :c: the player context handle.
 
     :rate: the sampling rate to use, in Hz (typically 44100). Valid values
-       range from 8kHz to 48kHz.
+       range from 8kHz to 384kHz.
 
     :flags: bitmapped configurable player flags, one or more of the
       following::
@@ -709,7 +709,8 @@ void xmp_get_frame_info(xmp_context c, struct xmp_frame_info \*info)
       This function should be used to retrieve sound buffer data after
       `xmp_play_frame()`_ is called. Fields ``buffer`` and ``buffer_size``
       contain the pointer to the sound buffer PCM data and its size. The
-      buffer size will be no larger than ``XMP_MAX_FRAMESIZE``.
+      buffer size will be no larger than ``XMP_MAX_FRAMESIZE`` if the player
+      is initialized with a rate equal to or less than ``XMP_MAX_SRATE``.
 
 .. _xmp_end_player():
 
--- a/src/common.h
+++ b/src/common.h
@@ -125,6 +125,7 @@ typedef int tst_uint64[2 * (8 == sizeof(uint64)) - 1];
 #endif
 
 /* Constants */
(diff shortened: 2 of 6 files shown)


sezero (approved):
OK by me. @sagamusix , @cmatsuoka ?

AliceLR:
> edit: if they allocate their own buffer based on XMP_MAX_FRAMESIZE and
then use xmp_frame_info::buffer_size to copy it, this could
catastrophically break. I'm skeptical of this situation due to the poor
documentation but it is a possibility.

I'm starting to lean towards saying "hey, this needs to be a minor version"
because I have a growing bad feeling that this case exists somewhere in the
wild. With a minor version the `XMP_MAX_SRATE` value in xmp.h could also
actually be changed directly (instead of this hacky workaround), if
everyone at least agrees that that would be a minor version change rather
than a major version change. I would like to hear Claudio's opinion on
either approach, though.

sezero:
> edit: if they allocate their own buffer based on XMP_MAX_FRAMESIZE and
> then use xmp_frame_info::buffer_size to copy it, this could catastrophically break. I'm skeptical of this situation due to the poor documentation but it is a possibility. I'm starting to lean towards saying "hey, this needs to be a minor version" because I have a growing bad feeling that this case exists somewhere in the wild. With a minor version the `XMP_MAX_SRATE` value in xmp.h could also actually be changed directly (instead of this hacky workaround), if everyone at least agrees that that would be a minor version change rather than a major version change. I would like to hear Claudio's opinion on either approach, though. 

I tend to agree that this would be better handled in 4.7.0 or something.

@cmatsuoka ?

sezero:
How are the concerns you mentioned above now, now that #742 is in?

AliceLR:
#742 is again completely irrelevant to this change aside from affecting the same function.

AliceLR:
I've switched this PR to draft to reduce any temptation to merge it for now :^) however, testing in MegaZeux since September hasn't really revealed any issues with it.

AliceLR:
Will make another pass over this before marking it ready. (Probably, the maximum rate can just be updated directly.)

sezero (approved):

PR #742, opened 2024-05-29
Fix crashes caused by xmp_set_tempo_factor.

AliceLR:
* `xmp_set_tempo_factor` now returns `-XMP_ERROR_STATE` when called prior to `xmp_start_player` (instead of causing crashes).
* Fix mixer crashes caused by previously valid tempo factors after sample rate or BPM changes.

If the new return value is too much of a problem for 4.6.1 I can change it to -1.

The crash is easily reproduceable by setting the time factor to some nonsense prior to calling `xmp_start_player` (`xmp_set_tempo_factor` will also generate an UndefinedBehaviorSanitizer error):
```C
int main(void)
{
  xmp_context ctx = xmp_create_context();
  xmp_load_module(ctx, "test-dev/data/test.it");
  xmp_set_tempo_factor(ctx, 10000.0);
  xmp_start_player(ctx, 44100, 0);
  xmp_play_frame(ctx); // crash
  xmp_free_context(ctx);
  return 0;
}
```

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -14,6 +14,10 @@ Stable versions
 	- Fix out-of-bounds reads in His Master's Noise Mupp instruments.
 	- Fix slow ProWizard testing.
 	- Replace rand() with a built-in reentrant alternative.
+	- xmp_set_tempo_factor now returns -XMP_ERROR_STATE when called prior
+	  to xmp_start_player (instead of causing crashes).
+	- Fix mixer crashes caused by previously valid tempo factors after
+	  sample rate or BPM changes.
 	Changes by Thomas Neumann:
 	- Fix XM envelope handling
 	- Bug fixes to DSMI loader
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -801,7 +801,8 @@ int xmp_set_tempo_factor(xmp_context c, double val)
     :val: the new multiplier.
 
   **Returns:**
-    0 on success, or -1 if value is invalid.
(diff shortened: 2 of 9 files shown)


sezero:
This actually looks good to me

> If the new return value is too much of a problem for 4.6.1 I can change it to -1.

No need, just fine as it is, IMO

This will clear the way to #741, yes?

docs/Changelog
```diff
@@ -14,6 +14,10 @@ Stable versions
 	- Fix out-of-bounds reads in His Master's Noise Mupp instruments.
 	- Fix slow ProWizard testing.
 	- Replace rand() with a built-in reentrant alternative.
+	- xmp_set_tempo_factor now returns -XMP_ERROR_STATE when called prior
+	  to xmp_start_player (instead of causing crashes).
+	- Fix mixer crashes caused by previously valid tempo factors after
+	  sample rate or BPM changes.
```
sezero, line 20: These need to go into lite/ChangeLog too.

And we don't have any information about sample rate or BPM changes in there.

AliceLR:
> This will clear the way to #741, yes?

This fix is for an unrelated bug I found while implementing #741, actually. I will have to update that pull request after this is merged.

docs/Changelog
AliceLR: OK, I duplicated these lines into lite/Changelog.

sezero (approved):

sezero:
OK, this is in.

PR #744, opened 2024-05-29
Reformat mixer macros in preparation for stereo mixers.

AliceLR:
* The interpolation macros now take an output sample and a sample number (required for stereo mixers for obvious reasons). The channel count is provided via a `const int` for lack of a better way of doing this (should not harm most builds). The alternative was yet another macro parameter to the interpolators and UPDATE_POS.
* Renamed the interpolation macros for consistency.
* I've separated the filter macros from the output macros since the filter macros are supposed to operate on the INPUT according to sagamusix. These macros take one or two input+output variables.
* The output macros have been greatly simplified, and now take one or two output variables.
* Similarly, the variable declarations needed an overhaul.
* Reverted 69653c24 as the underlying cause seems to have been fixed during fuzzing and reverting it tidied up the macros.

Marking as a draft until I've had some time to verify this is complete. Stereo mixers will be a separate patch.

### DIFF
--- a/src/mix_all.c
+++ b/src/mix_all.c
@@ -1,5 +1,5 @@
 /* Extended Module Player
- * Copyright (C) 1996-2023 Claudio Matsuoka and Hipolito Carraro Jr
+ * Copyright (C) 1996-2024 Claudio Matsuoka and Hipolito Carraro Jr
  *
  * Permission is hereby granted, free of charge, to any person obtaining a
  * copy of this software and associated documentation files (the "Software"),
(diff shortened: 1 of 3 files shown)


AliceLR:
This has been pretty much done for a while but is awaiting the split-up of my stereo support working branch, which has been cursed by the Digital Tracker loader being in a fairly poor state.

AliceLR:
I don't need to move anything else from my working branch into this patch, so it should be ready for review.

sezero (approved):

PR #746, opened 2024-06-10
Add stereo sample loading support for IT, S3M, XM, and MED.

AliceLR:
Patch 3 for stereo sample support. This adds loader support for IT, S3M, XM, and MED stereo samples. Merging this will temporarily break some modules as the stereo samples are now loaded as interleaved stereo (instead of just the left channel). This will be fixed by two mixer patches, one of which is already open.

Unfortunately, to make this nicely reviewable there's a little bit of a chicken/egg issue (mix function regression tests rely on having at least one working loader), so this has to be merged before I can finish the second mix function patch.

edit: if you want to hear how stereo samples will sound currently, this would be the branch to test: https://github.com/libxmp/libxmp/compare/master...AliceLR:mixer-working-branch


sezero (approved):
Minor nit: Fails building using c++ compilers:
```diff
diff --git a/src/loaders/it_load.c b/src/loaders/it_load.c
index 5011492..66c6605 100644
--- a/src/loaders/it_load.c
+++ b/src/loaders/it_load.c
@@ -769,7 +769,7 @@ static uint8 *unpack_it_sample(struct xmp_sample *xxs,
 			pos += xxs->len;
 		}
 	}
-	return decbuf;
+	return (uint8 *)decbuf;
 }
 
 static int load_it_sample(struct module_data *m, int i, int start,
```

OK with that change.

AliceLR:
> OK with that change.

Fixed this oversight in a slightly different way: I just changed the return value of the function to match since the caller also uses it as `void *`.

sezero:
OK, feel free to merge when CI completes

Issue #600, opened 2022-08-06
support for stereo samples in MMDx missing


raboho3:
I found some MMD3 files on aminet which contain stereo samples and those files are basically replayed silently.
Example file is misc/NR2019Collection.lha#/Musikk/Trackers/Soundstudio/HKMods/NinjaRemix/DungeonNinja

I found that commit 6f74dc886e26931081fc451d23241de20c90e2aa basically disabled stereo sample loading during refactoring.

When using the load code for non-stereo samples in mmd3_load.c:608, it kinda works but still some samples are silent (left channel at the beginning of the example file).

It looks like commit 35b716dadd78a262e04d96fd6efb32432125be7c changed the octave handling which silences these samples.

So there are multiple issues when playing these files. Best case would be to implement support for the stereo samples (and fix octave handling), but until this fix is available, I would rather fail on loading such mod files instead `of` playing back silence.

As for now I use this patch so I can check with --load-only if the file is supported or not.

```
diff --git a/src/loaders/mmd3_load.c b/src/loaders/mmd3_load.c
index 327eb818..4cffeb0e 100644
--- a/src/loaders/mmd3_load.c
+++ b/src/loaders/mmd3_load.c
@@ -606,9 +606,8 @@ static int mmd3_load(struct module_data *m, HIO_HANDLE *f, const int start)
 
 			continue;
 		} else if ((instr.type & STEREO) != 0) {
-			D_(D_WARN "stereo sample unsupported");
-			mod->xxi[i].nsm = 0;
-			continue;
+			D_(D_CRIT "stereo sample unsupported");
+			goto err_cleanup;
 		} else if ((instr.type & S_16) == 0) {	/* Sample */
 			int ret;
 

```

AliceLR:
Allowing modules to load with broken/partial stereo samples is currently the norm for these, even though it's not ideal. Currently IT and XM load just the left channel.

For when I get around to properly addressing this: here's the full list of MMD3s on Modland with stereo samples:
```
- unknown/for the unknown cause.mmd3
- unknown/jolpoid.mmd3
- unknown/nazina.mmd3
Asarhad/calculated risk [asarhad rmx].mmd3
Caustic/ocean legends.mmd3
CJ/absolution2.mmd3
CJ/dizzy goes cruising.mmd3
CJ/gotter dammerung.mmd3
Dazzy/perfection (dazzy's perfect mix).mmd3
Dazzy/seduction.mmd3
[...]

PR #747, opened 2024-06-10
Add mix functions for interleaved stereo samples.

### LINKED ISSUE #588: [Bug] Psirius Memories of Time differs between OMPT and libxmp
This have playback issues between OMPT and Libxmp with weird artifact plays on Ord 1 on libxmp, where OpenMPT doesn't.

[Modarchive](https://modarchive.org/index.php?request=view_by_moduleid&query=185766)

AliceLR:
Patch ~~4~~ 5 for stereo sample support. Adds and enables interleaved stereo sample mixers, and updates the loop wraparound smoothing code to correctly handle stereo samples. I reorganized the start of mixer.c slightly as well.

Closes #588 and #600.

Marked as a draft pending the regression tests, which are going to blow up the line count of this PR quite a bit. Feel free to review my progress as-is.


AliceLR:
Okay, I'm (sort of) certain I didn't screw anything up here. I left it as two separate commits if you don't want to sift through the regression test junk.

sezero (approved):
OK with me. Waiting for @sagamusix as you requested.

src/mixer.c
```diff
@@ -59,33 +54,52 @@ struct loop_data
...
+ * bit 2: 0=mono output, 1=stereo output
+ * bit 3: 0=unfiltered, 1=filtered
+ */
+
+#define FLAG_16_BITS	0x01
+#define FLAG_STEREO	0x02
+#define FLAG_STEREOOUT	0x04
+#define FLAG_FILTER	0x08
+#define FLAG_ACTIVE	0x10
+/* #define FLAG_SYNTH	0x20 */
```
sagamusix, line 70: Maybe it's time to retire this flag, as there is absolutely no indication what it's even supposed to do.

AliceLR: I have no clue either—I *assume* it's meant to signal to the mixer that some notes shouldn't be played as samples, but instead should be passed to a synthesizer. It seems like a mistake to me to send notes/effects to the synthesizer *after* the `read_event` functions and player.c have processed them but whatever.

I was just going to leave figuring this flag out until I add a new OPL player, whenever that is.

sagamusix: Looking back at a really old code revision, this does indeed seem to be the only remnant from OPL emulation. Maybe it has to be this way in libxmp but in OpenMPT such voices wouldn't even make their way into the mixer, so this seems strange to me.

src/mixer.c
```diff
@@ -95,79 +109,45 @@ MIX_FN(stereoout_mono_a500);
...
-	libxmp_mix_monoout_mono_8bit_spline_filter,
-	libxmp_mix_monoout_mono_16bit_spline_filter,
-	libxmp_mix_stereoout_mono_8bit_spline_filter,
-	libxmp_mix_stereoout_mono_16bit_spline_filter
+	LIST_MIX_FUNCTIONS(spline_filter)
 #endif
 };
 
 #ifdef LIBXMP_PAULA_SIMULATOR
+#define LIST_MIX_FUNCTIONS_PAULA(type) \
```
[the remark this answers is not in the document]
AliceLR: > Does libxmp allow the Paula filter to be used with MED modules?

Doesn't look like it: all Paula mixer handling is gated with `IS_AMIGA_MOD()`, which requires both `m->read_event_type == READ_EVENT_MOD` and `m->period_type == PERIOD_MODRNG`. Supporting it for MED sounds reasonable but also like a separate issue; aside from new mixers, something would also have to be done about the `PERIOD_MODRNG` restriction.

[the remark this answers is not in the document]
sagamusix: Alright, so no issue here (unless there is any other MOD-like format that could have 16-bit samples...)

sagamusix (approved):

sezero:
Is there anything left to do with this patch?

AliceLR:
> Is there anything left to do with this patch?

No, I don't think so.

sezero:
OK, this is in.

PR #748, opened 2024-06-12
Deduplicate repetitive mixer regression test code.

AliceLR:
Patch 4 for stereo samples support. The mixer output regression tests contained a ton of duplicate code and some other tests also relied on similar routines, so I combined it all into one place.

This raises a question that should probably be resolved before I finish #747: the test data for mono and stereo output could be separated *now* and I could put panning effects in the preset mixer test data. This would double the number of mixer output data files (which are going to be doubled again by #747) but get rid of a few ugly hacks in `compare_mixer_samples_ext` and ensure that panning is being accounted for in the mixer output. Should we do this now or save it for some other point in the future?

edit: OpenWatcom failure is not this branch's fault, for some reason CI couldn't download the tarball.


sezero:
> This raises a question that should probably be resolved before I finish #747: the test data for mono and stereo output could be separated _now_ and I could put panning effects in the preset mixer test data. This would double the number of mixer output data files (which are going to be doubled again by #747) but get rid of a few ugly hacks in `compare_mixer_samples_ext` and ensure that panning is being accounted for in the mixer output. Should we do this now or save it for some other point in the future?

Really your choice. However, it's an improvement, so if you're willing to do it now, why do later?

> edit: OpenWatcom failure is not this branch's fault, for some reason CI couldn't download the tarball.

Bad timing, I guess: Looks like we hit the time when ow was updating their release tarball. Completed with success now.

AliceLR:
> Really your choice. However, it's an improvement, so if you're willing to do it now, why do later?

Okay, I'll go ahead and do it now. I don't feel that strongly about it and I just wanted to make sure there weren't any objections to making these tests even *more* annoying.

sezero (approved):

PR #754, opened 2024-06-28
Added an extra check on the restart value for XM modules

neumatho:
Added an extra check on the restart value for XM modules, so it is possible to play "10 Days Of Abstinence.xm".

This module has a value of 0xffff as the restart and before this change, the loader failed. Now it can load this module and play it.

You can find the module here:

https://modland.com/pub/modules/Fasttracker%202/Andy%20(DE)/10%20days%20of%20abstinence.xm

### DIFF
--- a/src/loaders/xm_load.c
+++ b/src/loaders/xm_load.c
@@ -814,10 +814,6 @@ static int xm_load(struct module_data *m, HIO_HANDLE * f, const int start)
 		return -1;
 	}
 
-	if (xfh.restart > 255 && xfh.restart != 0xffff) {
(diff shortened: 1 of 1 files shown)


sagamusix:
Instead of adding a second special case for rejecting files, xmp should probably just not reject files based on the restart position at all, and simply ignore it (set it to 0) whenever the restart position is past the end of the order list.

Technically the XM format supports an order list length of up to 65536 due to the variable-length header so there's nothing inherently malformed about such a file.

neumatho:
I have now updated my PR with @sagamusix suggestions.

AliceLR (approved):
The revised version of this looks good to me. Thanks!

Issue #724, opened 2024-05-03
integrate libxmp-lite test to make check in full tree


sezero:
As noted by @AliceLR in #721: Since libxmp-lite is now integrated into
the main build systems, the libxmp-lite test should also be integrated
into make check when lite is enabled.

Since lite can be built w/o it support, lite test needs making to work
with a new xm (or protracker) test module, with its MD5s generated for
both IEEE-754 and x87 floating point.

Here is a first draft patch for lite-test in full tree: [test-lite.patch.txt](https://github.com/libxmp/libxmp/files/15204058/test-lite.patch.txt)

PR #764, opened 2024-10-02
integrate libxmp-lite test to 'make check' in full tree.

sezero:
disables 'make check' for lite version with an XFAIL message if IT
loader is disabled, for now.

Also fixes a bug in cmake script preventing exporting of sybols in
unix builds (looks like something missed in PR/721)

Closes https://github.com/libxmp/libxmp/issues/724

@AliceLR: Feel free to push to my branch to add an xm or a mod test
for lite test w/o it and revert the xfail change to test/test.c


test/Makefile
```diff
@@ -24,3 +27,10 @@ $(TEST_PATH)/libxmp-test: $(T_OBJS)
 	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo LD $@ ; fi; \
 	eval $$CMD
 
+check-lite: $(TEST_PATH)/libxmp-lite-test
```
AliceLR, line 30: It might be worth adding this to the distribution packages CI job as part of this patch, to verify it passes under CI. (It doesn't look like the MSVC/Watcom regression tests are currently building lite...)

AliceLR:
> @AliceLR: Feel free to push to my branch to add an xm or a mod test for lite test w/o it and revert the xfail change to test/test.c

Alright, I will look into this.

test/Makefile
AliceLR: I pushed a commit adding this just to see how it turns out—it seemed fine locally.

AliceLR (approved):
I added a new test.xm with some very basic envelope and other tests (probably not enough...). If my changes are okay with you, let's merge this.

sezero:
All looks good, merging this.

BTW, We should  make a release checklist for 4.6.1 some time (or should it be 4.7.0 ??)

sezero:
Pushed 65b2d3ed5315fb1029a331a9fbc5d0b5d5db2faf manually so that test.xm is isn't missed in distribution tarballs.

AliceLR:
> BTW, We should make a release checklist for 4.6.1 some time (or should it be 4.7.0 ??)

I think a 4.6.1 checklist would be of fairly reasonable length (mostly triage of the new issues...).

sezero:
OK, [4.6.1 milestone](https://github.com/libxmp/libxmp/milestone/7) seems to have 4 open tickets. Are all of them doable? Is there any other than should be moved into 4.6.1?

AliceLR:
I went ahead and added a couple of the open patches. I will look at issues tonight and try to at least triage some of the new ones.

I also have an unfinished Pack-Ice support patch that I stalled on, since I couldn't get any two compilers to agree on half of my optimizations.

sezero:
PR/695 seems like a work-in-progress and may be dangerous.

AliceLR:
> PR/695 seems like a work-in-progress and may be dangerous.

Yeah, it would need an overhaul, but I don't think it would be much effort. (I assume at this point it's abandoned by the author, so I'd just be patching it myself.)

AliceLR:
> Is any part of 624 doable ?

Unlikely. I need to reverse engineer Desktop Tracker's packing again (not a big deal) and the other formats have no progress at all.

AliceLR:
> triage some of the new ones.

Of the ones I could verify issues with, only one of them seems really appropriate for 4.6.1 and I opened a draft patch for it.

PR #766, opened 2024-10-04
Fix restart position for >64k sample and Digital Tracker MODs.

### LINKED ISSUE #763: Restart Positions seem to not be respected on .MOD files
Hi, I'm trying out the example program with an example .MOD file I created (`player-openal-buffer` in specific, since that one doesn't stop after looping), and it seems like the module's Restart Position does not get respected when looping.

I've attached the file that shows what I mean, the file should loop back to the point where the screaming starts,
however when testing it with the example program it loops back to the beginning of the file.

[TEST.MOD.zip](https://github.com/user-attachments/files/17190610/TEST.MOD.zip)

Is there something I'm missing here? Or is this an issue with the library?

AliceLR:
Workaround for a bug caused by a change prior to version tracking. I should note that libxmp-lite is not setting the restart position for MOD anywhere whatsoever as far as I can tell.

Ideally it'd be nice to know why that line was commented out so it can be uncommented.

Fixes #763 for libxmp and  libxmp-lite.

### DIFF
--- a/src/loaders/mod_load.c
+++ b/src/loaders/mod_load.c
@@ -542,11 +542,13 @@ static int mod_load(struct module_data *m, HIO_HANDLE *f, const int start)
     strncpy(mod->name, (char *) mh.name, 20);
 
     mod->len = mh.len;
-    /* mod->rst = mh.restart; */
-
-    if (mod->rst >= mod->len)
-	mod->rst = 0;
     memcpy(mod->xxo, mh.order, 128);
+
+    if (mh.restart < 0x7f && mh.restart != 0x78 && (int)mh.restart < mod->len) {
+	/* TODO: an older version of this code was commented out 23+ years ago
+	 * and adding this may have rebroke something. */
+	mod->rst = mh.restart;
+    }
 
     for (i = 0; i < 128; i++) {
 	/* This fixes dragnet.mod (garbage in the order list) */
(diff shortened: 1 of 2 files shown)


AliceLR:
(Note this probably needs some work. I will ask Saga Musix about sensible default behavior for libxmp-lite.)

AliceLR:
> (Note this probably needs some work. I will ask Saga Musix about sensible default behavior for libxmp-lite.)

Summary of the feedback: the "standard" handling that is done by OpenMPT et al. is to use this byte as a restart position if it is less than the length of the module and *not* equal to `0x78`. I will revise the patch to just do this for both libxmp and libxmp-lite and we can see what (if anything) breaks. FWIW this seems equivalent to what the tracker ID routine is already doing.

AliceLR:
Doing it this way also happened to fix this feature for Digital Tracker modules :)

sezero:
Is this ready?

BTW, is mh.restart == 0x78 always illegal to be a restart point?

AliceLR:
> Is this ready?

Probably, yes.

> BTW, is mh.restart == 0x78 always illegal to be a restart point?

I think this is going to depend on whichever tracker makes the module, but it's an extremely *unlikely* starting point. Finding a 0x78 here is much more likely sign of Noisetracker IIRC, so that's how it's interpreted.

sagamusix:
I think I have two modules in my collection whose order list is longer than 0x78 and that have a restart position of 0x78 at all - in both cases not to be interpreted as a restart position. Like Alice said I think it's pretty unlikely to find a module where this is not the case.

sezero:
OK then. Merge?

sezero (approved):

PR #772, opened 2024-11-10
Fix test_mixer_bidi_sync to test different interpolation modes.

### LINKED ISSUE #689: test_mixer_bidi_sync part 2
Hi again,

Created a new issue. This is a reference to issue #681.

I just saw now, that you have replaced the wrong argument in xmp_set_player for interp. It is the last argument in the method call that should be replaced. I tried to make a new PR with this change, but when running the test it begin to fail on the first mixer test. Probably because it now uses the nearest instead of spline, which was hardcoded before.

So I don't know what is needed to fix it or if you still want the test.

AliceLR:
Fixes #689. I just removed the single broken test case (XM, nearest, 8000Hz) as it inexplicably contains a pop 3/4 through that only seems to affect very low rates and doesn't otherwise break the test. Nearest passes at 11025Hz for both modules.

### DIFF
--- a/test-dev/test_mixer_bidi_sync.c
+++ b/test-dev/test_mixer_bidi_sync.c
@@ -24,7 +24,7 @@ static void bidi_sync_test_mode(xmp_context opaque, int rate, int interp, const 
 	int i, j;
 
 	xmp_start_player(opaque, rate, XMP_FORMAT_MONO);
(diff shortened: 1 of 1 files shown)


AliceLR:
thanks codecov

sezero (approved):

PR #774, opened 2024-11-11
Merge song file instrument path detection routines.

AliceLR:
I found a few glaring issues that needed addressing while fixing a MED2 loader snprintf truncation warning related to sample loading. The MOD, STM, MED2, and (bitrot) SSMT loaders all contain separate routines to generate the path of an instrument/sample file when loading song files. These are implemented roughly as follows:

* MOD: combine `m->dirname` with the cleaned sample name, open.
* STM: combine `m->dirname` with the cleaned sample name, open.
* SSMT: combine `m->dirname` with the cleaned sample name, open.
* MED2:
  1) Get the instrument path set by `xmp_set_instrument_path` or if it is not set, get the environment variable `XMP_INSTRUMENT_PATH`, or if it is not set, `.`. Combine this with the cleaned sample name and attempt a case-insensitive match, then open.
  2) Combine `m->dirname` with the cleaned sample name, attempt a case-insensitive match, then open. This was added by #416 for consistency with the other loaders.

Clearly, one of these is not like the others! There are a few major issues here.

1) The MOD, STM, and SSMT loaders totally ignore
   `xmp_set_instrument_path` and/or `XMP_INSTRUMENT_PATH`, which
   sucks and means this feature is only useful for MED2, a format
   where literally no test files existed and I had to make my own
   for libxmp's regression tests. Note that this feature allows
   memory modules to load instruments, and `m->dirname` only works
   for file modules.
2) The MOD, STM, and SSMT loaders fail to check for instrument files
   case-insensitively. MOD is from the same platform as MED2 and STM
   is explicitly from a case-insensitive operating system, so this is
   a strange omission.
3) MED2, on the other hand, attempts to check the current directory
   if neither `xmp_set_instrument_path` or `XMP_INSTRUMENT_PATH` is
   set, which is strange and seems like a worse version of loading
   from `m->dirname`. Worse, if this was applied to MOD/STM, it could
   cause new misbehavior that `xmp_set_instrument_path` and
   `XMP_INSTRUMENT_PATH` are less likely to, so I removed it.
   Intentional "regressions" to MED2 are better than to MOD/STM.
4) No failure checks on any of the `snprintf` occurrences, so really
   these `fopen` calls could be opening anything if they truncate.

The new, combined handling for all loaders is now:

1) Get `xmp_set_instrument_path` or `XMP_INSTRUMENT_PATH` if it is
   set, combine it with the cleaned sample name, and attempt a
   case-insensitive search. If all the copies were successful and
   the file was found, open.
2) Get `m->dirname` if it is set, combine it with the cleaned sample
   name, and attempt a case-insensitive search. If all the copies
   were successful and the file was found, open.

TANGENT TIME: `strncpy` is the worst C standard library function. `strncpy` will pad its entire buffer with `\0` even when this isn't wanted or requested, but when the input string is longer than the output buffer, it will not null terminate the output buffer. It is not safe. After merging the handlers to address the above issues, GCC's stringop analyzer found several misuses of this function that have been hiding in plain sight for years and opening us up to more out-of-bounds reads. I've replaced them all with `snprintf` as this also allows the detection of truncation with its return value, which in turn can be carried through as a fail state for sample loading.

Please review carefully: I'm not sure that I actually have any song files around to test with except for the MED2 in test-dev. Marking as a draft for now so I can re-review it myself after some sleep.

### DIFF
--- a/src/bitrot/loaders/ssmt_load.c
+++ b/src/bitrot/loaders/ssmt_load.c
@@ -14,10 +14,10 @@
 /* From the deMODifier readme:
  *
  * SoundSmith was arguably the most popular music authoring tool for the
- * Apple IIgs.  Introduced in the IIgs's heyday (which was, accurately 
+ * Apple IIgs.  Introduced in the IIgs's heyday (which was, accurately
  * enough, just about one day), this software inspired the creation
- * of countless numbers of IIgs-specific tunes, several of which were 
- * actually worth listening to.  
+ * of countless numbers of IIgs-specific tunes, several of which were
+ * actually worth listening to.
  */
 
 #include "loader.h"
@@ -76,7 +76,7 @@ static int mtp_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	blocksize = hio_read16l(f);
 	mod->spd = hio_read16l(f);
 	hio_seek(f, 10, SEEK_CUR);		/* skip 10 reserved bytes */
-	
+
 	mod->ins = mod->smp = 15;
 	if (libxmp_init_instrument(m) < 0)
 		return -1;
@@ -182,17 +182,14 @@ static int mtp_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	for (i = 0; i < mod->ins; i++) {
 		struct xmp_instrument *xxi = &mod->xxi[i];
 		HIO_HANDLE *s;
-		char filename[1024];
+		char filename[XMP_MAXPATH];
 		char tmpname[32];
 
-		if (!m->dirname) {
-			return -1;
-		}
-
-		if (!xxi->name[0] || libxmp_copy_name_for_fopen(tmpname, xxi->name, 32))
+		if (libxmp_copy_name_for_fopen(tmpname, xxi->name, 32) != 0)
 			continue;
 
-		snprintf(filename, 1024, "%s%s", m->dirname, tmpname);
+		if (!libxmp_find_instrument_file(m, filename, sizeof(filename), tmpname))
+			continue;
 
 		if ((s = hio_open(filename, "rb")) != NULL) {
 			asif_load(m, s, i);
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -443,7 +443,7 @@ int libxmp_check_filename_case(const char *dir, const char *name, char *new_name
 int libxmp_check_filename_case(const char *dir, const char *name, char *new_name, int size)
 {
 	int found = 0;
-	int ret;
+	int ret = size;
 	DIR *dirp;
 	struct dirent *d;
 
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -299,7 +299,7 @@ int libxmp_copy_name_for_fopen(char *dest, const char *name, int n)
 	 * malicious when given to fopen. This should only be used on song files.
 	 */
 	if (!strcmp(name, ".") || strstr(name, "..") ||
-	    name[0] == '\\' || name[0] == '/' || name[0] == ':')
+	    name[0] == '\\' || name[0] == '/' || name[0] == ':' || name[0] == '\0')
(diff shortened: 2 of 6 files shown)


AliceLR:
The original MED2 warning:
```
src/loaders/med2_load.c: In function 'med2_load':
src/loaders/med2_load.c:199:57: warning: '%s' directive output may be truncated writing up to 255 bytes into a region of size between
4 and 259 [-Wformat-truncation=]
  199 |                         snprintf(path, XMP_MAXPATH, "%s/%s", ins_path, name);
      |                                                         ^~             ~~~~
src/loaders/med2_load.c:199:25: note: 'snprintf' output between 2 and 512 bytes into a destination of size 260
  199 |                         snprintf(path, XMP_MAXPATH, "%s/%s", ins_path, name);
      |                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

The additional `strncpy` warnings:
```
In function 'libxmp_get_instrument_path',
    inlined from 'libxmp_find_instrument_file' at src/loaders/common.c:484:6:
src/loaders/common.c:472:17: warning: 'strncpy' specified bound 256 equals destination size [-Wstringop-truncation]
  472 |                 strncpy(path, getenv("XMP_INSTRUMENT_PATH"), size);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/loaders/common.c:469:17: warning: 'strncpy' specified bound 256 equals destination size [-Wstringop-truncation]
  469 |                 strncpy(path, m->instrument_path, size);
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function 'libxmp_check_filename_case',
    inlined from 'libxmp_find_instrument_file' at src/loaders/common.c:485:6:
src/loaders/common.c:438:9: warning: 'strncpy' specified bound 256 equals destination size [-Wstringop-truncation]
  438 |         strncpy(new_name, name, size);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In function 'libxmp_check_filename_case',
    inlined from 'libxmp_find_instrument_file' at src/loaders/common.c:493:6:
src/loaders/common.c:438:9: warning: 'strncpy' specified bound 256 equals destination size [-Wstringop-truncation]
  438 |         strncpy(new_name, name, size);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

AliceLR:
Had to revise the MED2 loader instrument debug print, because it was relying on boolean spaghetti to print in most cases and switching to early-return-style failure broke it. Instruments now have two debug lines. If the file is not found, it will print "not found" and no length.

```
med2_load [src/loaders/med2_load.c:194] [ 0] med2test_PanFlute                ---- 0000 0000   V40
med2_load [src/loaders/med2_load.c:214]      (OK)                             26ac
```

sezero (approved):
For the record, I hadn't seen the warnings with my gcc-14 at -O2 or -O3.
However, this looks good to me.

I got one warning:
````
src/loaders/common.c: In function ‘libxmp_check_filename_case’:
src/loaders/common.c:446: warning: ‘ret’ may be used uninitialized in this function
````

OK with that fixed. Maybe something like the following?
````diff
diff --git a/src/loaders/common.c b/src/loaders/common.c
--- a/src/loaders/common.c
+++ b/src/loaders/common.c
@@ -444,5 +444,5 @@ int libxmp_check_filename_case(const char *dir, const char *name, char *new_name
 {
 	int found = 0;
-	int ret;
+	int ret = INT_MAX;
 	DIR *dirp;
 	struct dirent *d; 
````

sezero:
> Seems bogus 

Yeah :) But yes, patch it.

sezero:
.. and feel free to merge when CI shows green. (the new 775 too.)

Issue #717, opened 2024-03-20
"sleeping waste" has broken PWM playback?


SergioFLS:
Using current `master`, version 4.6.0 and Xmp for Android v4.12.0 (which reportedly uses libxmp 4.4.1).

When I play meo-sleeping_waste.mod (https://modarchive.org/module.php?68014) on openmpt123, there's a PWM effect on the fourth channel playing at around 41 seconds into the song. On the listed libxmp versions I've tried, the wave only seems to be a pulse wave that doesn't change.

sezero:
CC: @AliceLR

nwoeanhinnogaehr:
It's because the module changes samples without retriggering the note. Protracker 2 and 3 differ in the handling of this - what you are looking for is what Protracker 2 does. Personally I'd also like to see xmp support this. As far as I'm aware there are not any Protracker 3 modules that break when samples changes without retriggering are allowed, but I may be wrong.

AliceLR:
As far as I can tell libxmp doesn't do much when playing MODs when an instrument number isn't accompanied by a note. It should be possible to support this when Protracker quirks are active (which libxmp applies to this module, but unfortunately the tracker ID routine is hit-or-miss for others).

As is typical with changing anything related to M.K. MODs I will need to gather some more data before patching this ):

TheEssem:
Plenty of 4mat's Protracker modules have this issue, including but probably not limited to:
- back again.mod (https://modarchive.org/module.php?76215)
- 4mat-menu1.mod (albeit to a lesser extent) (https://modarchive.org/module.php?66427)
- velocity.mod (https://modarchive.org/module.php?65281)

nwoeanhinnogaehr:
Interestingly I noticed that 4mat-menu1.mod is identified as Fast Tracker M.K., and my understanding is that this will disable Protracker quirks (certainly it disables EFx at least). A couple more affected mods which are identified as Protracker M.K.:

- http://modland.com/pub/modules/Protracker/Goto80/minipitt.mod (the whole thing sounds bad in xmp, but especially the arpeggios starting at position 14)
- http://modland.com/pub/modules/Protracker/Clawz/hotwires.mod (channel 4, position 3+)
- http://modland.com/pub/modules/Protracker/MC%20MP/happy%20new%20year%201993.mod (channel 1, position 2+)

[...]

PR #777, opened 2024-11-14
Make the presence of invert loop override tracker ID guesses.

AliceLR:
1) Finding any instance of EFx in a "Noisetracker" MOD switches tracker ID to Protracker immediately during pattern loading (to allow the loading of these effects).
2) Finding any instance of EFx in a MOD without any "decisive" tracker ID result and with all notes in-range switches tracker ID to Protracker unless it is already OpenMPT (to allow playing of EFx as invert loop).

I also added a bunch of debug printing for tracker ID in a second commit. It might be more helpful to read the two commits separately.

This resolves a SIDE issue discussed in #717, but does not affect the primary reported issue there. For a breakdown of the affected MODs, [here is a spreadsheet](https://docs.google.com/spreadsheets/d/12M2qpdDtkG6pBTFZva4ldMwEtInybjOllC43xoNKCjQ/edit?usp=sharing). Out of the 87 total M.K. MODs in the Modland Protracker directory which contain `EFx` in a pattern but are *not* identified by tracker ID as one of the two trackers which would enable that effect (Protracker and OpenMPT):

* 53 use `EFx` in a manner consistent with Invert Loop. These are now recognized as Protracker.
* 11 use `EFx` in a manner NOT consistent with Invert Loop and are now recognized as Protracker (potentially wrongly). Nothing really explains why they contain this effect in the places that they do; in some it's inaccessible, and in others, it's used in channels that do not play looping samples or have very long loops where Invert Loop would be essentially unnoticeable. I don't think any of these have been affected negatively by enabling Protracker quirks.
* 4 use `EFx` in a manner NOT consistent with Invert Loop and are NOT marked as Protracker due to out-of-range notes or something else. Like the former category, these are usually unreachable or nonsense.
* 19 contain `EFx` in corrupt pattern data and have no indication of intentional usage. Out-of-range notes prevent all of these being marked as Protracker. I reported these to the Modland Discord channel and Asle has been dealing with them where appropriate.

This patch does not affect any MODs in the Modland data set aside from the 64 changed to Protracker (that I am aware of).

### DIFF
--- a/src/loaders/mod_load.c
+++ b/src/loaders/mod_load.c
@@ -297,6 +297,8 @@ static int get_tracker_id(struct module_data *m, struct mod_header *mh, int id)
 	int has_vol_in_empty_ins = 0;
 	int i;
 
+	D_(D_INFO "attempting initial tracker ID via header details");
+
 	/* Check if has instruments with loop size 0 */
 	for (i = 0; i < 31; i++) {
 		if (mh->ins[i].loop_size == 0) {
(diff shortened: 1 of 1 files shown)


AliceLR:
Gave the 11 a listen and compared some of them and they seem fine. A handful are relevant to #717 and made it into my list for that. This is OK for review.

PR #781, opened 2024-11-16
Fix crash when calling xmp_set_instrument_path with NULL.

AliceLR:
The API function `xmp_set_instrument_path` crashes when provided with a path of `NULL` due to passing that `NULL` on to `libxmp_strdup`. I've fixed `libxmp_strdup` to return `NULL` when this occurs, and I've revised the behavior of `xmp_set_instrument_path` to clear the instrument path when given `NULL`. This is undefined behavior for `xmp_set_instrument_path` so it should not require a minor version change.

### DIFF
--- a/docs/libxmp.rst
+++ b/docs/libxmp.rst
@@ -919,13 +919,15 @@ int xmp_set_instrument_path(xmp_context c, char \*path)
 ```````````````````````````````````````````````````````
 
   Set the path to retrieve external instruments or samples. Used by some
-  formats (such as MED2) to read sample files from a different directory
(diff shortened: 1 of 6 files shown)


sezero (approved):
This is good: feel free to merge.

P.S.: Should we document the new behavior about calling with NULL parameter in libxmp.rst?

AliceLR:
> P.S.: Should we document the new behavior about calling with NULL parameter in libxmp.rst?

I wasn't sure about this but it'd probably be better to. Will follow up with a second commit, 1 sec...

PR #785, opened 2024-11-26
Implement support for Protracker instrument swapping.

### LINKED ISSUE #717: "sleeping waste" has broken PWM playback?

AliceLR:
When Protracker 1/2 encounter an event with an instrument number but no note (or a porta note), instead of ignoring it (as some other players) or triggering that instrument (like other formats), they replace the current instrument info in the channel and current sample without interrupting the currently playing sample. When the currently playing sample in that channel reaches its end or the end of its loop, the *new* sample will seamlessly start playing from the start or the start of its loop instead. (*Speculation*: the new sample position and bounds in memory aren't fed to Paula until an interrupt triggers at the end of playback.)

This is very commonly used in MOD tunes, especially in chiptunes for PWM effects. Since it is Protracker specific, it currently only applies to `QUIRK_PROTRACK`, which should cover the vast majority of affected modules and *probably* not hurt much else. I don't have real numbers here because, when I searched for MODs containing an instrument number with no note, I got 26k results in Modland (out of 80k). I will happily verify them one-by-one if someone pays me but don't expect results until next year :)

I *think* this is ready for review, but also note that it is being reviewed in #717 too, so it might be worth delaying merge for a few days.

Closes #717. Resolves 8 pending tasks in #497.


sezero (approved):
OK with me.

Merge whenever you like.

AliceLR:
(Rebased so this can be merged without squash, will merge once tests complete.)

PR #787, opened 2024-11-28
Add Digital Tracker stereo sample support and other major bugfixes.

AliceLR:
* Add support for loading Digital Tracker 2.03 DTMs (MOD patterns).
* Add support for loading Digital Tracker 1.9 DTMs (VERS/SV19).
* Add format documentation from Digital Tracker 2.04.
* Better Digital Tracker version fingerprinting.
* Fix Digital Tracker 2.03 global sample rate field.
* Fix Digital Tracker 2.04 pattern note loading (was off-by-one).
* Fix Digital Tracker instrument loops (loop end was off-by-one).
* Add Digital Tracker mono mode and panoramic stereo support.
* Allow patterns up to 396 rows in Digital Home Studio DTMs. (This doesn't help much because they won't load currently.)
* Add support for Digital Tracker 1.9 "MIDI note" transpose.
* Partial support for Digital Tracker stereo samples.
* Simulate Digital Tracker effects bugs where possible.
* Fix loading of Digital Tracker module names (not always 32 bytes).

I realized I had this sitting around mostly complete and that #707 may be dependent on it, as later Digital Tracker/Digital Home Studio versions have different `E6x` behavior. I cut the time-consuming part of this patch (full effects loader tests for coverage) and finished the rest.

I didn't add support for Digital Home Studio DTMs and they will still fail to load, and the old stereo mode hackery is not supported yet, but otherwise this format should be in a much better place now.


sezero (approved):

AliceLR:
I had to make an additional commit to add effect `8xx` for setting panning, fix some things related to panoramic stereo mode (incorrectly referred to as "mono" prior to Digital Home Studio), and support another undocumented 2.03 header field that forces sample depth globally.

AliceLR:
> [codecov actually working again]

Most of the misses are Digital Home Studio, effects (skipped tests due to time constraint; will deal with when Digital Home Studio is revisted), panoramic stereo (can be covered by one of those modules), module text (couldn't get it to save in Digital Tracker 1.917, will try again with Digital Home Studio).

AliceLR:
...and one of the two modules that uses `8xx` on Modland relies on version 2.04 bugs with the extreme values. Another commit coming...

AliceLR:
I added some horrible pan law compensation to properly convert the old `8xx` variant that might need more review. Otherwise, I think I need to stop working on this branch (before it turns into adding an entire new pan law system to libxmp or I get stuck performance testing the Pack-Ice depacker again or something).

sezero:
I'm lacking in knowlegdge to review this further, so merge when you feel it's readey unless someone else points at some issue.

AliceLR:
OK, will do.

PR #788, opened 2024-11-29
build: don't use -W if -Wno-unused-parameter isn't supported

sezero:
otherwise it leads to unused-parameter warnings, e.g. with gcc2.95


AliceLR (approved):
This looks OK to me.

Copying this part of the GCC docs here because the order this patch adds the flags confused me at first:
> Some options, such as -Wall and -Wextra, turn on other options, such as -Wunused, which may turn on further options, such as -Wunused-value. The combined effect of positive and negative forms is that more specific options have priority over less specific ones, independently of their position in the command-line. For options of the same specificity, the last one takes effect.

PR #802, opened 2024-12-14
silence MSVC C4018, C4389, C4244 and C4267 warnings in all source tree

sezero:
The warnings disabled are `signed/unsigned mismatch` for `<`, `<=`, `>`, `>=`, `==` and `!=`, and `conversion from 'type' to 'int', possible loss of data` stuff.

After this, combined with #801, there is one warning from MSVC that remain:

```
loaders\sample.c(75,12): warning C4245: '*=': conversion from 'int' to 'uint8', signed/unsigned mismatch
```
This is for `p[i] *= -1` where p is an `uint8*`: Anything to do?

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -118,6 +118,11 @@ typedef int tst_uint64[2 * (8 == sizeof(uint64)) - 1];
 
 #ifdef _MSC_VER
 #pragma warning(disable:4100) /* unreferenced formal parameter */
+#pragma warning(disable:4389) /* signed/unsigned mismatch ( <, <=, >, >= ) */
+#pragma warning(disable:4018) /* signed/unsigned mismatch ( ==, != ) */
+#pragma warning(disable:4761) /* integral size mismatch in argument; conversion supplied (for MSVC6 and older.) */
+#pragma warning(disable:4244) /* conversion from 'type' to 'int', possible loss of data */
+#pragma warning(disable:4267) /* conversion from 'size_t' to 'type', possible loss of data */
 #endif
 
 #ifndef LIBXMP_CORE_PLAYER
--- a/src/depackers/bunzip2.c
+++ b/src/depackers/bunzip2.c
@@ -68,6 +68,10 @@ struct bwdata {
   unsigned int *dbuf;
 };
 
+#if defined(_MSC_VER)
+#pragma warning(push)
+#pragma warning(disable:4324) /* structure was padded due to alignment specifier */
+#endif
 // Structure holding all the housekeeping data, including IO buffers and
 // memory that persists between calls to bunzip
 struct bunzip_data {
@@ -99,6 +103,9 @@ struct bunzip_data {
   /* libxmp: For I/O error handling */
   jmp_buf jmpbuf;
 };
+#if defined(_MSC_VER)
+#pragma warning (pop)
+#endif
 
 /* libxmp addition */
 struct bunzip_output {
--- a/src/loaders/sample.c
+++ b/src/loaders/sample.c
@@ -66,13 +66,13 @@ static void convert_7bit_to_8bit(uint8 *p, int l)
 static void convert_vidc_to_linear(uint8 *p, int l)
 {
 	int i;
+	int8 amp;
 	uint8 x;
 
 	for (i = 0; i < l; i++) {
 		x = p[i];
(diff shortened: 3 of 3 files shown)


sezero:
> ```
> depackers\bunzip2.c(100,17): warning C4324: 'bunzip_data': structure was padded due to alignment specifier
> ```
> 
> I don't think we should do anything about that. But we can add a pragma to silence it?

OK silenced it in bunzip.c using a pragma

AliceLR:
> ```
> loaders\sample.c(75,12): warning C4245: '*=': conversion from 'int' to 'uint8', signed/unsigned mismatch
> ```
> 
> This is for `p[i] *= -1` where p is an `uint8*`: Anything to do?

I will look into it; there should be a more obvious way to rephrase this line which is sign-friendly.

AliceLR:
> > I don't think we should do anything about that. But we can add a pragma to silence it?
> 
> OK silenced it in bunzip.c using a pragma

I'm not sure I understand why it's complaining about this and not the 5 bajillion other places every C program in existence is using struct padding...

AliceLR (approved):
In order:
```
#pragma warning(disable:4389) /* signed/unsigned mismatch ( <, <=, >, >= ) */
#pragma warning(disable:4018) /* signed/unsigned mismatch ( ==, != ) */
```
I looked into the GCC version of this warning a few years ago; these are pretty much unfixable in libxmp's current state. Many, many places rely on signed/unsigned comparison. I've been trying to avoid adding new ones but a few places are tricky about it, to put it lightly. Like silencing `-Wsign-compare` in configure.ac, this is the correct call for the foreseeable future.

```
#pragma warning(disable:4761) /* integral size mismatch in argument; conversion supplied (for MSVC6 and older.) */
#pragma warning(disable:4244) /* conversion from 'type' to 'int', possible loss of data */
#pragma warning(disable:4267) /* conversion from 'size_t' to 'type', possible loss of data */
```
libxmp relies a lot on C-style type narrowing I assume (I haven't looked into this as much), so I think this is okay for now too. Since these are just explicit casts they might be easier to deal with someday.

AliceLR:
> I will look into it; there should be a more obvious way to rephrase this line which is sign-friendly.

Is this adequate to fix the warning?
```diff
diff --git a/src/loaders/sample.c b/src/loaders/sample.c
index 1817547a..72755261 100644
--- a/src/loaders/sample.c
+++ b/src/loaders/sample.c
@@ -67,12 +67,12 @@ static void convert_vidc_to_linear(uint8 *p, int l)
 {
 	int i;
 	uint8 x;
+	int8 amp;
 
 	for (i = 0; i < l; i++) {
 		x = p[i];
-		p[i] = vdic_table[x >> 1];
-		if (x & 0x01)
-			p[i] *= -1;
+		amp = vdic_table[x >> 1];
+		p[i] = (uint8)((x & 0x01) ? -amp : amp);
 	}
 }
 
```

sezero:
> > > I don't think we should do anything about that. But we can add a pragma to silence it?
> > 
> > 
> > OK silenced it in bunzip.c using a pragma
> 
> I'm not sure I understand why it's complaining about this and not the 5 bajillion other places every C program in existence is using struct padding...

I guess because the jmp_buf struct has an align-to-16 directive for x64

sezero:
> > I will look into it; there should be a more obvious way to rephrase this line which is sign-friendly.
> 
> Is this adequate to fix the warning?
> 
> ```diff
> diff --git a/src/loaders/sample.c b/src/loaders/sample.c
> index 1817547a..72755261 100644
> --- a/src/loaders/sample.c
> +++ b/src/loaders/sample.c
> @@ -67,12 +67,12 @@ static void convert_vidc_to_linear(uint8 *p, int l)
>  {
>  	int i;
>  	uint8 x;
> +	int8 amp;
>  
>  	for (i = 0; i < l; i++) {
>  		x = p[i];
> -		p[i] = vdic_table[x >> 1];
> -		if (x & 0x01)
> -			p[i] *= -1;
> +		amp = vdic_table[x >> 1];
> +		p[i] = (uint8)((x & 0x01) ? -amp : amp);
>  	}
>  }
>  
> ```

Yes: amended.

Now, wit this and #801 combined, MSVC builds seem warning-free.

AliceLR:
OK, merging then..

PR #803, opened 2024-12-15
Fix defects found by libFuzzer, holiday edition.

AliceLR:
* The recent Digital Tracker patch introduced a couple of signed overflows, which I've fixed.
* Very long OctaMED modules (25+ days) have a duration that breaks float→int conversion in the module scan. I've made the module scan clamp the calculated duration to `INT_MAX` for now.
* !Tracker/MUSX sample loops were not being sanitized, causing signed overflows.
* IT NNAs plus some bad bounding of instrument global volume and sample global volume could cause the mix functions to signed overflow. The former is not fixable, but I've added clamps for the global volume fields to make this slightly harder (and libFuzzer of course has already found other ways to make modules this loud...).
* Real Tracker allows negative vibrato rate, though you can only set it with a hex editor. I fixed the LFOs to allow negative rates and tweaked the Real Tracker vibrato parameters here.
* MED4 has two defects related to synths: the seek after loading a synth could signed overflow, and hybrids were missing their seek, which allowed the sample count scan and sample load to attempt to load different numbers of samples. The latter bug caused out-of-bounds heap writes.
* DBM had a bug in its pattern bounding that would allow it to load an event's effect but not convert it to libxmp effects. libFuzzer put effect `FX_ICE_SPEED` in a module and it crashed with a division-by-zero during playback.
* Megatracker could signed overflow attempting to load track data if the track offset was too large.
* IT (and probably S3M, XM...) Mod Plugin ADPCM samples could cause large sample allocations before failing when loading truncated ADPCM samples. I overhauled the way truncated sample shortening is implemented to more accurately handle normal samples and to handle ADPCM samples *at all*, so these truncated ADPCM samples properly load now.

~~NOTE: draft because the Real Tracker test fails on Le Potato for some reason, but feel free to review.~~ As usual, these are best reviewed as individual commits.


AliceLR:
The issue is `char` is `unsigned` by default on some platforms, including Debian 12 AArch64, and the Real Tracker loader left the byte fields that rely on signedness as `char` instead of replacing them with `int8`.

sezero (approved):

PR #804, opened 2024-12-16
cmake: remove disabling of MSVC warnings

sezero:
we're disabling them in common.h already


sezero:
(I had noticed this late, which was the reason why builds through cmake and nmake had different population of warnings. better late than never..)

AliceLR (approved):

PR #805, opened 2024-12-18
Do not apply FT2 bugs quirk to Modplug Tracker XMs.

AliceLR:
The libxmp test module "dans la rue" by xylo/cerror relies on E60 not setting the pattern break position in order 18, but libxmp was incorrectly applying this quirk (due to Modplug Tracker detection being relatively new), causing libxmp to skip half of order 19. I can't speak for the other bugs represented by this quirk but I highly doubt they are supported by old MPT either.


sezero:
We already clear QUIRK_FT2BUGS at line 895 for is_mpt_116 case: where is it set again, libxmp_set_player_mode??

AliceLR:
> We already clear QUIRK_FT2BUGS at line 895 for is_mpt_116 case: where is it set again, libxmp_set_player_mode??

That is for the *very* old Modplug Tracker versions which failed to accurately forge a Fast Tracker signature, I think. Most Modplug Tracker modules I have include the "correct" tracker signature string.

sezero (approved):

AliceLR:
(To clarify for reference: most original Modplug Tracker modules are passing through the case at 868 and only get disambiguated later when libxmp checks for the MPT extension section.)

sezero:
OK

PR #806, opened 2024-12-18
Fix some misinformation in the DTM loader comments.

AliceLR:
* Digital Tracker versions 2.015 and 2.02 (and most likely the missing 2.025) supported DTM as well, despite the UI advertising it as "MOD". I let the UI confuse me when I wrote these comments despite the 2.015 digi_tal.txt clearly stating that it is the first version with the new format.
* I forgot to test effect Fxx in earlier versions of Digital Tracker. 1.01 through 2.02 share the behavior of 2.03 and the beta acts more like a VBlank MOD tracker (and since it only outputs M.K., VBlank detection should support it).

> Version 2.15
> 
> DEMO KIT of DIGIPLAY and DIGITAL TRACKER F030 by MC JEE from KAMIKAZES
> 
> Toulouse 23/2/94 France
> 
> Digital Tracker
> 
> Hello, I'm put this intermediate version because of new format in
> export for the modules, instruments,...
> This format is funny and very powerful !!

If I had bothered to re-read my own research this might have been caught :) https://github.com/sezero/mikmod/pull/59#issuecomment-933356462

### DIFF
--- a/src/loaders/dt_load.c
+++ b/src/loaders/dt_load.c
@@ -41,7 +41,7 @@
  * they were just picked so the "commercial" formats with VERS would
  * have higher values than those determined by the pattern format.
  * Note that commercial versions 1.0 through 1.9 come after 2.04, and
- * versions prior to 2.03 did not support the DTM format. VERS and
+ * versions prior to 2.015 did not support the DTM format. VERS and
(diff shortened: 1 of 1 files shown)


AliceLR:
(New missing coverage line will be covered by a loop mode test module...)

sezero (approved):

PR #820, opened 2024-12-31
Fix mixer uninitialized reads caused by truncated samples.

AliceLR:
The new sample truncation code would allow some extra bytes that were not frame-aligned to exist past the end of the sample and before the extra interpolation bytes, which would never be properly initialized for stereo samples. These bytes are now properly removed.

I also merged the other two IT truncated fuzzer tests.

Didn't bump the copyright since it's still technically 2024 here...


sezero:
> Didn't bump the copyright since it's still technically 2024 here...

🤣

sezero (approved):

PR #827, opened 2025-01-02
Empty sample in oxm modules

neumatho:
The Four Ages.oxm module have a sample which is Ogg Vorbis packed. When depacking it, it result in a sample with 0 length. Because LibXmp check the result after depacking, which is the length of the depacked sample or negative if failed as <= 0, the load fails. This fix will make sure the module can be loaded correctly.

The realloc change is at least needed on Windows. If not, Xmp.exe just exit.

[Four Ages.zip](https://github.com/user-attachments/files/18288539/Four.Ages.zip)

### DIFF
--- a/src/loaders/xm_load.c
+++ b/src/loaders/xm_load.c
@@ -422,13 +422,13 @@ static int oggdec(struct module_data *m, HIO_HANDLE *f, struct xmp_sample *xxs, 
 		return -1;
 	}
 
-	if ((xxs->flg & XMP_SAMPLE_16BIT) == 0) {
+	if ((xxs->flg & XMP_SAMPLE_16BIT) == 0 && n > 0) {
 		uint8 *pcm = (uint8 *)pcm16;
 
 		for (i = 0; i < n; i++) {
 			pcm[i] = pcm16[i] >> 8;
 		}
-		pcm = (uint8 *)realloc(pcm16, n == 0 ? 1 : n);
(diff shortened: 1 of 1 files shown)


src/loaders/xm_load.c
```diff
@@ -417,7 +417,7 @@ static int oggdec(struct module_data *m, HIO_HANDLE *f, struct xmp_sample *xxs,
 	n = stb_vorbis_decode_memory(data, len, &ch, &rate, &pcm16);
 	free(data);
 
-	if (n <= 0 || ch != 1) {
+	if (n < 0 || ch != 1) {
 		free(pcm16);
 		return -1;
```
sagamusix, line 422: It would probably make a lot more sense to just return 0 here in the case of n==0. That would also avoid the special case for realloc.

AliceLR: Seconded, I think something like
```C
if (n < 0 || ch != 1) {
  [...]
} else if (n == 0) {
  xxs->len = 0;
  free(pcm16);
  return 0;
}
```
ought to be safe and is consistent with how zero length samples are handled elsewhere. `stb_vorbis_decode_memory` *will* allocate a buffer even if the decoded length is 0.

neumatho: I have now changed the code as @AliceLR suggest.

AliceLR (approved):

PR #831, opened 2025-02-08
Moved the check for edit history out of the midi configuration part

neumatho:
When making a check for usage of DSP effects/VST plug-ins, I found out that some modules do have edit history but not midi configuration. Then the file pointer was at the wrong place when I needed to do my extra check.

Currently, it does not make any difference for Xmp, since it make a seek when loading instruments/samples. But if you want in the future to support some of the extra OpenMpt features, this fix will help :-)

An example module is Digital loop by Saga Musix (https://modland.com/pub/modules/Impulsetracker/Saga%20Musix/digital%20loop.it)

### DIFF
--- a/src/loaders/it_load.c
+++ b/src/loaders/it_load.c
@@ -1275,13 +1275,14 @@ static int it_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	for (i = 0; i < mod->pat; i++)
 		pp_pat[i] = hio_read32l(f);
 
+	/* Skip edit history if it exists. */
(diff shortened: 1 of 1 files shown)


AliceLR (approved):
Trying to support VSTs outside of the tracker is a bad idea, but if it helps, alright I guess...

sagamusix:
FWIW, these structures are also used for built-in DSP effects (e.g. emulation of legacy DMO effects, or the DigiBooster echo DSP) which share the same "plugin" system as VSTs; so indeed it would be useful to do this correctly to support those in the future, if libxmp ever gets a DSP system.

PR #838, opened 2025-04-19
Fix pattern loop effect in Astroidea XMF

neumatho:
The Astroidea XMF module Astro.xmf (attached) uses the pattern loop effect twice at position 16 / pattern 15. Without this fix, the first loop will play correctly but the last one only play once. Besides that, Xmp think there are two sub-songs / sequences.

The fix make sure that the E60 effect is stored as an event, which indicate the start of the pattern loop.

[Astro.zip](https://github.com/user-attachments/files/19820378/Astro.zip)

### DIFF
--- a/src/loaders/xmf_load.c
+++ b/src/loaders/xmf_load.c
@@ -211,7 +211,6 @@ static void xmf_translate_effect(struct xmp_event *event, uint8 effect, uint8 pa
 		switch (param >> 4) {
 		case 0x01:		/* fine slide up */
 		case 0x02:		/* fine slide down */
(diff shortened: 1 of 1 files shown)


AliceLR:
Looks OK at a glance, will verify tomorrow.

sezero:
Any luck ?

AliceLR:
Sorry, got sidetracked with other issues. Looking into it now.

AliceLR (approved):
Alright, confirmed that XMFPLAY at the very least contains an implementation of this effect. Presumably it was implemented at some point after Imperium Galactica (I don't think support for the old broken behavior is necessary).

PR #841, opened 2025-04-30
vorbis: sync with stb_vorbis-sezero branch in sezero's fork stb.

sezero:
(sample-accurate-tell stuff excluded)

### DIFF
--- a/src/loaders/vorbis.c
+++ b/src/loaders/vorbis.c
@@ -71,11 +71,6 @@
 #define STB_VORBIS_C
 #include "vorbis.h"
 
-#ifdef _MSC_VER
-#pragma warning(disable:4456) /* shadowing (hides previous local decl) */
-#pragma warning(disable:4457) /* shadowing (hides function parameter.) */
-#endif
-
 //////////////////////////////////////////////////////////////////////////////
 //
 //  HEADER BEGINS HERE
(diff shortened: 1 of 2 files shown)


src/loaders/vorbis.c
```diff
@@ -1391,7 +1396,6 @@ static uint8 get8(vorb *z)
    return c;
    }
    #endif
-   return 0;  /* silence warnings */
```
AliceLR, line 1394: Are you sure about these? They come from [here](https://github.com/libxmp/libxmp/commit/bf129468db9cc0840e3ea61f50e54f07dcf4dcb5), and are likely for old compilers (guessing circa GCC 4.0).

(The `calloc`s look like an old Claudio thing with little commit documentation. Since I'm fuzzing the seek functions currently I will watch to see if those cause any unusual behavior after merge.)

sezero: I thought watocm was the only oddball here and had added warning disablement to makefike,
but you turned out to be right. gcc-4.0 emits the following at -O3. Put them back now.

````
src/loaders/vorbis.c: In function 'vorbis_find_page':
src/loaders/vorbis.c:4660: warning: control may reach end of non-void function 'get8' being inlined
src/loaders/vorbis.c:4662: warning: control may reach end of non-void function 'libxmp_vorbis_get_file_offset' being inlined
src/loaders/vorbis.c:4714: warning: control may reach end of non-void function 'set_file_offset' being inlined
src/loaders/vorbis.c: In function 'libxmp_vorbis_stream_length_in_samples':
src/loaders/vorbis.c:5098: warning: control may reach end of non-void function 'getn' being inlined
````

AliceLR (approved):

PR #843, opened 2025-05-03
Fix loading of Extreme Pinball modules

neumatho:
The Extreme Pinball modules in Epic MegaGames MASI format could not be played at all. Xmp just stopped immediately. This fix will make Xmp to load these module correctly.

The problem was the parsing of the subchunks in the SONG chunk.

Link to modules:
https://modland.com/pub/modules/Epic%20Megagames%20MASI/Robert%20Allen/


sezero (approved):
Looks good to me. @AliceLR ?

AliceLR (approved):
Verified it does fix all 5 Extreme Pinball modules I have locally. Thanks!

PR #845, opened 2025-05-03
Fix bugs in position seek functions and clarify API seek edge cases.

### LINKED ISSUE #839: Possible Infinite Loop in control.c
Hello, we have been doing some fuzzing for our University Course and chose your library as a target. We did not find any crashes during our campaigns (👏), but we found an infinite loop. The harness, the input files, and details about the setup that we used to create this issue can be found here: https://github.com/James-sys/CS5963-fuzzing-libxmp/blob/main/harness_2.c. 
When calling xmp_next_position() followed by xmp_prev_position(), we were able to get stuck in a while loop on line 91 of src/control.c. There seems to be a specific set of values of the variables pos, dir, and start that prevents continuing past this while loop. We caused this issue using both .s3m and .it file types, but we don’t know if these are the only file types that can cause this issue.

AliceLR:
+ Fix crashes when using `xmp_set_position` to set a negative position. This now returns `-XMP_ERROR_INVALID` instead.
+ Fix crashes when using `xmp_set_row` to set a negative row. This now returns `-XMP_ERROR_INVALID` instead.
+ Fix hangs when `xmp_prev_position` is used on the first position of a sequence which is preceded by an S3M/IT skip marker.
+ Fix hangs when seeking to S3M/IT end-of-module markers assigned to a non-existent sequences.
+ Fix out-of-bounds reads past the end of `mod->xxo` when `xmp_next_position` is used at the end of a 256 position module.
+ Added `xmp_next_position` regression test.
+ Added `xmp_set_position`, `xmp_next_position`, `xmp_prev_position`, `xmp_set_row`, and `xmp_seek_time` calls to libxmp_fuzz.c.
+ Added `xmp_set_position`/`xmp_next_position`/`xmp_prev_position` interactions with `xmp_stop_module`/`xmp_restart_module` and -1 return value bugs for `xmp_set_position`/`xmp_next_position` to libxmp.rst. `xmp_prev_position` return value of -1 bug was previously fixed in libxmp 4.2.0; the other two have been around so long I just documented them instead of fixing them.

Fixes #839.

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -1,6 +1,21 @@
 Stable versions
 ---------------
 
+4.6.3 (?):
+	Changes by Alice Rowan:
+	- Fix crashes when xmp_set_position is used to set a negative position.
+	  Negative positions now correctly return -XMP_ERROR_INVALID.
+	- Fix crashes when xmp_set_row is used to set a negative row.
+	  Negative positions now correctly return -XMP_ERROR_INVALID.
+	- Fix hangs when xmp_prev_position is used on the first position of
+	  a sequence which is preceded by an S3M/IT skip marker.
+	- Fix out-of-bounds reads when xmp_next_position is used at the end of
+	  a 256 position module.
+	- Fix hangs when seeking to an end-of-module marker caused by these
+	  positions getting assigned a non-existent sequence.
+	- Document xmp_set_position/xmp_next_position/xmp_prev_position
+	  interactions with xmp_stop_module/xmp_restart_module.
+
 4.6.2 (20250224):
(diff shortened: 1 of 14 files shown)


AliceLR:
I fuzzed these changes for a bit and encountered no new bugs outside of what my old corpus had already found. The "API" tag might not be appropriate since this isn't an API change, just documenting some edge nonsense that has existed for over a decade at this point.

AliceLR:
Marking as a draft while the tests run since I need to double check the hang files that the reporter uploaded, will mark ready soon.

sezero (approved):

PR #852, opened 2025-05-08
Remove remaining loader pattern loop integer division usage.

### LINKED ISSUE #400: Usage of integer division.

AliceLR:
The Oktalyzer and SFX loaders are the only remaining loaders that needlessly generate `idiv` instructions, so I've rewritten both to instead use an inner row loop. These are the last two "easy" `idiv` removals, as the rest are necessary for instrument/effects conversion (Startrekker, Digital Tracker, AMOS Music Bank, etc.) or are embedded in the player code and Paula mixers. The player divisions are extremely hard to miss and can be dealt with some other time.

Closes #400.

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -16,6 +16,7 @@ Stable versions
 	- Document xmp_set_position/xmp_next_position/xmp_prev_position
 	  interactions with xmp_stop_module/xmp_restart_module.
 	- Fix stack underflow in Pha Packer loader (CVE-2025-47256).
+	- Slight performance improvements for the Oktalyzer and SFX loaders.
 	Changes by Thomas Neumann:
 	- Fix the pattern loop effect in Astroidea XMF loader.
 	- Fix loading of Extreme Pinball modules.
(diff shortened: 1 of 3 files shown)


sezero:
A changelog entry, and updated copyright years?

sezero (approved):

PR #860, opened 2025-05-10
Fix loading XMs with harmless pattern truncation on event bounds.

### LINKED ISSUE #855: FastTracker 2 modules cannot be loaded
Found some FastTracker 2 modules, which cannot be loaded. It's atomic subculture, luvdup.xm and phuture.xm from Sonic (UK). You can find them here:

https://modland.com/pub/modules/Fasttracker%202/Sonic%20(UK)/

It seems OpenMPT can load them.

AliceLR:
Some XMs from an unknown source (possibly a converter) have patterns that end on an event boundary, typically near the end of the module. Presumably, this was intended to slightly reduce the size of patterns by omitting trailing blank events. This may also have just been a bug. These are now treated as valid.

Fixes:
* Balrog/f0rtify.xm
* Decayer-9/purification.xm
* Falcon (PL)/eaten vinyl.xm
* Headcrasher/microcosm.xm
* Jazztiz/ta-da-da-da.xm
* Jisemdu/smile.xm
* Markus Plomgren/cool jazzy jeff!!!.xm
* Orange/optical.xm
* Skyraver/spirit of life.xm
* Sonic (UK)/atomic_subculture.xm
* Sonic (UK)/luvdup.xm
* Sonic (UK)/phuture.xm
* Teemu/speed.xm
* Warhawk/anaconda.xm

(Regression caused by commit b320d7af4dea01c1c47a370630a81e096ef4b723.)

Closes #855.


src/loaders/xm_load.c
```diff
@@ -112,10 +112,21 @@ static int load_xm_pattern(struct module_data *m, int num, int version,
...
+			/* Some XMs have cleanly truncated patterns. See:
+			 * Balrog/f0rtify.xm; Decayer-9/purification.xm;
+			 * Falcon (PL)/eaten vinyl.xm; Headcrasher/microcosm.xm;
+			 * Jazztiz/ta-da-da-da.xm; Jisemdu/smile.xm;
+			 * Markus Plomgren/cool jazzy jeff!!!.xm;
+			 * Orange/optical.xm; Skyraver/spirit of life.xm;
+			 * Sonic (UK)'s atomic_subculture.xm, luvdup.xm,
+			 * phuture.xm; Teemu/speed.xm; Warhawk/anaconda.xm.
+			 */
+			if ((pat - patbuf) == (ptrdiff_t)xph.datasize) {
```
sezero, line 124: `==` is enough instead of `>=`? (didn't fully follow the code)

AliceLR: Yes—the error checks everywhere else in the loop won't let it consume more than `xph.datasize`, and even if they did, the later check on `--size < 0` would error before anything is read. This check is looking specifically for the case where there's clean truncation after the end of an event but before the start of where the next event would be.

sezero: OK then

sezero (approved):

PR #862, opened 2025-05-14
Add Software Visions DMF support to the MOD loader.

AliceLR:
Adds support for the variant of MOD used with the DOS game Apocalypse Abyss and for the Software Visions 1995 Catalog. These are just ModEdit M.K. MODs with the extension .DMF, the title and sample names stripped, and the first 2108 bytes word-flipped to little endian (either as a weak obfuscatory technique or to make loading the samples slightly faster).

I was originally requested and was going to adapt the ProWizard SV.c converter for this, but after reviewing the MOD loader I realized the required changes aren't that bad. I haven't measured any perceivable change in testing/loading speed.

(There is a converter in the wild that incorrectly swaps 2104 bytes instead of 2108. I used FROZEN.DMF for the test module as a check on this in case there's a well-intentioned attempt to "correct" this in the future, since it's one of the two Apocalypse Abyss modules to use the last event of the first pattern.)


src/loaders/mod_load.c
```diff
@@ -756,6 +790,12 @@ static int mod_load(struct module_data *m, HIO_HANDLE *f, const int start)
 	}
 
 #ifndef LIBXMP_CORE_PLAYER
+	/* First pattern of Software Visions DMF is word flipped (see above). */
+	if (tracker_id == TRACKER_SOFTWAREVISIONS && i == 0) {
+		assert(mod->chn == 4);
```
sezero, line 795: If we ever hit this, we'll face unexpected aborts from the library. The cmake build system defines NDEBUG in release builds, the MSVC project files always seem to define it, but autotools and other ones don't. We should either (i) add an `--enable-debug` switch to autotools configury (too much work not related to this patch), or (ii) adopt what we do on top of vorbis.h and miniz.h more globally - e.g. in common.h. Or, (iii) you can conditionalize the assert here to DEBUG builds, do an error-return with a `D_()` message or something.

src/loaders/mod_load.c
```diff
@@ -32,12 +32,14 @@
  * - Atari Octalyser CD61 and CD81
  * - Digital Tracker FA04, FA06 and FA08
  * - TakeTracker TDZ1, TDZ2, TDZ3, and TDZ4
+ * - Software Visions DMF (ModEdit M.K. with first 2108 bytes flipped)
```
sezero, line 35: Update copyright year at top

src/loaders/mod_load.c
sezero: (if it were me, I'd lean to the 3rd option..)

AliceLR: I forgot we don't really support this right now; I replaced it with reusing the pattern length and a comment instead.

sezero: P.S.: Not worth bailing out `if (mod->chn != 4)`?

AliceLR: > P.S.: Not worth bailing out `if (mod->chn != 4)`?

Nope, that was just to guard the 1024. It uses the calculated pattern length now so it's safe.

sezero (approved):
OK with me after removing the now-unnecessary assert.h include

PR #874, opened 2025-07-10
depacker: read stdout before cmd exit

### LINKED ISSUE #873: unrar hangs when playing RAR
I'm trying to play RAR-compressed modules, and finding that `xmp` hangs forever when trying to play them. I'm using unrar 7.11 freeware installed via Homebrew. I've tried manually calling [the same command](https://github.com/libxmp/libxmp/blob/75d334a3cf5c0b425fe71a2b07996db659149dac/src/depackers/depacker.c#L306-L316) that libxmp executes and found that it prints the module to stdout as expected. Maybe something about the way libxmp is calling it is causing a deadlock that leaves the `unrar` process to just hang forever?

I'm using xmp 4.2.0 linked against libxmp built from 75d334a on macOS 15.5.

It seems to happen with any `.rar` file I try, but here's a specific example: https://archive.org/download/jp-mod-music/20150524%20modules%20in%20JP.zip/modules%20in%20JP%2FР╘ВвЬaРп%2F2001%2Fempathyit.rar

mistydemeo:
The unrar process won't exit while it's still waiting to write content to stdout, so waiting for the process to exit first will cause libxmp to hang indefinitely.

Fixes #873.

### DIFF
--- a/src/depackers/depacker.c
+++ b/src/depackers/depacker.c
@@ -173,6 +173,17 @@ static int execute_command(const char * const cmd[], FILE *t)
 		exit(errno);
 	}
 	close(fds[1]);
+
+	if ((p = fdopen(fds[0], "rb")) == NULL) {
+		D_(D_CRIT "failed fdopen");
+		close(fds[0]);
+		return -1;
+	}
+
+	while ((n = fread(buf, 1, BUFLEN, p)) > 0) {
+		fwrite(buf, 1, n, t);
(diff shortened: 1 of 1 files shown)


sezero (approved):
Looks OK to me. @AliceLR ?

AliceLR (approved):

AliceLR:
Thanks!

PR #876, opened 2025-08-07
cmake, android: support Google Play 16 kiB page size requirement

sezero:
See https://developer.android.com/guide/practices/page-sizes

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -170,6 +170,11 @@ if(BUILD_SHARED)
         endif()
     endif()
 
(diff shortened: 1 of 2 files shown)


sezero:
OK, canning this. Users should use ndk 28c

AliceLR:
This patch does seem reasonable, though—it is essentially what asie added to MegaZeux so it can stay on r23 until the next minor version.

sezero:
AFAIK, the linker flag is for 64 bit targets only.

Let the users take care of things. SDL reverted the similar changes too.

PR #880, opened 2025-09-25
Add missing LIBXMP_BEGIN_DECLS and LIBXMP_END_DECLS to headers.

AliceLR:
I recently had a use case that involved directly linking a few compilation units from libxmp, but as the test runner was C++, it failed to find the functions declared in hio.h. Since this is *sometimes* useful, I added the missing macros to fix this in most places it could potentially be useful.


src/format.h
```diff
@@ -92,6 +92,8 @@ extern const struct format_loader libxmp_loader_stc;
 #define NUM_FORMATS 3
 #endif
 
+LIBXMP_BEGIN_DECLS
+
```
sezero, line 96: Function-only, but not data? (There are possibly other similar places, so, same for them too.)

AliceLR: Not sure what I was thinking there, will fix in a little bit.

PR #888, opened 2025-10-19
CMake: Win32: fix installation of .cmake scripts on MinGW

carlo-bramini:
After compiling the source codes of libxmp, installing on MinGW will put the scripts for CMake in a wrong position.

immagine
Actually they should be installed into the `lib` directory instead.
Attached patch fixes this issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -301,7 +301,7 @@ add_subdirectory(docs)
 
 include(CMakePackageConfigHelpers)
 
-if(WIN32)
+if(WIN32 AND NOT MINGW)
     set(cmake_install_cmakdir "cmake")
     set(cmake_install_cmakdir_lite "cmake")
 else()


sezero (approved):
LGTM. @AliceLR, @madebr?

AliceLR (approved):

AliceLR:
> P.S.: lite/CMakeLists.txt version needs the same treatment

```
$ cmake ../lite
CMake Error at CMakeLists.txt:7 (include):
  include could not find requested file:

    C:/a/libxmp/lite/cmake/libxmp-checks.cmake
```
Is the lite CMakeLists.txt etc. intended to overwrite the regular files/only be run from a distribution package? That seems suboptimal...

(Also noticed we aren't CI testing libxmp-lite's CMake buildsystem—would be worth adding explicit lite tests to all of the regression tests at some point.)

sezero:
> Is the lite CMakeLists.txt etc. intended to overwrite the regular files/only be run from a distribution package? That seems suboptimal...

Yes, old sad story.. That's why we made it possible to build lite package from top-level.

> (Also noticed we aren't CI testing libxmp-lite's CMake buildsystem—would be worth adding explicit lite tests to all of the regression tests at some point.)

Indeed.

PR #890, opened 2025-10-31
ice_unpack.c: fix warning: integer constant is too large for 'long' type

sezero:


AliceLR:
Which compiler(s) reproduce this warning? I never encountered it during my testing and it didn't seem to be in the CI.

AliceLR (approved):
This looks good to me—I made one small tweak (bumped the `ICE_UINT64_C` define over to ice_unpack.h to be closer to its corresponding typedefs).

AliceLR:
> gcc 4.4 in my case when targeting i386.
> 
> Happens with old gcc when targeting ILP32. New gcc seems to warn only with `-std=gnu89 -Wpedantic`

Was able to reproduce both with GCC 4.4.7 and with newer GCC with those flags. I'm not sure why these didn't come up in my previous testing with GCC 4.4.7 (via CentOS 6) but the original 16-entry bitplanes table might have had `ull`. I will keep this in mind in the future for 64-bit integer lookup tables.

PR #913, opened 2025-11-25
Fix MED synthetic instrument waveform table speeds.

AliceLR:
libxmp was running waveform tables at the same speed at the volume table instead of using the waveform table speed. I don't know how this possibly could have been missed, but both MED synth tests heavily rely on these two speeds being separate, hence the huge diff. I did verify that the new behavior is correct for these modules.

I standardized the MED synth mixer data tests so I could add a third one and regenerate the first two. These tests are a bit different from the regular mixer tests (mainly, they test the sample number, something the regular mixer tests don't bother with right now).


AliceLR:
Test failures are just the missing OpenWatcom snapshots.

PR #927, opened 2025-12-07
Simplify read_event_ft2 sample default volume/pan handling.

AliceLR:
Sample default volume is implemented in a strange way in all of the `libxmp_read_event` implementations. This needs to be cleaned up, as it's a significant part of why these functions are a headache.

This patch tidies up the handling for `read_event_ft2` ONLY.

Summary of the old behavior (headache warning):

* `use_ins_vol = 0`
* VALID INSTRUMENT NUMBERS ONLY:
  - If `!IS_VALID_NOTE(key - 1)`, retrieve the previous valid subinstrument and use it to apply default volume [A]. This does not affect `use_ins_vol`, but uses `SET(NEW_VOL)`.
* Any instrument number unaccompanied by a note off KEY (note that the loader translates these to `XMP_KEY_FADE`):
  - General instrument setup, set `use_ins_vol = 1`. This does not update the channel instrument in several situations, including INVALID INSTRUMENT NUMBERS. If an invalid instrument number accompanies a note, this clears the current note, and is irrelevant here.
* VALID INSTRUMENT NUMBERS ONLY:
  - If `IS_VALID_NOTE(key - 1)`, get a new subinstrument using the new instrument and new key and use it to apply default volume [B]. In this case, if there is no subinstrument, use 0 [C] (this case should ALSO cut--new test). Neither case affects use_ins_vol, but both use `SET(NEW_VOL)`.
  - This and the previous valid case account for all valid instrument numbers, so the remaining cases are all for invalid instrument numbers.
* Various cases relating to `XMP_KEY_OFF`: this is irrelevant to sample default volume, as instrument + off key events are translated to `XMP_KEY_FADE` instead (see above). This `use_ins_vol = 0` does nothing.
* If key is set, most `new_invalid_ins` cases will call `libxmp_virt_resetchannel`, making volume changes irrelevant. The one case this doesn't happen in is toneporta, where it's supposed to set the old instrument's volume [D]. This case clears key, preventing the next case from occurring.
* Various cases relating to `IS_VALID_NOTE(key - 1)` etc. None of the invalid instrument cases matter here because they all already reset the channel, cleared key [D], or don't have a valid note to begin with [E].
  - `xc->key` is only finally set here.
  - The `sub == NULL` case for clearing `use_ins_vol` is redundant with the later `sub == NULL` early return.
  - The `new_invalid_ins` case for clearing `use_ins_vol` is redundant with the `libxmp_virt_resetchannel`.
* The final subinstrument load is done from `xc->ins` and `xc->key`, neither of which have been changed in invalid instrument cases.
* Effects are applied—anything that sets volume directly will `SET(NEW_VOL)`.
* Early return on `sub == NULL` prevents fully invalid instruments (both old number and new number are invalid) from setting volume.
* Finally, instrument volume (and no other field) is applied if `use_ins_vol && !TEST(NEW_VOL)`. The only cases that this applies to are invalid instrument with either valid note plus toneporta [D] or no valid note [E], both of which will reuse old `xc->ins` and `xc->key` values.

New behavior:

* Symbols `ins`, `use_ins_vol`, `new_invalid_ins` are fully removed in favor of attempting to actually model what FT2 does.
* Preload the subinstrument from the old instrument/key (may be invalid). This simulates retaining the old sample pointer.
* Set up the instrument as before, except don't set `xc->ins` (see changed behaviors below; doing this broke things).
* If `IS_VALID_NOTE(key - 1) && !is_toneporta`:
  - Instrument memory: `ins = ev.ins ? ev.ins : xc->old_ins`
  - If `IS_VALID_INSTRUMENT(ins - 1)` set `xc->ins`.
  - If `ev.ins` is set and not valid, or `ev.ins` is not set and the previous referenced instrument is invalid, unset the channel instrument. This is a combination of two edge cases moved from later in the function to here, where they're more relevant. Doing this allowed the removal of `new_invalid_ins`.
  - Load the subinstrument. If the instrument and subinstrument are valid, this replaces the active subinstrument [B].
  - If `IS_VALID_INSTRUMENT(ev.ins - 1)`, set `xc->ins_fade`. If the subinstrument is invalid, substitute `0x80 << 1` here (`test_player_ft2_instrument_fade_update`).
  - If the subinstrument is invalid, either from the instrument being cleared above or from the subinstrument being invalid [C], or if the sample length is zero, reset the channel (`test_player_ft2_invalid_ins_defaults`).
* If the instrument was valid but the note wasn't valid [A], the instrument was invalid but the note was valid with toneporta [D], or the instrument was invalid with no note [E]: the preloaded subinstrument remains active.
* If ANY instrument number is present in the event, load default volume and panning from the active subinstrument. If the subinstrument is invalid, substitute `0` for volume and `0x80` for panning (`test_player_ft2_invalid_ins_defaults`). Since this is performed long before effects, they will overwrite whatever this set the volume/panning to if relevant.

CHANGED BEHAVIOR:

* Fix `read_event_ft2` changing `xc->ins` in situations where a valid instrument number is used, but the old instrument is retained (`test_no_note_valid_ins_ft2`, `test_porta_valid_ins_ft2`, `test_player_ft2_note_noins_after_invalid_ins`). This would cause repeat instances of cases that should reuse the same old instrument to use volume/pan/etc. from other referenced instruments instead. This would also cause the wrong envelopes to be used (`test_player_ft2_no_note_ins_envelope`).
* Fix sample default volume not being applied for invalid instrument numbers with a key off note (`test_no_note_invalid_ins_ft2`). This was caused by `use_ins_vol = 1` being gated behind `key != XMP_KEY_FADE`.
* Fix sample default panning not being applied for invalid instrument numbers (`test_no_note_invalid_ins_ft2`, `test_porta_invalid_ins_ft2`). This was caused by the `use_ins_vol` block only setting volume.
* Fix `read_event_ft2` not reseting the envelope positions on an invalid instrument number (`test_player_ft2_envelope_invalid_ins`).
* Fix sample default volume/panning and instrument fade rate not being applied for active invalid instruments. Previously, only one case set volume to 0 (`test_player_ft2_invalid_ins_defaults`). Invalid instruments have default volume 0, default panning 0x80, and instrument fade rate 0x80.
* Fix instrument fade rate being copied to the channel in a handful of cases that weren't specifically note + ins. + no toneporta/K00.
* Fix zero length samples not cutting the current channel.

Note: a fade out bug alluded to in commit 0f4da30d was user error, but the other issues in those tests are fixed and all of the FIXMEs from that patch are resolved.


AliceLR:
> Another one of the modules cited in `read_event_ft2` (Quazar/lucky stars.xm) has render differences at cited position 5, both channels 5 and 10. The other modules are fine.

...this is caused by libxmp misinterpreting XM autovibrato waveform 3 (ramp-up) as the random waveform. If I force the waveform for this instrument to 2, there's no diff issue.

sezero:
> Saga Musix on Discord:
> 
> > Probably will be busy during all the evenings this week, so feel free to merge without me...
> 
> I'm going to go ahead and merge this.

OK

AliceLR:
The rage.xm final pattern differences were a side effect of fixing `xc->ins` getting updated at the wrong times. This ALSO made libxmp play notes with the wrong envelopes (`test_player_ft2_no_note_ins_envelope`). The new playback is correct.

PR #958, opened 2026-01-28
Fix xmp_start_player unmuting regression from 4.5.0.

AliceLR:
Prior to 4.5.0, `xmp_start_player` unconditionally unmuted every player channel and formats with mute flags were implemented by zeroing channel volume (`XMP_CHANNEL_MUTE` was informational). This was changed to make `XMP_CHANNEL_MUTE` functional in e99acd32, but this broke unmuting *non-muted* channels.

This bug can be observed by building the latest xmp-cli master (34154d12 or newer) and switching the player between two modules with different muted channels: mute values from the previous module leak into the next module.

This patch also fixes the GDM loader to not clobber channel volumes on load (the equivalent fix for IT was also in 4.5.0: df978f9).

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -100,6 +100,9 @@ Stable versions
 	- PTM: actually fix pattern break hex parameter.
 	- STM: always use 0 for pattern break parameter (ST2 ignores it).
 	- STM: import ST2 tempo values as both speed and BPM.
+	- GDM: don't zero channel volume for muted channels.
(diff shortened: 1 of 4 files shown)


AliceLR:
Fixing this might necessitate bumping the version to 4.7.0 (on the off chance anyone actually relied on mute state persisting between modules when using the same context...).

sezero (approved):

AliceLR:
> I'm not opposed to it.

I'll start updating the milestone names then...

Issue #836, opened 2025-02-27
Tempo-invariant playback time drift


mmontag:
Now that we have `xmp_set_tempo_factor`, we need a tempo-invariant playback time. We have this - or, well, it almost works.

When using `xmp_set_tempo_factor` to change playback speed, the `time` returned from `xmp_get_frame_info` drifts away from the actual playback position between patterns. It is not scaled with the tempo.  The time is only corrected/reconciled at the beginning of each pattern. 

A similar issue was recently addressed in libgme: https://github.com/libgme/game-music-emu/pull/111

AliceLR:
I will double check to make sure there won't be any problems (some formats like MED and FAR use non-standard tempo factors), but I think it should be doable.

mmontag:
Oh, hmm… I didn’t look at the implementation but I was hoping tempo factor was a player concept, multiplied on top of intrinsic song tempo, and immune to format quirks.

AliceLR:
It really ought to be, such that `1.0` = normal speed for the module regardless of format. Right now it sets the current format's tempo factor directly. The default is usually `10.0`, but `2.64` for MED in tempo mode, module-dependent for MED in BPM mode, and slightly over 4 for FAR.

@sezero opinions on fixing `xmp_set_tempo_factor`'s scale with a versioned symbol in 4.7.0?

sezero:
> @sezero opinions on fixing `xmp_set_tempo_factor`'s scale with a versioned symbol in 4.7.0?

OK

mmontag:
Closely related to #734

mmontag:
Yes, I'm asking for the more useful first option.

AliceLR:
Alright, I've implemented this in #853. Rescanning the module will still cause the new tempo factor to take effect, but the frame times are unaffected until then. Setting a fully relative player-only tempo factor can wait until libxmp 4.7.0 (probably a new function, as `xmp_set_tempo_factor` in its current form may still be useful for editing).

PR #966, opened 2026-01-30
Add API functions to set relative tempo factor, get tempo factor.

AliceLR:
* `xmp_set_tempo_factor_relative`: this function sets a relative, player-only user time factor. The module's time factor is multiplied by this value when computing the tick size. This function was implicitly requested in #836.
* `xmp_get_tempo_factor`: returns the module time factor set by loading a module or by calling `xmp_set_tempo_factor`.
* `xmp_get_tempo_factor_relative`: returns the relative time factor set by `xmp_start_player` or `xmp_set_tempo_factor_relative`.

MINGW32 currently fails the regression tests. (MINGW64 is fine.)


test-dev/test_api_set_tempo_factor.c
AliceLR: > MINGW32 currently fails the regression tests. (MINGW64 is fine.)

It's x87 as usual, with the added interesting bit that this doesn't seem to happen when using CMake on my end, only Autoconf. The bandaid fix seems to be:
```C
volatile double factor
````

sezero (approved):

PR #976, opened 2026-02-24
Add Pack-Ice to the list of depackers in README.

AliceLR:


AliceLR:
(Not clear if OXM and MUSE should remain in this list, but that can be left for another time.)

sezero (approved):

PR #981, opened 2026-03-13
Add BUILD.md, CONTRIBUTING.md, pull request template.

### LINKED ISSUE #980: Pull request template
It would probably be useful to have a pull request template, plus additional policy files as-needed, to address a few things that have mostly been unstated/implicit:

* Licencing: all new code (or old code being reactivated) should be MIT licensed. It should not be based on code with more restrictive licenses.
* Copyright assignment: make it explicit that PRs are copyright assigned to whichever copyright is listed at the top of a file, and new files should be assigned to one of the existing copyright holders (exception: contributed code like xz...).
* Reverse engineering: should be clean room (e.g. black box, wall). Closed source, "source available", unstated license, contradictory license, disassemblies, decompilations, etc. are unclean and should be avoided unless the original copyright holders gave permission (the ST2 doc) or unless the usage would fall under reverse engineering protections (FAR tempo).
* LLM policy: possibly allow reasonable uses (examples: machine translation for developer communication; small-scale autocomplete if local and trained exclusively on libxmp's MIT-licensed source files (edit: I am being assured in DM that this is actually impossible); code scanning/etc. as part of CI, if it can be [more reliable than this](https://github.com/AliceLR/megazeux/pull/574)); forbid *everything* else.
  
  Rationale: OpenMPT has received or been asked to evaluate at least 3 large patches recently containing LLM code smells, including high verbosity, strange constructs, programming errors, etc. The license of such code is effectively indeterminate and we've heard the ethics issues a zillion times. libxmp is less prominent and AFAIK has not been affected, but it's still used enough (SDL_mixer, Doom source ports) that it'd be better to nip this in the bud now and assure our downstreams that we're not going to turn into a liability e.g. chardet (see their issue 327).

I wrote [this draft](https://github.com/AliceLR/megazeuxtests/commit/9a2ed05f5b40f30f67391bc677450dc2d438073f) a while ago ([example pull request](https://github.com/AliceLR/megazeuxtests/pull/10)), mostly just to test pull request templates, but it needs work.

AliceLR:
***** This is very much a draft, currently. Feedback is welcome. *****

* This makes two implicit processes of accepting libxmp pull requests (MIT licensing and copyright assignment) explicit to the contributor.
* Add a new CONTRIBUTING.md file with explicit policies on reverse engineered code and LLM usage, and require pull requests to check a box indicating that this document has been read and followed.
  - AFAIK there's no clear statement anywhere else regarding clean contributions, hence the reverse engineering policy. Encouraging clean implementations and transparency re: which code is potentially "tainted" should also potentially reduce liability for us.
  - Recently, numerous open source projects have received large amounts of LLM slurry and OpenMPT has received large patches (one unsubmitted, two submitted and refused for merge) of dubious quality containing indicators of LLM "authorship". The chardet licensing situation was kind of the last straw for me on this. As noted elsewhere, it's not clear if this will actually prevent bad-faith LLM patches, but it's nice to have a clear position statement for our downstreams nonetheless. Saga Musix has indicated that he also intends to do something similar for OpenMPT/libopenmpt.

Closes: #980

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -1,0 +1,389 @@
+# Building libxmp
+
+This document covers build commands to compile libxmp.
+See CONTRIBUTING.md for guidelines on how to contribute patches.
+
+This document doesn't cover every configuration option for libxmp.
+See the `./configure` help text, CMakeLists.txt, or the relevant
+Makefile.(arch) for more information.
+
+
+## Autoconf/GNU Make (CLI)
+
+libxmp can be compiled with Autoconf and GNU Make, a POSIX shell, and GCC
+(or clang). This is libxmp's primary build system, and is recommended for
+development on Linux/BSD, macOS, Haiku, and others.
+
+(MSYS2 users will get better compilation times from CMake, but this build
+system is still required to regenerate the other build system Makefiles
+when developing in a git clone of libxmp and adding new compilation units
+or removing existing ones.)
+
+```sh
+# Configure libxmp
+./autogen.sh
+./configure
+
+# Build and test libxmp
+make
+make check
+
+# Run regression tests (only works when building from the Git repository)
+(cd test-dev && autoconf && ./configure)
+make devcheck
+```
+
+GNU Make supports parallel compilation via the command line option `-j#`, where
+`#` is the number of compilation workers. This can be used reduce compilation
+time. If `#` is omitted, GNU Make will spawn as many workers as possible.
+
+```sh
+# Build with 8 workers
+make -j8
+```
+
+### Debug printing
+
+To enable debug printing (warning: it is extremely verbose and should not
+be enabled for CI), edit Makefile, uncomment this line, and rebuild:
+
+```Makefile
+#CFLAGS += -DDEBUG
+```
+
+### Preparing other build systems
+
+This build system is used to regenerate the Makefiles for every other
+build system, which should be performed every time a new compilation unit
+is added:
+
+```sh
+make cmake-prepare
+make vc-prepare
+make watcom-prepare
+make emx-prepare
+```
+
+The regression tests system uses automatically constructed source file lists
+for CMake and NMAKE. Running the regression tests once with the GNU Make
+build system is sufficient to rebuild the list of test source files. Adding
+other new source files to the regression tests requires the following:
+
+```sh
+# Rebuild test-dev/Makefile.vc
+(cd test-dev && make vc-prepare)
+
+# There is currently no target to rebuild test-dev/CMakeLists.txt.
+# Base source files must be edited in manually.
+```
+
+### libxmp-lite (optional)
+
+libxmp-lite can be built alongside libxmp using the following commands.
+(If you are building libxmp-lite from its release tarball, just use the
+same instructions as above.)
+
+```sh
+./configure --enable-lite
+make lite
+make check-lite
+```
+
+### Static library (optional)
+
+Provide the `--enable-static` option to `./configure`, then build as usual:
+
+```sh
+./configure --enable-static
+```
+
+The shared library can also optionally be disabled with `--disable-shared`:
+
+```sh
+./configure --enable-static --disable-shared
+```
+
+### Installing libxmp
+
+This build system can install the library (or libraries) to the `--prefix`
+provided to `./configure` (by default, `/usr/local`).
+
+Combine with `doas` or `sudo` as-needed:
+
+```sh
+make install
+```
+
+
+## CMake (CLI)
+
+libxmp can also be built using CMake. This build system supports GCC, clang,
+and Microsoft Visual C++ (MSVC).
+
+```sh
+# Configure libxmp
+mkdir builddir
+cmake -B builddir -S . -DCMAKE_BUILD_TYPE=Release
+# When building from the Git repository, use this to enable regression tests.
+cmake -B builddir -S . -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=ON
+
+# Build libxmp
+cmake --build builddir
+
+# Test libxmp and run regression tests (if enabled)
+(cd builddir && ctest --output-on-failure)
+```
+
+For MSVC, the configuration must be provided at build time:
+
+```sh
+# Build libxmp
+cmake --build builddir --config Release
+
+# Test libxmp and run regression tests (if enabled)
+cd builddir
+ctest -C Release --output-on-failure
+```
+
+Like GNU Make, CMake supports parallel compilation via the command line
+option `-j#` to reduce compilation time:
+
+```sh
+# Build with 8 workers
+cmake --build builddir -j8
+```
+
+### Debug printing
+
+To enable debug printing (warning: it is extremely verbose and should not
+be enabled for CI), reconfigure with this build type and rebuild.
+
+```sh
+cmake -B builddir -S . -DCMAKE_BUILD_TYPE=Debug
+```
+
+For MSVC, the configuration must be provided at build time:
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+
+### GCC/Clang with Autoconf/GNU Make or CMake (CLI)
+
+libxmp can be compiled with Autoconf and GNU Make (or CMake), a POSIX shell,
+and a C compiler.
+
+Autoconf and GNU Make:
+```sh
+# Configure libxmp (Autoconf)
+autoconf
```
sezero: `./autogen.sh` (to be future-proof..)

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+./configure
+
+# Build and test libxmp (Autoconf)
+make -j8
+make -j8 check
```
sezero: `-j8` here not needed

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+
+# Run regression tests (Autoconf)
+(cd test-dev && autoconf && ./configure)
+make devcheck -j8
```
sezero: Flip positions of `-j8` and `devcheck` ?

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+can be installed independently of Visual Studio using the Visual Studio
+installer. They can be used directly via the installed MSVC command lines,
+or with an IDE that supports specifying a custom shell and compile commands.
+
+### Other buildsystems
+
+libxmp has buildsystems for Open Watcom (Makefile.dos, Makefile.os2,
+Makefile.w32, and Makefile.lnx) and EMX on OS/2 (Makefile.emx).
+
+```sh
```
sezero: The following block implies running all of those commands one after the other without a `make distclean` in between. Suggest breaking them into target/compiler items:

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+# Make and check libxmp (DOS)
```
sezero: `- Make and check libxmp for DOS using Open Watcom:`

(see above)

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+wmake -f Makefile.dos
+wmake -f Makefile.dos check
+
+# Make and check libxmp (OS/2)
```
sezero: `- Make and check libxmp for OS/2 using Open Watcom:`

(see above)

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+wmake -f Makefile.os2
+wmake -f Makefile.os2 check
+
+# Make and check libxmp (Windows)
```
sezero: `- Make and check libxmp for Windows (x86) using Open Watcom:`

(see above)

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+wmake -f Makefile.w32
+wmake -f Makefile.w32 check
+
+# Make and check libxmp (Linux)
```
sezero: `- Make and check libxmp for Linux (x86) using Open Watcom:`

(see above)

CONTRIBUTING.md
```diff
@@ -0,0 +1,337 @@
...
+wmake -f Makefile.lnx
+wmake -f Makefile.lnx check
+
+# Make and check libxmp (EMX)
```
sezero: `- Make and check libxmp for OS/2 using gcc (EMX KLIBC):`

(see above)

sezero:
In general: Should we move build instructions  (which is a welcome documentation) into own file, e.g. BUILD.txt or INSTALL.txt or COMPILE.txt or whatever??

AliceLR:
> In general: Should we move build instructions (which is a welcome documentation) into own file, e.g. BUILD.txt or INSTALL.txt or COMPILE.txt or whatever??

This would be reasonable and would give more room to document things like `-j#` (instead of putting it in the default GNU Make instructions), libxmp-lite, etc.

Should I split it from this patch, or is it fine as part of this one?

sezero:
> Should I split it from this patch, or is it fine as part of this one?

Um, your call. But I guess it'd be easier if it's done here?

sezero:
It's better now that BUILD.MD is split off

BUILD.md
```diff
@@ -0,0 +1,313 @@
...
+nmake -f Makefile.vc check-lite
+```
+
+The MSVC toolchain, command line tools, and Windows SDK (and optionally CMake)
+can be installed independently of Visual Studio using the Visual Studio
+installer. To use these tools, CMD.exe should be launched from one of the
+preconfigured shells installed by the Visual Studio installer.
+
+NMAKE doesn't support parallel compilation, but instead relies on the MSVC
+compiler to provide this via `/MP`. Makefile.vc provides this option to MSVC.
```
AliceLR: This is actually a lie right now: only the test-dev Makefile.vc supports `/MP`. It should be easy to patch the main and libxmp-lite makefiles for this, though. 

My only concern here is whether or not there's a way to make MSVC keep intermediate objects when all source files are provided in one go. The way test-dev does this right now is fine (since most of the time, they're all being rebuilt anyway), but this isn't necessarily true for the library. I can just remove this paragraph if there's no way to make this work.

AliceLR: Seems like the original way to fix this was `/Gm`, which is deprecated, produces a compiler warning, and as far as I can tell doesn't work. The replacement the Microsoft documentation vaguely alludes to looks like a VS-only project file option.

sezero: Just remove /MP from documentation, yes?

AliceLR: Yeah, I don't see any way around this aside from reimplementing dependency resolution for NMAKE... 🤢

AliceLR:
I *think* this is okay to officially mark ready for review.

(I also fixed the MSVC CI debug output issue, now that I know why it was still doing it :slightly_smiling_face:. That and/or BUILD.md can still be split off to separate patches if necessary.)

BUILD.md
```diff
@@ -0,0 +1,387 @@
...
+
+
+## Autoconf/GNU Make (CLI)
+
+libxmp can be compiled with Autoconf and GNU Make, a POSIX shell, and GCC
+(or clang). This is libxmp's primary build system, and is recommended for
+development on Linux/BSD, macOS, Haiku, and others.
+
+(MSYS2 users will get better compilation times from CMake, but this build
+system is still required to regenerate the other build system Makefiles.)
```
sezero: Matters only in git clones:

`when developing in a git clone of libxmp.)`

AliceLR: Or any time a compilation unit is being added/renamed with intent to submit a patch, but *hopefully* anyone doing that is doing that is using a cloned repository...

sezero: > Or any time a compilation unit is being added/renamed with intent to submit a patch,

Yes, but regenerating makefies (as opposed to manually editing them - which is always possible),
is only possible in git clones: relevant *.in files aren't in tarballs.

> but _hopefully_ anyone doing that is doing that is using a cloned repository...

Well, yes. Hopefully.

sezero (approved):

BUILD.md
AliceLR: I have a partially finished follow-up that moves the offending bit and developer-only commands to its own section at the end of the document, since stuff like `make utilities` should also be documented. I will submit a separate PR for it, but this change works for now.

PR #985, opened 2026-04-29
store os2/emx dll def files in revision control and in tarballs.

sezero:


genemxdef.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "DATA MULTIPLE NONSHARED" >> emx.def
+echo "DESCRIPTION \"libxmp $LIBXMP_VERSION\"" >> emx.def
+echo EXPORTS >> emx.def
+grep xmp_ libxmp.map | sed -e "s/\;//g" | sed -e "s/^[ \t]*//g" | sed -e "s/xmp_/\ _xmp_/g" | sort | uniq >> emx.def
+
+echo "LIBRARY LIBXMPLT INITINSTANCE TERMINSTANCE" > emxlite.def
+echo "DATA MULTIPLE NONSHARED" >> emxlite.def
+echo "DESCRIPTION \"libxmp-lite $LIBXMP_VERSION\"" >> emxlite.def
+echo EXPORTS >> emxlite.def
+grep xmp_ libxmp.map | sed -e "s/\;//g" | sed -e "s/^[ \t]*//g" | sed -e "s/xmp_/\ _xmp_/g" | sort | uniq >> emxlite.def
```
AliceLR: Would it make more sense to have this be a part of `emx-prepare` (since it would *probably* only be done occasionally for API changes), or is there a use case for doing this more often?

(I think getting rid of `$(SED)` etc. was the right call though, since the `*-prepare` targets all just use sed directly... 🙂)

sezero: Hm, maybe.

But I actually don't like those *-prepare stuff residing in our main
makefile: If run from tarball, they surely fail. That's why I separated
emx def stuff. 

I can do it that way? Although having to run configure for just that
seems tiresome. And similarly tiresome for all other *-prepare stuff.
Maybe we should separate those from configuration?

AliceLR: As in, adding a new Makefile that only has the includes and `cmake-prepare`/`vc-prepare`/`watcom-prepare`/EMX defs/etc.? That seems doable.

sezero: I moved def generation to emx-prepare for now. 

We can do *-prepare clean-up later.

AliceLR (approved):

PR #999, opened 2026-06-05
stb_vorbis: Increase setup_malloc() size parameter to uint64

### LINKED ISSUE #996: Integer overflow → heap buffer overflow in OGG Vorbis loader (stb_vorbis codebook multiplicands)
## Summary

In libxmp 4.7.0, `setup_malloc(int sz)` in `src/loaders/vorbis.c:988` uses `int` as its parameter type. At line 3953, the triple multiplication `sizeof(codetype) * c->entries * c->dimensions` is computed in `size_t` (64-bit), then truncated to `int32` when passed to `setup_malloc`. By crafting entries=80581 and dimensions=13325, the product becomes 0x100000004, truncated to int32=4, allocating only 8 bytes on the heap. The subsequent loop at line 3956 writes entries×dimensions float values, causing a massive heap buffer overflow (~4GB write into an 8-byte buffer).

## Root Cause

```c
// vorbis.c:988
static void *setup_malloc(vorb *f, int sz)
{
    f->setup_memory_required += sz;
    return malloc(sz > 0 ? sz : 1);
}

// vorbis.c:3953
c->multiplicands = (codetype *) setup_malloc(f,
    sizeof(codetype) * c->entries * c->dimensions);
PoC

gcc -fsanitize=address -g -o vorbis_poc vorbis_poc.c -lm -Iinclude
./vorbis_poc
ASAN Output

==14190==ERROR: AddressSanitizer: heap-buffer-overflow
WRITE of size 4 at 0x502000000058
    #0 start_decoder src/loaders/vorbis.c:3962

0x502000000058 is located 0 bytes after 8-byte region [0x502000000050,0x502000000058)
    allocated by #0 setup_malloc src/loaders/vorbis.c:999
Trigger Parameters
entries=80581, dimensions=13325, 4×entries×dims=0x100000004 → int32=4, alloc 8 bytes, writes ~4GB

Impact
CVSS 9.8 (Critical). Heap overflow → DoS, potential RCE. Similar to CVE-2019-13217~13223 (stb_vorbis).

Suggested Fix

static void *setup_malloc(vorb *f, size_t sz)

### PR DESCRIPTION
**Pull request by sezero**
Expands the setup_malloc size parameter, and various calculations
that the Vorbis spec allows to exceed 32 bits, to uint64. setup_malloc
and setup_temp_malloc still reject allocations that exceed INT_MAX,
as the underlying memory fields are still 32 bit signed integers.

Mainstream P/R: https://github.com/nothings/stb/pull/1960
Reference issue: https://github.com/libxmp/libxmp/issues/996
Also see: https://github.com/nothings/stb/issues/1947
Also see: https://github.com/nothings/stb/issues/1933
Also see: https://github.com/nothings/stb/issues/1168#issuecomment-1082625780

### DIFF
--- a/src/loaders/vorbis.c
+++ b/src/loaders/vorbis.c
@@ -646,10 +646,17 @@ enum STBVorbisError
 #define MAX_BLOCKSIZE_LOG  13   // from specification
 #define MAX_BLOCKSIZE      (1 << MAX_BLOCKSIZE_LOG)
 
-
 /* libxmp-specific change */
 #if 1
 #include "../common.h"
+#else
+
+#if defined B_BEOS_VERSION
+   #include <SupportDefs.h>
+#elif defined __amigaos4__
+   #include <exec/types.h>
+#elif defined _arch_dreamcast /* KallistiOS */
+   #include <arch/types.h>
 #else
 typedef unsigned char  uint8;
 typedef   signed char   int8;
@@ -657,7 +664,15 @@ typedef unsigned short uint16;
 typedef   signed short  int16;
 typedef unsigned int   uint32;
 typedef   signed int    int32;
-#endif
+#ifdef _MSC_VER /* MSVC6 has no long long */
+typedef unsigned __int64 uint64;
+#elif defined(_LP64) || defined(__LP64__)
+typedef unsigned long uint64;
+#else
+typedef unsigned long long uint64;
+#endif /* uint64 */
+#endif /* !BeOS && !AmigaOS4 && !KallistiOS */
+#endif /* !LIBXMP */
 
 #ifdef __has_feature
 #if __has_feature(undefined_behavior_sanitizer)
@@ -711,7 +726,7 @@ typedef struct
    uint8  lookup_type;
    uint8  sequence_p;
    uint8  sparse;
-   uint32 lookup_values;
+   uint64 lookup_values; /* entries (24 bits) * dimensions (16 bits) */
    codetype *multiplicands;
    uint32 *codewords;
    #ifdef STB_VORBIS_FAST_HUFFMAN_SHORT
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by sezero** [comment]
Alice: Hopefully I adapted this correctly.

Crazy thing is it gracefully rejects the exploit.ogg after a total allocation of 280MB memory. Sigh..

If this is good, I'll also apply this to SDL_sound and SDL_mixer, with you as the author.

File: src/loaders/vorbis.c [hunk 1]
```diff
@@ -3929,7 +3946,11 @@ static int start_decoder(vorb *f)
          } else {
             /* unsigned multiply to suppress (legitimate) warning.
              * https://github.com/nothings/stb/issues/1168 */
```
AliceLR: This comment is outdated since it's for one of the same issues, can be discarded. (May be worth adding that issue to the new comment's list?)

src/loaders/vorbis.c
```diff
@@ -985,14 +1000,15 @@ static void *make_block_array(void *mem, int count, int size)
   }
 }
 
-static void *setup_malloc(vorb *f, int sz)
+static void *setup_malloc(vorb *f, uint64 siz)
 {
-   if (sz <= 0 || INT_MAX - 7 < sz) return NULL;
+   int sz = (int)siz;
+   if (sz <= 0 || (uint64)INT_MAX - 7 < siz) return NULL;
```
AliceLR, line 1006: Minor (re: both): negatives get converted to >`INT_MAX` when passed as a `uint64`, so the handling in my original patch should be adequate. (Not sure if it affects anything in practice.)

src/loaders/vorbis.c
```diff
@@ -646,18 +646,33 @@ enum STBVorbisError
 #define MAX_BLOCKSIZE_LOG  13   // from specification
 #define MAX_BLOCKSIZE      (1 << MAX_BLOCKSIZE_LOG)
 
-
 /* libxmp-specific change */
 #if 1
 #include "../common.h"
 #else
+
```
AliceLR, line 653: Do we still need the common.h include? (Haven't checked yet..)

AliceLR (commented):
Looks mostly OK to me, will give it a better read when I can.

src/loaders/vorbis.c
sezero: OK, merged comments

src/loaders/vorbis.c
sezero: We might get type redfinition errors, otherwise.

src/loaders/vorbis.c
sezero: I preferred >= INT_MAX to be on the safe side, hence the flipped version.

I can change it though??

src/loaders/vorbis.c
AliceLR: Okay, better to leave this, then.

src/loaders/vorbis.c
AliceLR: It's probably fine, will re-read after I sleep.

AliceLR: I think this is fine for now since it's more or less what our fork had before. (If needed it can be adjusted if the upstream patch is merged.)

AliceLR (approved):

AliceLR:
> Crazy thing is it gracefully rejects the exploit.ogg after a total allocation of 280MB memory. Sigh..

libfuzzer's favorite thing for the past several years has been to find OggMod files that allocate multiple GiB of RAM via sparse codebooks and then `qsort` the generated codewords before failing due to other decoding issues. It's sadly more likely than you'd think and I don't know how to deal with it :(

PR #1000, opened 2026-06-05
Fix FT2 channel default instrument memory.

### LINKED ISSUE #989: drozerix_-_playful_girl.xm pattern 0 plays incorrectly
This is likely an oversight from the `read_event_ft2` overhaul: in [this module](https://modarchive.org/index.php?request=view_by_moduleid&query=185337), only the first four channels should play during the first repeat of the first pattern, but the later channels also play. FT2 and FT2-clone play this pattern as-intended.

Initializing `xc->old_ins` to -1 at playback start instead of 1 (?!) seems to fix it.

AliceLR:
In FT2, the initial active instrument is 0 (no instrument) or some other equivalent state. Notes with no instrument number at the start of a module should not play. They were instead playing with instrument 1, a regression caused by the new (correct) implementation of instrument memory added in 4.7.0 combined with xc->old_ins having the wrong default value for FT2.

Fixes drozerix_-_playful_girl.xm order 0 and probably others.

Fixes #989.


AliceLR:
It's actually supposed to be 0 for IT too, so now I need to investigate why libxmp was even using 1 here at all.

AliceLR:
> It's actually supposed to be 0 for IT too, so now I need to investigate why libxmp was even using 1 here at all.

Correct: IT has a default of 0. This change didn't affect IT because `read_event_it` only uses `xc->old_ins` for one check before switching to `xc->ins` for later instrument memory handling. I made the default value 0 for everything.

AliceLR:
NetBSD CI failure is from failure to fetch pkgsrc and not related to the patch.

sezero:
Can we hit https://github.com/libxmp/libxmp/blob/master/src/read_event.c#L997 with this?

sezero:
> Can we hit https://github.com/libxmp/libxmp/blob/master/src/read_event.c#L997 with this?

Ah, then `IS_VALID_INSTRUMENT(-1)` -> `(uint32)(-1) < mod->ins` being false will be our saving grace.

sezero (approved):

PR #1001, opened 2026-06-06
Fix Magnetic Fields Packer track loader, optimize tracks.

### LINKED ISSUE #872: Magnetic Fields Packer is not loaded correctly
When playing modules in Magnetic Fields Packer format, they are not played correctly. This can specially be heard in the Tower of souls - intro and title module.

The problem is that the pat_table array is used as a pattern lookup to get the offset for at pattern. This is wrong, instead it should be treated as the position list. The length of the array is the number of position * 4, one value per channel per position.

If you look at the values and compare to the real position list, you will see the values are repeated for the same patterns in the position list. I have made a fix in my C# converter (convert to ProTracker format), which is based/inspired on LibXmp implementation. You would probably make in a little bit different, since I use a Dictionary and you don't need to convert it. But here is my code for inspiration.

```
		/********************************************************************/
		/// <summary>
		/// Return a collection with all the patterns
		/// </summary>
		/********************************************************************/
		protected override IEnumerable<byte[]> GetPatterns(ModuleStream moduleStream)
		{
			ushort size1 = moduleStream.Read_B_UINT16();
			moduleStream.Seek(2, SeekOrigin.Current);

			ushort[,] patternTable = new ushort[4, size1];

			for (int i = 0; i < size1; i++)
			{
				for (int j = 0; j < 4; j++)
					patternTable[j, i] = moduleStream.Read_B_UINT16();
			}

			byte[] inputBuffer = new byte[moduleStream.Length - moduleStream.Position];
			moduleStream.ReadInto(inputBuffer, 0, inputBuffer.Length);

			// Convert all the patterns
			Dictionary<int, byte[]> convertedPatterns = new Dictionary<int, byte[]>();

			for (int i = 0; i < numberOfPositions; i++)
			{
				int patternNumber = positionList[i];

				if (!convertedPatterns.TryGetValue(patternNumber, out byte[] outputBuffer))
				{
					outputBuffer = ConvertPattern(inputBuffer, [ patternTable[0, i], patternTable[1, i], patternTable[2, i], patternTable[3, i] ]);
					if (outputBuffer == null)
						yield return null;

					convertedPatterns[patternNumber] = outputBuffer;
				}
			}

			// And return them
			for (int i = 0; i < numberOfPatterns; i++)
			{
				if (convertedPatterns.TryGetValue(i, out byte[] outputBuffer))
					yield return outputBuffer;
				else
					yield return new byte[1024];    // If pattern is not used, return an empty pattern
			}
		}

		/********************************************************************/
		/// <summary>
		/// Convert a single pattern
		/// </summary>
		/********************************************************************/
		private byte[] ConvertPattern(byte[] inputBuffer, ushort[] patternStartOffsets)
		{
			byte[] outputBuffer = new byte[1024];

			for (int j = 0; j < 4; j++)
			{
				int startInputOffset = patternStartOffsets[j];
				if (startInputOffset >= (inputBuffer.Length - 4))
					return null;

				for (int row = 0, k = 0; k < 4; k++)
				{
					for (int x = 0; x < 4; x++)
					{
						for (int y = 0; y < 4; y++, row++)
						{
							int inputOffset = startInputOffset + inputBuffer[startInputOffset + k] + x;
							if (inputOffset >= inputBuffer.Length)
								return null;

							inputOffset = startInputOffset + inputBuffer[inputOffset] + y;
							if (inputOffset >= inputBuffer.Length)
								return null;

							inputOffset = startInputOffset + inputBuffer[inputOffset] * 2;
							if (inputOffset >= inputBuffer.Length)
								return null;

							int outputOffset = row * 16 + j * 4;

							outputBuffer[outputOffset++] = inputBuffer[inputOffset++];
							outputBuffer[outputOffset++] = inputBuffer[inputOffset++];
							outputBuffer[outputOffset++] = inputBuffer[inputOffset++];
							outputBuffer[outputOffset] = inputBuffer[inputOffset];
						}
					}
				}
			}

			return outputBuffer;
		}

```

AliceLR:
Magnetic Fields Packer stores a list of order-tracks after the
order list, rather than the previously assumed pattern-tracks.
This patch fixes the MFP module loader to ignore the order list
and instead use the sequential list of patterns that libxmp
already creates.

The above fix is sufficient to load these modules, but MFP also
merges duplicate tracks, which libxmp ignored (leading to
potentially loading the same track dozens of times). I've added
a track optimization routine to merge duplicate track offsets.
Tracks are now loaded and numbered in the order they're found
inside of the file data, only loaded once, and much less data
is read per track, hopefully resulting in a net faster load.

Closes #872.


AliceLR:
The hold up on this one is I'm fuzzing these changes and also I'm trying to verify every MFP songfile sounds correct (since apparently it was only used in 3 games). Unfortunately one of these games has a very long and repetitive multi-hour soundtrack :(

sezero (approved):

PR #1004, opened 2026-06-14
Replace StarTrekker AM envelope conversion with flt_extras.c

### LINKED ISSUE #94: Startrekker synthesized instruments should "overwrite" existing samples
The code block for loading synthesized AM instruments is guarded by `if (mod->xxs[i].len == 0) {`.
Sample 1 of [fa.worse face.mod](https://www.exotica.org.uk/mediawiki/index.php?title=Special%3AModland&md=search&mod=fa.worse+face.mod&aut=&for=&col=&ins=) however has a non-zero length while being an AM instrument at the same time. I think the check for `am_synth && is_am_instrument(nt, i)` is enough in this case. In fact, I think the regular sample should always be loaded, and then replaced by the synthesized sample if requried, because otherwise the 64 bytes from the first sample will be added in front of the next sample, so all samples are shifted a bit.
In summary: Load the "normal" sample regardless of `is_am_instrument`, and then call `read_am_instrument` afterwards if required (do not forget to free the old sample in case this is not done automatically).

AliceLR:
* Fix broken stage durations for most stages (flt_am_envelope.mod).
* Attack 1 now applies immediately on tick 0 (flt_am_envelope.mod).
* Sustain now lasts one tick longer than ST (flt_am_envelope.mod).
* Speeds of 0 now correctly halt the envelope progression unless the current level is equal to the target (flt_am_envelope.mod).
* Envelope volume now overrides channel volume altogether, which which required switching from XM envelopes to extras (flt_am_ignore_volfx.mod).

I've left a corrected envelope conversion routine with all of the above fixes (except for envelope volume overriding channel volume) in flt_load.c. This may be useful in the future for converting StarTrekker extras to XM envelopes if libxmp gets an editing API.

Other things fixed by switching to extras:

* Support StarTrekker's weird AM instrument number note cutting semantics (flt_am_new_note.mod). This may need a new read_event implementation, but for now, it only affects AM instruments.
* Implement P.FALL by adding it to the current period every tick, fixing interactions with portamento effects (flt_am_period_fall.mod).
* StarTrekker does NOT apply FQ to the toneporta target, so reverse-transpose the target in extras (flt_am_period_fall.mod).

Other related fixes attached to issue 94 (flt_am_override.mod):

* AM instruments should take precedence over sampled instruments (fixes GTS/fa.worse face.mod and WOTW/intro 12.mod).
* AM instruments should be loaded even if they are truncated (fixes GTS/fa.worse face.mod).

Closes #94.


sezero:
Fix build in c++ mode :

```diff
diff --git a/src/flt_extras.h b/src/flt_extras.h
index 28cb1e40..e73b980f 100644
--- a/src/flt_extras.h
+++ b/src/flt_extras.h
@@ -20,17 +20,19 @@ struct flt_instrument_extras {
 	uint16 fq;
 };
 
+enum env_stage {
+	FLT_ATTACK_1,
+	FLT_ATTACK_2,
+	FLT_DECAY,
+	FLT_SUSTAIN,
+	FLT_RELEASE
+};
+
 struct flt_channel_extras {
 	uint32 magic;
 	int volume;
 	int sustain;
-	enum {
-		FLT_ATTACK_1,
-		FLT_ATTACK_2,
-		FLT_DECAY,
-		FLT_SUSTAIN,
-		FLT_RELEASE
-	} env_stage;
+	int env_stage;
 };
 
 struct flt_module_extras {
```

AliceLR:
> Fix build in c++ mode :

Should we have CI for this?

edit: pushed mostly the same fix, but since the enum is named now, I used it directly for `env_stage` and `libxmp_flt_tick_envelope` `next_stage`.
edit 2: force pushed again to rename it to `enum flt_env_stage` so it's less likely to collide in the future.

sezero (approved):

sezero:
> > Fix build in c++ mode :
> 
> Should we have CI for this?

Well, yes, would be nice to add one such instance to linux autotools runs.

PR #1005, opened 2026-06-14
Add Linux Autotools/g++ regression test to CI.

AliceLR:
CMake throws a fit if you specify g++ as the C compiler, so this is Autotools-only.


sezero:
Can you look at #1006 ?

PR #1006, opened 2026-06-14
bump CI, and add a linux g++ run.

sezero:
Note: We have `checking for ceil in -lm... no` in configure log, but I guess it's some autoconf-2.71 issue: Tested autoconf 2.69, 2.71, 2.72 and 2.73 locally, only happens with 2.71 for me when `CC=g++` is specified.

### DIFF
--- a/.github/workflows/regression.yaml
+++ b/.github/workflows/regression.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
       - name: CMake configure
         run: |
           mkdir build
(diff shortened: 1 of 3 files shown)


AliceLR:
Closed the wrong one..

.github/workflows/regression.yaml
```diff
@@ -36,11 +36,31 @@ jobs:
...
+  linux-cxx:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      - name: Create and run configure script
+        run: |
+          autoconf
+          CC=g++ ./configure
+          (cd test-dev && autoconf && CC=g++ ./configure)
```
AliceLR: Should probably be left as part of devcheck like the others..

AliceLR:
I had an initial patch to do the rest of this ready to go after #1005 was merged, but this looks better IMO, so let's go with this.

.github/workflows/regression.yaml
sezero: Done

AliceLR (approved):

PR #1007, opened 2026-06-15
Fix IT high sample offset bugs.

### LINKED ISSUE #1003: Some .it samples don't loop properly
repro: https://sdomi.pl/tmp/sdomi's%20mods/shoob2000_-_that.it 
at the beginning of the song there are several gaps with no audio at all; libmodplug plays it just fine (AFAIK this was made in modplug tracker, so that libmodplug should be authoritative)

I haven't encountered this in any other modules; Tested on latest master libxmp + xmp-cli. Running on Linux, but output driver doesn't matter (happens on pulse and wav output; haven't checked others)

AliceLR:
* High sample offset should NOT be applied unless Oxx is actually present (fixes shoob2000_-_that.it and probably many others).
* Offsets past the end of a sample should set the sample offset to 0 in new effects mode. These offsets should continue to clamp to the end of the sample in old effects mode.

Closes #1003.


AliceLR:
I guess it was only a matter of time 🙂 
```
LD libxmp-tests
/bin/sh: line 3: /mingw64/bin/gcc: Argument list too long
make: *** [Makefile:817: libxmp-tests] Error 126
```

sezero:
> I guess it was only a matter of time 🙂
> 
> ```
> LD libxmp-tests
> /bin/sh: line 3: /mingw64/bin/gcc: Argument list too long
> make: *** [Makefile:817: libxmp-tests] Error 126
> ```

We should switch to using response files for linkage.

AliceLR:
> We should switch to using response files for linkage.

Wasn't familiar with them, but it sounds better than my idea (putting test subsets in their own `.a`).

edit: I can put together a patch for that tomorrow...

AliceLR:
> > We should switch to using response files for linkage.
> 
> Wasn't familiar with them, but it sounds better than my idea (putting test subsets in their own `.a`).

To allow `clean`, `all_tests.c`, and response file generation natively on MinGW, the test names need to be split up anyway, so it works out to about the same changes as splitting the test objects into several archives.

Response files also might be a GCC 4.2 feature, so we'd need to have an archive fallback anyway?

sezero:
> Response files also might be a GCC 4.2 feature, so we'd need to have an archive fallback anyway?

I don't think anyone will ever run test-dev with gcc < 4.2 under windows, no?

sezero (approved):
IMO, windows too-many-args issue can be decoupled from this one as a non-blocker: I'm OK with this patch as it is.

AliceLR:
> > Response files also might be a GCC 4.2 feature, so we'd need to have an archive fallback anyway?
> 
> I don't think anyone will ever run test-dev with gcc < 4.2 under windows, no?

Any patch made for MinGW would either also have to apply to all other platforms using Autotools or would need special MinGW-only enablement (and require a fallback for non-MinGW). Switching to a response file without a fallback would break GCC <4.2 *on Linux and Mac OS X*.

Because the fallback would just be `ar`, I just went with that for my current draft (#1008). In my testing so far, it also fixes the problem, and is only slightly less clean than a response file (or maybe cleaner given that the response file would require either 1) an initial phony target to clear the response file before the .c.o target appends each object to it, which would also potentially foul up parallel builds, and this would cause Make to relink `libxmp-tests` every time it's run, or 2) might be fine with a `for` loop since it's a Bash built-in, but I haven't tested that conclusively yet. Conversely, `ar` plays well with Make's expected dependency chaining).

sezero:
> > I don't think anyone will ever run test-dev with gcc < 4.2 under windows, no?
> 
>  Switching to a response file without a fallback would break GCC <4.2 _on Linux and Mac OS X_.

Well, you're right. OK, then.

AliceLR:
> Well, you're right. OK, then.

I'm ok with litigating this further in #1008. :) Let's just merge this for now and let CI break..

PR #1008, opened 2026-06-15
Split test-dev objects into intermediate archives.

AliceLR:
See discussion in #1007. This patch is not finished.

- [x] Can revert the `for` loop changes for all_tests.{c,txt} apparently?
- [x] Can revert the `clean` changes since wildcard expansion apparently also isn't an issue?
- [x] Response file instead of recursive Makefile to future-proof Makefile.vc?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,7 +28,7 @@ test-dev/xmpchk
 *.lib
 *.suo
 *.ncb
-Makefile.vc.tmp
+*.rsp
 
 # VSCode/clangd files
 /.cache
--- a/test-dev/Makefile.in
+++ b/test-dev/Makefile.in
@@ -6,11 +6,22 @@ LDFLAGS		= @LDFLAGS@
 LIBS		= @LIBS@
 V		= 0
 
+TEST_PATH	= .
+SRC_PATH	= ../src
+
 GCLIB		= libxmp-gc.a
+
+default-test: check
 
 .c.o:
 	@CMD='$(CC) -MMD $(CFLAGS) -o $*.o $<'; \
 	if [ "$(V)" -gt 0 ]; then echo $$CMD; else echo CC $*.o ; fi; \
+	eval $$CMD
+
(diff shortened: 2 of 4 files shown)


AliceLR:
Interestingly, it seems like the `for` statements *aren't* affected by the 32767 command line limit since they're Bash built-ins?

edit: likewise, even with 40 added files with long names to get the total test names length over 32767, `rm -f *.o` seems unaffected by the limit.

AliceLR:
MSYS2 just gave me an example of why having the `.c.o` rule append to a file would probably break parallel builds:

image

AliceLR:
> Response file instead of recursive Makefile to future-proof Makefile.vc?

Yep, confirmed necessary :)

sezero:
> > Response file instead of recursive Makefile to future-proof Makefile.vc?
> 
> Yep, confirmed necessary :)

Looks like I didin't touch test-dev in https://github.com/libxmp/libxmp/pull/732 ...

AliceLR:
> Looks like I didin't touch test-dev in #732 ...

The test-dev Makefile.vc is IIRC something I bodged together with even less knowledge of nmake and batch scripting than I have now, so it's understandable that you didn't want to touch it:
* https://github.com/libxmp/libxmp/commit/3041e27c44c14a9020fe3d1d38626c54bcd8a808 (initial)
* https://github.com/libxmp/libxmp/commit/201271acf000ced8d456df4fbd858dc1b32be1d9 (the origin of all_tests.txt and the very weird recursive nmake invocation)

At the time, I was not aware of response files being a thing. This patch is (IMO) a huge improvement and more sensible. Do you think there's a way to make this work with a fully temporary .rsp (like the root Makefile.vc uses), or is this fine?

AliceLR:
(I think this is actually technically ready for review, though it's not perfect.)

sezero (approved):
Tested locally with msvc and works, so it's good enough.

Issue #974, opened 2026-02-19
Use int64 instead of long for hio_tell/hio_seek/hio_size.


AliceLR:
The main/probably only usage for this would be the niche situation where an archive containing multiple files is >2 GiB. Probably very rare, but it'd be nice to be 64-bit compliant internally and with stdio (`ftello`/`fseeko`/`_FILE_OFFSET_BITS == 64` where available).

Memory loading and callbacks would still be limited to `long` due to API limitations (I would be OK introducing some form of `size_t` or `int64` to the public API in the future, but we decided not to when callbacks IO was added).

An additional problem is libxmp doesn't actually support full unarchiving of multi-file formats (it only extracts the first match). To be useful this also needs an unarchiving API.

PR #1013, opened 2026-06-21
Rewrite smix WAV loader and fix smix sample load/free bugs.

AliceLR:
* `xmp_smix_load_sample`: add support for loading stereo samples.
* `xmp_smix_load_sample`: fix signedness of loaded 8-bit samples.
* `xmp_smix_load_sample`: slightly better WAV format integrity checks.
* `xmp_smix_load_sample`: always call `xmp_smix_release_sample` on the destination instrument if the requested file successfully opens. Previously, this would leak old sample/subinstrument data.
* `xmp_smix_release_sample`: fix crashes if the released sample is currently being played by a mixer voice.


src/read_event.c
```diff
@@ -1575,10 +1575,10 @@ static int read_event_smix(struct context_data *ctx, const struct xmp_event *e,
 	RESET_NOTE(NOTE_END);
 
 	if (ins >= mod->ins && ins < mod->ins + smix->ins) {
-		sub = &xxi->sub[0];
-		if (sub == NULL) {
+		if (xxi->sub == NULL) {
 			return 0;
 		}
+		sub = &xxi->sub[0];
```
AliceLR, line 1581: Not actually sure this achieves anything (UBSan didn't complain), but the old version *looks* like a `NULL` dereference.

AliceLR:
Test failure on MINGW32 and macOS likely due to a cosmic ray making them bad at floating point math, will look at it tomorrow.

src/smix.c
```diff
@@ -23,10 +23,136 @@
...
+		if (hio_error(f)) {
+			return -XMP_ERROR_FORMAT;
+		}
+
+		if (magic == WAV_MAGIC_DATA) {
+			break;
+		}
+
+#if LONG_MAX <= 2147483647L
+		if (len > (uint32)LONG_MAX) {
```
sezero, line 125: Why only for `LONG_MAX <= 2147483647L` ,  why not simply `if (len > 0x7fffffff)` ?

sezero:
> Test failure on MINGW32 and macOS likely due to a cosmic ray making them bad at floating point math

Same test fails for me on i686-linux

src/smix.c
AliceLR: It is a guard for 32-bit `long`. 64-bit `long` has no issue with values >`0x7fffffff`

AliceLR:
> > Test failure on MINGW32 and macOS likely due to a cosmic ray making them bad at floating point math
> 
> Same test fails for me on i686-linux

I vaguely remember having to carefully select an OggMod test case that didn't break on x87, but that wouldn't explain why macOS also broke (maybe the cosmic ray flipped the bit that makes Xcode forget that every Intel Mac ever made has SSE2... 🙂)

AliceLR:
Looks like regenerating stereo.oxm at quality level 1 instead of whatever I used before (-1?) was enough to get a stream that renders identically for x87 and IEEE (and it also made a smaller file...).

src/smix.c
AliceLR: (I know this is already marked resolved, but I wanted to add: a nearly identical check is present in iff.c and IMO #974 is the correct way to get rid of both hacks.)

src/read_event.c
sezero: Compiler generates identical code for `ptr = &array[0];` and `ptr = array;` hence no complaints. The new version looks clearer, though.

sezero:
> Looks like regenerating stereo.oxm at quality level 1 instead of whatever I used before (-1?) was enough to get a stream that renders identically for x87 and IEEE

Great.

How did x86_64-darwin fail, though....

src/read_event.c
AliceLR: [cppreference](https://en.cppreference.com/c/language/operator_member_access) has some helpful info about this:

> By definition, the subscript operator `E1[E2]` is exactly identical to `*((E1)+(E2))`.

> 4) special case: `&` and the `*` that is implied in `[]` cancel each other, only the addition implied in `[]` is evaluated.

The old version evaluates to `NULL + 0`, which doesn't break anything, but would potentially encounter issues if smix ever got multi-sample instrument support and that 0 is replaced by any other index. I agree the new version looks cleaner (and can't be mistaken for a dereference).

AliceLR:
I *think* this is OK to review now. (Technically, I think `fmt ` isn't required to be the first chunk, but I've never seen a WAV where it isn't. If the sample loader logic gets generalized further it could be switched to use iff.c or something.)

sezero (approved):

Issue #1002, opened 2026-06-12
Change XMP_PLAYER_DEFPAN default to 50


elf-alchemist:
Reporting on the behalf of another user experiencing sound panning issues.
The following MOD file sounds differently when played between libopenmpt and libxmp.

[_wandering_.zip](https://github.com/user-attachments/files/28880844/_wandering_.zip)

Example from the "DSDA-Doom" doom source port:

older build using libopenmpt:
https://www.youtube.com/watch?v=2wYuKeeQOu0

and current build using libxmp:
https://www.youtube.com/watch?v=rSK25K4Kq3o

Cryoniq:
Greetings. It's me who reported that issue on dsda-doom discord. 
"wandering.zip" is not the only MOD that has such an issue. There are severak more MOD files that have same sound panning behavior. I'm attaching some of the MODs, also attaching another two examples, audio only this time.

[jf357_libopenmpt.mp3](https://github.com/user-attachments/files/28891609/jf357_libopenmpt.mp3)
[jf357_libxmp.mp3](https://github.com/user-attachments/files/28891610/jf357_libxmp.mp3)

In the attached .zip file there are 6 music modules with the same problem and a text file with the MOD sources.
[libxmp panning issue.zip](https://github.com/user-attachments/files/28891659/libxmp.panning.issue.zip)

Since these mods seem to have exactly the same issue, maybe there is a way to fix everything at once?

P.S. thanks Elf for opening this issue :)

AliceLR:
To clarify: is the issue that libxmp is using full panning separation, vs. libopenmpt using 50% panning separation?

If so, this is not a libxmp bug. DSDA-Doom needs to add this line after `xmp_create_context` and before `xmp_load_module*`:
```C
xmp_set_player(ctx, XMP_PLAYER_DEFPAN, 50);
```

(This option only affects hard-panned formats like MOD, 669, etc. Formats with native panning like XM/IT are not affected.)

elf-alchemist:
[nosp2_map07_mod_test.zip](https://github.com/user-attachments/files/28952216/nosp2_map07_mod_test.zip)

I tested setting this property on Woof (it has basically the same setup as DSDA-Doom, as it was also done by the same contributor), and the effect bore no noticeable difference. Perhaps I did something wrong, this isn't my area of expertise :v I'm just a Doom dev concerned with compatibility.

Comment from Cryoniq on Discord
[...]

PR #1014, opened 2026-06-24
Update libxmp.rst and examples to set XMP_PLAYER_DEFPAN to 50.

AliceLR:
* In all places where relevant, the examples set the panning separation now. The showinfo example sets it to 100 with a comment explaining why, and the rest set it to 50.
* libxmp.rst now explicitly states that `XMP_PLAYER_SMPCTL` and `XMP_PLAYER_DEFPAN` must be provided before loading a module.
* Fix examples/Makefile when `CC`, `CFLAGS`, `LDFLAGS` are externally set.
* Add a clean rule to examples/Makefile.
* Fix SDL examples ignoring exit events.

Assuming default panning is the problem with #1002, this addresses everything actionable regarding that issue that doesn't require a major version change. (Changelog entry here should go before the one from #1013.)

(I did not add SDL3 versions of the SDL examples, but I can in another patch if it'd be helpful.)

### DIFF
--- a/docs/Changelog
+++ b/docs/Changelog
@@ -3,6 +3,9 @@ Stable versions
 
 4.7.1 (?):
 	Changes by Alice Rowan:
+	- Add xmp_set_player(ctx, XMP_PLAYER_DEFPAN, 50) to relevant examples
+	  in libxmp.rst and in the examples directory, clarify it must be
+	  used BEFORE calling xmp_load_module, and recommend its usage.
 	- Fix XM channel default instrument memory (should be no instrument).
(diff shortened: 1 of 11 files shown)


sezero (approved):

sezero:
>  In all places where relevant, the examples set the [...]

P.S.: Should we include the examples directory in release tarballs?

AliceLR:
> > In all places where relevant, the examples set the [...]
> 
> P.S.: Should we include the examples directory in release tarballs?

That would be reasonable, I think.

PR #1016, opened 2026-06-24
flt_load.c: Remove use of addressof operator on an array

sezero:
Fixes W007 warning from Watcom C++:

```
src/loaders/flt_load.c(225): Warning! W007: col(16) '&array' may not produce intended result
```

Why do we not have an equivalent warning option for this in gcc?...

### DIFF
--- a/src/loaders/flt_load.c
+++ b/src/loaders/flt_load.c
@@ -222,7 +222,7 @@ static int read_am_instrument(struct module_data *m, HIO_HANDLE *nt, int i)
 	uint8 buf[30];
 	int8 am_noise[1024];
 
-	memset(&buf, 0, sizeof(buf));
+	memset(buf, 0, sizeof(buf));
 
 	hio_seek(nt, 144 + i * 120 + 2 + 4, SEEK_SET);
 	/* Allow partial/missing AM instruments (GTS/fa.worse face.mod). */


AliceLR (approved):
> Why do we not have an equivalent warning option for this in gcc?...

Not sure. This sneaked in because the original `memset` I added was on the AM instrument struct (which wouldn't have fixed the issue)... 🥲

Clownacy:
What's wrong with it? If anything, an array implicitly decaying into a pointer to its first element is the one that "may produce an unintended result". I imagine GCC lacks an option for it because it would be nonsensical.

sezero:
> What's wrong with it? If anything, an array implicitly decaying into a pointer to its first element is the one that "may produce an unintended result". I imagine GCC lacks an option for it because it would be nonsensical.

It's not about `&array[0]`, it's about `&array`

AliceLR:
> What's wrong with it? If anything, an array implicitly decaying into a pointer to its first element is the one that "may produce an unintended result". I imagine GCC lacks an option for it because it would be nonsensical.

Given `uint8_t buffer[16]`, `buffer` and `&buffer` have the same value, but pointer arithmetic breaks because the new type is `uint8_t (*)[16]` [[1]](https://stackoverflow.com/questions/38202077/what-does-getting-the-address-of-an-array-variable-mean) [[2]](https://learn.microsoft.com/en-us/answers/questions/87357/address-of-array-in-c) (trying to find an actual C standard citation...). This is what Watcom is trying to warn about.

sezero:
And it can lead to disastrous results -- I remember scratching my head for a long time until I noticed this in someone else's code (which came to light only via watcom's warning - I guess I was blind then).: https://github.com/sezero/SDL-OS2/commit/f94a368694fb70acc4a7f1de7ca6e63ae5fc73b4

PR #1018, opened 2026-07-04
xm_load.c: Detect and handle NitroTracker quirks.

asiekierka:
This change handles the most egregious NitroTracker-specific quirks by adjusting loaded XM instrument/pattern data - the same way as its fork, NitrousTracker 0.7.0+, will handle backwards compatibility going forward now that it uses a derivative of FastTracker 2's play routines. It probably does not make songs play identically by itself, but it fixes cases of dragged notes and incorrect panning that I've observed myself.

The accuracy can be checked with the [demo songs](https://github.com/NitrousTracker/nitroustracker/releases/download/release%2F0.6.4/NitrousTracker-0.6.4.zip) packaged with the tracker.

## Licensing

- [x] I hereby assign my copyright for this contribution to the libxmp

I am the developer of NitrousTracker, which is a fork of NitroTracker. The latter is licensed under the GNU GPL, however I have received permission from its original developer (0xtob) to use information from libntxm for the purpose of open source contributions under any relevant license, including the MIT license.

### DIFF
--- a/src/loaders/xm_load.c
+++ b/src/loaders/xm_load.c
@@ -749,6 +749,30 @@ static int load_instruments(struct module_data *m, int version,
 	return 0;
 }
 
+#ifndef LIBXMP_CORE_PLAYER
+static void load_fix_nitrotracker(struct xmp_module *mod)
+{
+	int i, j, k;
+	struct xmp_event *event;
+
+	for (i = 0; i < mod->ins; i++) {
+		/* NT writes uninitialized memory to new instrument pan envelope flags */
+		if (!mod->xxi[i].pei.npt)
+			mod->xxi[i].pei.flg = 0;
+	}
+
+	for (i = 0; i < mod->trk; i++) {
+		for (j = 0; j < mod->xxt[i]->rows; j++) {
+			event = &mod->xxt[i]->event[j];
+
+			/* NT ignores instrument numbers on non-note lines */
+			if (event->note == XMP_KEY_OFF || !event->note)
+				event->ins = 0;
+		}
+	}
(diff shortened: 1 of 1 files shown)


AliceLR:
Adding a note here: asie confirmed in a follow-up DM that the original developer signed off on this. (I would like to figure out how intrusive it would be to handle this during load rather than after load, which can be done some other time.)

sezero:
> Adding a note here: asie confirmed in a follow-up DM that the original developer signed off on this.

Then we're good, as far as licencing is concerned,

> (I would like to figure out how intrusive it would be to handle this during load rather than after load, which can be done some other time.)

How much work is left?

asiekierka:
So far, no issues with the solution as presented have come up in our own testing.

AliceLR:
> How much work is left?

I don't think getting quirk flags to the instrument/pattern load functions would be an issue, but there isn't a regression test module yet to compare against.

@asiekierka: please add a test module to the patch which requires both the instrument and event quirks (`test-dev/data/m/` for a real module, `test-dev/data/nitro_quirks.xm` or something similar for a crafted test module; either way, smaller is better). I can then set up a regression test for it.

AliceLR:
asie recommended delaying this until 4.7.3 due to being busy with other things.

sezero:
> 4.6.3

4.7.3, yes?

> asie recommended delaying this until 4.6.3 due to being busy with other things.

Pity, but OK.

AliceLR:
> > 4.6.3
> 
> 4.7.3, yes?

yes..

PR #1019, opened 2026-07-13
stb_vorbis.c: Fix floor1 inverse_db_table indexing

sezero:
CVE-2019-13220 fix added y&255 mask in draw_line(). However, do_floor()
has an inline optimization equivalent to draw_line(target,lx,ly,n,ly,n2)
without the same mask.
draw_line() uses inverse_db_table[y&255], but this doesn't. If bad input
causes ly to exceed 255, this can read past the end of inverse_db_table.
c.f.: https://github.com/nothings/stb/issues/1934.
Description from mainstream PR https://github.com/nothings/stb/pull/1973

## Licensing

- [ ] I hereby assign my copyright for this contribution to the libxmp

[ This is a third party's patch to third party code as noted above, ]

### DIFF
--- a/src/loaders/vorbis.c
+++ b/src/loaders/vorbis.c
@@ -3172,7 +3172,7 @@ static int do_floor(vorb *f, Mapping *map, int i, int n, float *target, YTYPE *f
       if (lx < n2) {
          // optimization of: draw_line(target, lx,ly, n,ly, n2);
          for (j=lx; j < n2; ++j)
-            LINE_OP(target[j], inverse_db_table[ly]);
+            LINE_OP(target[j], inverse_db_table[ly&255]);
          CHECK(f);
       }
    }


AliceLR (approved):
This is fine. I think the upstream patch could do slightly better (`ly` doesn't change in this loop, so I think it could be bounded prior?) but it's easier to just accept this as-is.

AliceLR:
[This is supposed to be a 7.1](https://nvd.nist.gov/vuln/detail/CVE-2019-13220), seriously? I can rename the milestones if we want to do a release over this, I guess.

sezero:
> [This is supposed to be a 7.1](https://nvd.nist.gov/vuln/detail/CVE-2019-13220), seriously? I can rename the milestones if we want to do a release over this, I guess.

No such need, IMHO.

PR #1021, opened 2026-07-15
cmake: Do a build+link check for pipe, wait, fork, execvp, and dup2

### LINKED ISSUE #1020: tvOS: Build failed due to "fork" and friends functions
Hello! I trying to build a thing for Apple's tvOS that is similar to iOS, and I noticed that it fails to build because of some functions:
```
Debug-appletvos/Objects-normal/arm64/depacker.o
/Users/vitaly/Repos/TheXTech/3rdparty/AudioCodecs/libxmp/src/depackers/depacker.c:162:13: error: 'fork' is unavailable: not available on tvOS
  162 |         if ((pid = fork()) < 0) {
      |                    ^
In file included from /Users/vitaly/Repos/TheXTech/3rdparty/AudioCodecs/libxmp/src/depackers/depacker.c:144:
/Applications/Xcode.app/Contents/Developer/Platforms/AppleTVOS.platform/Developer/SDKs/AppleTVOS26.5.sdk/usr/include/unistd.h:459:8: note: 'fork' has been explicitly marked unavailable here
  459 | pid_t    fork(void) __WATCHOS_PROHIBITED __TVOS_PROHIBITED;
      |          ^
/Users/vitaly/Repos/TheXTech/3rdparty/AudioCodecs/libxmp/src/depackers/depacker.c:174:3: error: 'execvp' is unavailable: not available on tvOS
  174 |                 execvp(cmd[0], (char * const *)cmd);
```
That means, these functions are presented at unistd.h, but compiler screams the error above when attempting to build.

And seems the build is fixed when I replaced at libxmp-checks.cmake this line:
```
if(HAVE_UNISTD_H)
```
With:
```
if(HAVE_UNISTD_H AND (NOT APPLE OR (APPLE AND NOT TVOS)))
```

sezero:
Also add iOS and tvOS workflows to CI

Fixes https://github.com/libxmp/libxmp/issues/1020

## Licensing

- [x] I hereby assign my copyright for this contribution to the libxmp

### DIFF
--- a/.github/workflows/regression.yaml
+++ b/.github/workflows/regression.yaml
@@ -262,7 +262,7 @@ jobs:
         run: brew install autoconf
       - name: CMake configure for Intel
         run: |
-          cmake . -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=ON -B build
+          cmake . -DCMAKE_BUILD_TYPE=Release -DCMAKE_OSX_ARCHITECTURES="x86_64" -DWITH_UNIT_TESTS=ON -B build
       - name: CMake build for Intel
         run: |
           cmake --build build --target all -- -j 4
@@ -276,6 +276,18 @@ jobs:
       - name: CMake build for M1
         run: |
           cmake --build build-m1 --target all
+      - name: CMake configure for iOS
+        run: |
+          cmake . -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=OFF -B build-ios
+      - name: CMake build for iOS
+        run: |
+          cmake --build build-ios --target all
+      - name: CMake configure for tvOS
+        run: |
+          cmake . -DCMAKE_SYSTEM_NAME=tvOS -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=OFF -B build-tvos
+      - name: CMake build for tvOS
+        run: |
+          cmake --build build-tvos --target all
       - name: Create and run Autotools configure script
         run: |
           autoconf
--- a/cmake/libxmp-checks.cmake
+++ b/cmake/libxmp-checks.cmake
@@ -144,16 +144,28 @@ cmake_pop_check_state()
 xmp_check_function(popen "stdio.h" HAVE_POPEN)
 xmp_check_function(fnmatch "fnmatch.h" HAVE_FNMATCH)
 xmp_check_function(umask "sys/stat.h" HAVE_UMASK)
-xmp_check_function(wait "sys/wait.h" HAVE_WAIT)
 xmp_check_function(mkstemp "stdlib.h" HAVE_MKSTEMP)
 
 check_include_file(unistd.h HAVE_UNISTD_H)
+check_include_file(sys/wait.h HAVE_SYS_WAIT_H)
 
-if(HAVE_UNISTD_H)
-    xmp_check_function(pipe "unistd.h" HAVE_PIPE)
-    xmp_check_function(fork "unistd.h" HAVE_FORK)
-    xmp_check_function(execvp "unistd.h" HAVE_EXECVP)
-    xmp_check_function(dup2 "unistd.h" HAVE_DUP2)
+if(HAVE_UNISTD_H AND HAVE_SYS_WAIT_H)
+    check_c_source_compiles("#include <unistd.h>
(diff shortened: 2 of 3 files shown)


cmake/libxmp-checks.cmake
```diff
@@ -144,16 +144,28 @@ cmake_pop_check_state()
...
+    int main(void) { wait(&status); return 0; }" HAVE_WAIT)
+  check_c_source_compiles("#include <unistd.h>
+    pid_t pid;
+    int main(void) { pid = fork(); return 0; }" HAVE_FORK)
+  check_c_source_compiles("#include <unistd.h>
+    int main(int argc, char **argv) { execvp(argv[0], (char* const*)argv);
+    return argc; }" HAVE_EXECVP)
+  check_c_source_compiles("#include <unistd.h>
+    int fds[2];
+    int main(void) { dup2(fds[1], STDOUT_FILENO); return 0; }" HAVE_DUP2)
```
Wohlstand: Please align these lines by 4 spaces to match the code style.

.github/workflows/regression.yaml
```diff
@@ -276,6 +276,18 @@ jobs:
...
+        run: |
+          cmake . -DCMAKE_SYSTEM_NAME=iOS -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=OFF -B build-ios
+      - name: CMake build for iOS
+        run: |
+          cmake --build build-ios --target all
+      - name: CMake configure for tvOS
+        run: |
+          cmake . -DCMAKE_SYSTEM_NAME=tvOS -DCMAKE_OSX_ARCHITECTURES="arm64" -DCMAKE_BUILD_TYPE=Release -DWITH_UNIT_TESTS=OFF -B build-tvos
+      - name: CMake build for tvOS
+        run: |
```
Wohlstand, line 289: You would like to disable shared libraries builds as the build might fail for missing code signing if Xcode generator was being used.

Wohlstand: OR! You might want to append these flags:
`-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED="NO"`
and 
`-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_REQUIRED="NO"`
to diable requirement of code signing and keep shared libs being built.

cmake/libxmp-checks.cmake
sezero: I don't care about those..

.github/workflows/regression.yaml
sezero: The workflows don't seem to fail, though ???

Wohlstand: Because it ran the `UNIX Makefiles` generator. These rules I noted are affected by `Xcode` generator, actually.

sezero: > Because it ran the `UNIX Makefiles` generator. These rules I noted are affected by `Xcode` generator, actually.

>   append these flags: `-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_ALLOWED="NO"` and `-DCMAKE_XCODE_ATTRIBUTE_CODE_SIGNING_REQUIRED="NO"

We never used those for normal macOS workflows either.. @AliceLR: Should we do that, and for all Apple-tageting checks?

Wohlstand: Xcode build is usually used to produce iOS applications rather than libraries, and these apps can be deployed to device, simulator, or packed into the archive thing.

sezero: Hm, I based the workflow stuff (cmake flags, etc) on SDL's workflows, and SDL doesn't set those code-sign cmake flags either.

I'm leaning towards not defining them here. Leaving final decision to @AliceLR .

AliceLR: IMO, we can fix it for CI later if it does break. If they change the default generator for macOS I guarantee it would break a *lot* more than just our CI.

cmake/libxmp-checks.cmake
AliceLR: We should probably stick with the 4 space indentation and justified strings for consistency (see lines 185-202 where `__attribute__((visibility))` is tested).

.github/workflows/regression.yaml
sezero: OK, not adding those cmake flags and leaving as-is for now.

cmake/libxmp-checks.cmake
sezero: Okay, done..

AliceLR (approved):

